PR #8, opened 2015-09-08
Do not clean on CI deploys.

JakeWharton:


mattprecious:
Logic LGTM. The syntax is beyond my bash skillset.

PR #10, opened 2015-09-14
Throw exceptions when Duktape is unhappy.

swankjesse:


duktape/src/main/jni/duktape-jni.c
```diff
@@ -20,11 +20,18 @@ Java_com_squareup_duktape_Duktape_evaluate__JLjava_lang_String_2(JNIEnv *env,
...
   const char *s = (*env)->GetStringUTFChars(env, s_, 0);
 
-  duk_eval_string(ctx, s);
-  jstring result = (*env)->NewStringUTF(env, duk_get_string(ctx, -1));
-  duk_pop(ctx);
+  jstring result = NULL;
 
-  (*env)->ReleaseStringUTFChars(env, s_, s);
+  if (duk_peval_string(ctx, s) != 0) {
+      jclass Exception = (*env)->FindClass(env, "com/squareup/duktape/DuktapeException");
```
JakeWharton: two-space indent?

duktape/src/main/jni/duktape-jni.c
```diff
@@ -20,11 +20,18 @@ Java_com_squareup_duktape_Duktape_evaluate__JLjava_lang_String_2(JNIEnv *env,
...
+      (*env)->ThrowNew(env, Exception, duk_safe_to_string(ctx, -1));
+      goto cleanup;
+  }
+
+  result = (*env)->NewStringUTF(env, duk_get_string(ctx, -1));
```
szurbrigg: Usually `goto` is considered bad form in C/C++.  You could avoid it by putting this line inside an else block.

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -42,7 +42,11 @@ private Duktape(long context) {
     this.context = context;
   }
 
-  /** Evaluate {@code script} and return any result. */
+  /**
+   * Evaluate {@code script} and return any result
```
JakeWharton: You destroyed the period!

JakeWharton:
LGTM

duktape/src/main/jni/duktape-jni.c
szurbrigg: Should we use `duk_safe_to_string` instead of `duk_get_string` here too?

duktape/src/main/jni/duktape-jni.c
swankjesse: Fixed. Fighting Android Studio on this one.

duktape/src/main/jni/duktape-jni.c
swankjesse: Fixed to not goto. I was using it as a moral equivalent of `finally`. I don't think we want `safe` here; that's only necessary if there's an error.

szurbrigg:
Beauty.

PR #13, opened 2015-10-02
Use 'build' task on CI to ensure compilation actually happens.

JakeWharton:


mattprecious:
Does `build` depend on `check`? If so, LGTM!

JakeWharton:
Yep.
```
duktape:build - Assembles and tests this project. [duktape:assemble, duktape:check]
```

PR #17, opened 2015-10-06
Get the timezone offset from Java rather than libc.

swankjesse:
The C library on Android doesn't know anything about timezones.

### DIFF
--- a/duktape/src/main/java/com/squareup/duktape/Duktape.java
+++ b/duktape/src/main/java/com/squareup/duktape/Duktape.java
@@ -16,6 +16,8 @@
 package com.squareup.duktape;
 
 import java.io.Closeable;
+import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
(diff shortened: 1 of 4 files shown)


swankjesse:
FYI @szurbrigg

JakeWharton:
This has unfortunate side-effects on our upgrade story. Do you know if Duktape 1.3.0's `duk_config.h` will allow us to do this in a way that doesn't modify the core files?

JakeWharton:
LGTM

swankjesse:
Yes, I think there's a cleaner way to implement this. We need to do something a _little_ bit interesting, relying on the C library where it is sufficient and on our supplements where it is not. That probably means a bit of configuration, as [explained here](https://github.com/svaarala/duktape/blob/master/doc/datetime.rst).

This PR is taking shortcuts to unblock some things that want timezones ASAP!

JakeWharton:
Yep. Really need to get some tests going on this project!

PR #23, opened 2015-10-26
Add an overload to Duktape.evaluate that accepts a filename argument …

szurbrigg:
…to be used for error reporting.


swankjesse:
LGTM

PR #24, opened 2015-10-27
Add the Duktape stack trace to the thrown DuktapeException as StackTr…

szurbrigg:
…aceElements rather than including it in the message string.

### DIFF
--- a/duktape/src/main/java/com/squareup/duktape/DuktapeException.java
+++ b/duktape/src/main/java/com/squareup/duktape/DuktapeException.java
@@ -15,8 +15,56 @@
  */
 package com.squareup.duktape;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public final class DuktapeException extends RuntimeException {
+  /**
+   * Duktape stack trace strings have multiple lines of the format "function filename.ext:line".
+   * "function" is optional, but we'll omit frames without a function, since it means the frame is
+   * in native code.
+   */
+  private final static Pattern STACK_TRACE_PATTERN =
+      Pattern.compile("\\s*([^\\s]+) ([^\\s]+):(\\d+).*$");
+  /** Java StackTraceElements require a class name.  We don't have one in JS, so use this. */
+  private final static String STACK_TRACE_CLASS_NAME = "JavaScript";
+
   public DuktapeException(String detailMessage) {
-    super(detailMessage);
+    super(getErrorMessage(detailMessage));
+    String[] lines = detailMessage.split("\n", -1);
+    if (lines.length <= 1) {
+      return;
+    }
+    // We have a stacktrace following the message.  Add it to this exception.
+    List<StackTraceElement> elements = new ArrayList<>();
+    for (int i = 1; i < lines.length; ++i) {
+      StackTraceElement stackTraceElement = toStackTraceElement(lines[i]);
+      if (stackTraceElement == null) {
+        continue;
+      }
+      elements.add(stackTraceElement);
+    }
+    Collections.addAll(elements, getStackTrace());
+    setStackTrace(elements.toArray(new StackTraceElement[elements.size()]));
+  }
+
+  private static String getErrorMessage(String detailMessage) {
+    int end = detailMessage.indexOf('\n');
+    return end > 0
+        ? detailMessage.substring(0, end)
+        : detailMessage;
+  }
+
+  private static StackTraceElement toStackTraceElement(String s) {
+    Matcher m = STACK_TRACE_PATTERN.matcher(s);
+    if (!m.matches()) {
+      // Nothing interesting on this line.
+      return null;
+    }
+    return new StackTraceElement(STACK_TRACE_CLASS_NAME, m.group(1), m.group(2),
+            Integer.parseInt(m.group(3)));
   }
 }
--- a/tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
+++ b/tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
@@ -47,7 +47,7 @@ public final class DuktapeTest {
       duktape.evaluate("nope();");
       fail();
     } catch (DuktapeException e) {
-      assertThat(e.getMessage()).startsWith("ReferenceError: identifier 'nope' undefined");
+      assertThat(e).hasMessage("ReferenceError: identifier 'nope' undefined");
     }
   }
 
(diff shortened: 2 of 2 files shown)


duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -15,8 +15,58 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public final class DuktapeException extends RuntimeException {
+  /**
+   * Duktape stack trace strings have multiple lines of the format "function filename.ext:line".
+   * "function" is optional, but we'll omit those frames from the exception, since it means the
```
swankjesse: ... but we'll omit frames without a function...

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -15,8 +15,58 @@
...
+   * frame is in native code.
+   */
+  private final static Pattern STACK_TRACE_PATTERN =
+      Pattern.compile("\\s*([^\\s]+) ([^\\s]+):(\\d+).*");
```
swankjesse: should this have a trailing `$` so the line is at the end? I'm wondering about hypothetical filenames like `hello:123.js`

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -15,8 +15,58 @@
...
+  /** Java StackTraceElements require a class name.  We don't have one in JS, so use this. */
+  private final static String STACK_TRACE_CLASS_NAME = "Javascript";
```
swankjesse: nit: `JavaScript` ?

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -15,8 +15,58 @@
...
+
   public DuktapeException(String detailMessage) {
-    super(detailMessage);
+    super(getErrorMessage(detailMessage));
+    String[] lines = TextUtils.split(detailMessage, "\n");
```
swankjesse: I don't know what `TextUtils` is, but you probably don't want it. It means you can't test on the desktop VM.

swankjesse: (Probably easier is `detailMessage.split("\n", -1)`)

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -15,8 +15,58 @@
...
+   * Duktape stack trace strings have multiple lines of the format "function filename.ext:line".
+   * "function" is optional, but we'll omit those frames from the exception, since it means the
+   * frame is in native code.
+   */
+  private final static Pattern STACK_TRACE_PATTERN =
+      Pattern.compile("\\s*([^\\s]+) ([^\\s]+):(\\d+).*");
+  /** Java StackTraceElements require a class name.  We don't have one in JS, so use this. */
+  private final static String STACK_TRACE_CLASS_NAME = "Javascript";
+
   public DuktapeException(String detailMessage) {
```
swankjesse: Can you just change this to pass two strings, the error message and the stacktrace? That way if the error message has a `\n` in it, we don't corrupt.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -15,8 +15,58 @@
...
+  }
+
+  private static String getErrorMessage(String detailMessage) {
+    int end = detailMessage.indexOf('\n');
+    return end > 0
+        ? detailMessage.substring(0, end)
+        : detailMessage;
+  }
+
+  private static StackTraceElement toStackTraceElement(String s) {
```
swankjesse: Awesome.

swankjesse:
This is going to be great. Lots of feedback & I'd like another round once you've addressed this feedback.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
szurbrigg: I could, but we get that combined string from Duktape, so we'd still have to split on `\n` in JNI.  Also, we'd lose the ability to use the `ThrowNew` shorthand function, so in JNI we'd have to look up the constructor method for the exception, allocate it, and then throw.

IMO, the way it is here is preferable, but I can change it if you feel strongly.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
szurbrigg: Good point. Will do.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
szurbrigg: Sure.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
szurbrigg: Yeah, that makes sense.  (Although, are colons in filenames legal?)

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
szurbrigg: Will change.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
swankjesse: Everything is legal. Some things are escaped.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
swankjesse: Nope, if we have to manually split we might as well do it in managed code.

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -15,8 +15,56 @@
...
+  private final static Pattern STACK_TRACE_PATTERN =
+      Pattern.compile("\\s*([^\\s]+) ([^\\s]+):(\\d+).*$");
+  /** Java StackTraceElements require a class name.  We don't have one in JS, so use this. */
+  private final static String STACK_TRACE_CLASS_NAME = "JavaScript";
+
   public DuktapeException(String detailMessage) {
-    super(detailMessage);
+    super(getErrorMessage(detailMessage));
+    String[] lines = detailMessage.split("\n", -1);
+    if (lines.length <= 1) {
```
swankjesse: ultranit: split never returns empty arrays, so this _could_ be `if (lines.length == 1)`

swankjesse: (Actually, `split()` can return an empty array, but only in broken mode where you don't pass `-1` as the 2nd arg)

PR #25, opened 2015-11-10
Update Duktape to 1.3.0

MichaelEvans:


MichaelEvans:
I guess this isn't as straightforward as I had hoped - based on the modifications in #17

swankjesse:
Next steps . . . we need to keep timezones working. Want to fix the C code? Or back this out?

swankjesse:
Once we have the timezone stuff worked out, this change will make more sense.

MichaelEvans:
I updated the script in #26 to patch `duk_config.h` with the necessary changes in order to keep the timezone bits working. This PR was generated from that script (I'm hardly a bash wizard, that script can probably be improved) but it maintains the timezone tweaks that were in place.

duktape/src/main/jni/duk_custom.h
```diff
@@ -0,0 +1 @@
+#define DUK_USE_DATE_GET_LOCAL_TZOFFSET android__get_local_tzoffset
```
JakeWharton: Trailing newline

/me ducks

MichaelEvans: don't you mean `/me duks`

JakeWharton: touché

swankjesse:
This is awesome.

Issue #16, opened 2015-10-06
Script to automate upstream updating


JakeWharton:
Should:

 * Download latest from http://duktape.org/download.html
 * Extract contents somewhere temporary
 * Copy contents of extracted `src/` to `duktape/src/main/jni/`
 * Delete temporary files

PR #26, opened 2015-11-10
Naive updater script

MichaelEvans:
As mentioned in #16, a script to update duktape from upstream automatically.

I'm hardly a bash expert, so let me know if anything isn't kosher (like the use of `wget`, or the use of grep to pull the url from Github) and I can come up with another solution.

I did create #25 using it though, FWIW.

Edit: Now the script will patch `duk_config.h` to modify the timezone offset to use the java implementation from #17 

Edit 2: Now the script does less...because it doesn't have to patch anything.

### DIFF
--- a/update_duktape.sh
+++ b/update_duktape.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+SRC_LOCATION=duktape/src/main/jni/
+#Download the things
+wget -O duktape.tar.xz $(curl -s https://api.github.com/repos/svaarala/duktape/releases/latest | grep 'browser_' | cut -d\" -f4) 
+#Extract the release
+mkdir tmp
+tar xvfC duktape.tar.xz tmp/
+#Copy over the new sources
+cp -R tmp/duktape-*/src/ $SRC_LOCATION
+#Cleanup after ourselves
+rm duktape.tar.xz
+rm -r tmp


update_duktape.sh
```diff
@@ -0,0 +1,14 @@
+SRC_LOCATION=duktape/src/main/jni/
+#Download the things
+wget -O duktape.tar.xz $(curl -s https://api.github.com/repos/svaarala/duktape/releases/latest | grep 'browser_' | cut -d\" -f4) 
+#Extract the release
+mkdir tmp
+tar xvfC duktape.tar.xz tmp/
+#Copy over the new sources
+cp -R tmp/duktape-*/src/ $SRC_LOCATION
+#Patch the duk_config file to use java's timezone offset
+sed -i bak -e 's/^#undef DUK_USE_DATE_GET_LOCAL_TZOFFSET.*$/#define DUK_USE_DATE_GET_LOCAL_TZOFFSET android__get_local_tzoffset/' $SRC_LOCATION/duk_config.h
```
JakeWharton: You can pass `''` to `-i` to skip the backup which will eliminate the need for the `rm` of it below.

update_duktape.sh
```diff
@@ -0,0 +1,14 @@
+SRC_LOCATION=duktape/src/main/jni/
```
JakeWharton: How about a shebang up here? We don't know what shell a user is running which might break even the most simple things like star expanding.

```
#!/bin/bash
```

update_duktape.sh
JakeWharton: We should use [this mechanism][1] instead, which allows us to not have to modify the upstream library.

 [1]: https://github.com/svaarala/duktape-releases/blob/v1.3.0/src/duk_config.h#L3037-L3047

JakeWharton:
Thanks!

PR #27, opened 2015-11-11
Add Chromium Octane benchmark activity to test app.

JakeWharton:
Two of these don't work for some reason, but the rest at least give a spread of numbers that we can use in the future to look at the underlying Duktape library and how its performance changes over time on Android.

screenshot_20151111-030600


tests/src/main/java/com/squareup/duktape/tests/OctaneActivity.java
```diff
@@ -0,0 +1,68 @@
+package com.squareup.duktape.tests;
```
swankjesse: Nit: copyright

tests/src/main/java/com/squareup/duktape/tests/OctaneActivity.java
```diff
@@ -0,0 +1,68 @@
...
+      throw new RuntimeException(e);
+    }
+    run.setEnabled(true);
+    output.setText(benchmark);
+  }
+
+  private String benchmark() throws IOException {
+    Duktape duktape = Duktape.create();
+
+    for (String file : getAssets().list("octane")) {
```
swankjesse: might also be worthwhile measuring the time to evaluate these assets. We care both about parse time & execution time.

tests/src/main/java/com/squareup/duktape/tests/OctaneActivity.java
```diff
@@ -0,0 +1,68 @@
...
+import android.os.Bundle;
+import android.view.View;
+import android.view.View.OnClickListener;
+import android.widget.TextView;
+import com.squareup.duktape.Duktape;
+import java.io.IOException;
+import okio.BufferedSource;
+import okio.Okio;
+
+public final class OctaneActivity extends Activity implements Runnable {
```
swankjesse: I was expecting a background thread & a spinner!

If an Android app holds the main thread down for many seconds does the OS give you an "App isn't responding" alert? I know Mac OS X does this...

tests/src/main/java/com/squareup/duktape/tests/OctaneActivity.java
JakeWharton: Yep, good idea

swankjesse:
LGTM

I don't insist on the spinner, you're just getting my raw reactions!

tests/src/main/java/com/squareup/duktape/tests/OctaneActivity.java
JakeWharton: Yeah it didn't do this on my test devices. It's easy to add though, and probably good to show some indication of work being done.

JakeWharton:
screen shot 2015-11-11 at 11 22 35 pm
screen shot 2015-11-11 at 11 31 55 pm

swankjesse:
Awesome

PR #31, opened 2016-02-08
Add function that binds a Java object to a named JavaScript global an…

szurbrigg:
…d allow it to be called from JavaScript.  Supports Strings and void argument/return types (will add more in a follow up).


swankjesse:
ultranit: nice commit messages
https://git-scm.com/book/ch5-2.html

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
...
 
   /**
+   * Binds {@code object} to {@code name} for use in JavaScript as a global object.  {@type} defines
+   * the interface implemented by {@code object} that will be accessible to JavaScript.
+   */
+  public <T> void bind(Class<T> type, String name, T object) {
+    if (!type.isInterface()) {
+      throw new UnsupportedOperationException("Only interfaces can be bound.");
+    }
+    if (type.getInterfaces().length > 0) {
```
swankjesse: awesome

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
   }
 
   /**
+   * Binds {@code object} to {@code name} for use in JavaScript as a global object.  {@type} defines
+   * the interface implemented by {@code object} that will be accessible to JavaScript.
+   */
+  public <T> void bind(Class<T> type, String name, T object) {
```
JakeWharton: I would flip the first two params here so the type and instance are next to each other. Plus that puts the name right after "bind" so it reads well:
```
duktape.bind("logger", Logger.class, logger);
```

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
...
+    if (!type.isInterface()) {
+      throw new UnsupportedOperationException("Only interfaces can be bound.");
```
JakeWharton: nit: include type in message
```
"Only interfaces can be bound. Received: " + type
```

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
...
+    }
+    if (type.getInterfaces().length > 0) {
+      throw new UnsupportedOperationException("Bound interfaces cannot extend other interfaces.");
```
JakeWharton: nit: include type in message
```
"Bound interface " + type + " must not extend other interfaces."
```

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
...
+    }
+    if (!type.isInstance(object)) {
+      throw new IllegalArgumentException("object is not an instance of the given type.");
```
JakeWharton: nit: include `type` and `object.getClass()` in message

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
...
+    }
+    HashMap<String, Method> methods = new HashMap<>();
```
swankjesse: Always LinkedHashMap. Never HashMap. I even wrote a long story about why.
https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -80,9 +106,30 @@ public String evaluate(String script) {
...
+      if (!isSupportedArgType(parameterType)) {
+        throw new UnsupportedOperationException(
+            String.format("Parameter type %s on %s is not supported",
+                parameterType.toString(), method.getName()));
+      }
+    }
+  }
+
+  private static boolean isSupportedArgType(Class<?> parameterType) {
+    // TODO: support some more types.
```
swankjesse: Weird to have a method named `arg type` with `void`, which can’t be used as an arg. (only a return type).

swankjesse: Not sure what else to call this. isSupportedType ?

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "DuktapeContext.h"
+#include "JString.h"
+#include "JavaMethod.h"
+
+namespace {
+
+// Internal names used to properties in the duktape context's global stash and bound variables.
```
swankjesse: used to properties –> used _for_ properties?

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+// The \xff\xff part keeps the variable hidden from JavaScript (visible through C API only).
+const char* JAVAVM_PROP_NAME = "\xff\xffjavaVM";
+const char* JAVA_THIS_PROP_NAME = "\xff\xffjava_this";
+const char* JAVA_METHOD_PROP_NAME = "\xff\xffjava_method";
+
+JNIEnv* getJNIEnv(duk_context *ctx) {
```
swankjesse: cute

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+  duk_push_global_stash(ctx);
+  duk_get_prop_string(ctx, -1, JAVAVM_PROP_NAME);
+  JavaVM* javaVM = static_cast<JavaVM*>(duk_require_pointer(ctx, -1));
+  duk_pop_2(ctx);
+  JNIEnv* env;
+  javaVM->AttachCurrentThread(&env, nullptr);
+  return env;
+}
+
+jobject getJavaThis(duk_context* ctx) {
```
swankjesse: this is really neat

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+duk_ret_t handler(duk_context *ctx) {
+  JavaMethod* method = getJavaMethod(ctx);
+  return method ? method->Invoke(ctx, getJNIEnv(ctx), getJavaThis(ctx)) : DUK_RET_INTERNAL_ERROR;
+}
+
+duk_ret_t javaObjectFinalizer(duk_context *ctx) {
+  jobject thisObject = static_cast<jobject>(duk_require_pointer(ctx, -1));
+  getJNIEnv(ctx)->DeleteGlobalRef(thisObject);
+  duk_enum(ctx, -1, DUK_ENUM_OWN_PROPERTIES_ONLY);
```
swankjesse: just curious - what does this do?

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+  while (duk_next(ctx, -1, true)) {
+    if (duk_get_prop_string(ctx, -1, JAVA_METHOD_PROP_NAME)) {
+      delete static_cast<JavaMethod*>(duk_require_pointer(ctx, -1));
```
swankjesse: does this undo the ref?

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+  } else {
+    result = env->NewStringUTF(duk_get_string(m_context, -1));
+  }
+
+  duk_pop(m_context);
+
+  return result;
+}
+
+void Duktape::bindInstance(JNIEnv *env, jstring instance, jobject object, jobjectArray methods) {
```
swankjesse: You need to declare this after javaObjectFinalizer? As Java programmer I like supporting methods beneath the methods that use ’em, but I don’t think that’s idiomatic here.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+  duk_push_global_object(m_context);
+  const duk_idx_t objIndex = duk_require_normalize_index(m_context, duk_push_object(m_context));
+
+  // Keep a reference in JavaScript to the object being bound.
+  duk_push_pointer(m_context, env->NewGlobalRef(object));
+  duk_put_prop_string(m_context, objIndex, JAVA_THIS_PROP_NAME);
+  duk_push_c_function(m_context, javaObjectFinalizer, 1);
+  duk_set_finalizer(m_context, objIndex);
```
swankjesse: this is really cool.

duktape/src/main/jni/DuktapeContext.h
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef DUKTAPE_ANDROID_DUKTAPE_CONTEXT_H
+#define DUKTAPE_ANDROID_DUKTAPE_CONTEXT_H
+
+#include <jni.h>
+#include "duktape.h"
+
+class Duktape {
```
swankjesse: Tough one. Should this file be called Duktape.h or should this be called DuktapeContext ?

duktape/src/main/jni/JString.h
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef DUKTAPE_ANDROID_JSTRING_H
+#define DUKTAPE_ANDROID_JSTRING_H
+
+class JString {
```
swankjesse: neat

duktape/src/main/jni/JavaMethod.h
```diff
@@ -0,0 +1,48 @@
...
+
+  JavaMethod(JNIEnv* env, jobject method);
+
+  duk_ret_t Invoke(duk_context* ctx, JNIEnv* env, jobject javaThis) const;
+
+  const char* getName() const {
+    return m_name.c_str();
+  }
+
+  enum Type {
```
swankjesse: top-level type? Does JNI give us anything useful here? Seems like this wants to be something other than an enum

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -0,0 +1,132 @@
...
+ * Convert the duktape value at the top of the stack to the given Java type.
+ * throws a JavaScript TypeError if the JS type is not {@code type}.
+ */
+jvalue popJavaType(JNIEnv* env, duk_context* ctx, JavaMethod::Type type) {
+  const duk_idx_t index = -1;
+  jvalue value;
+  switch (type) {
+    default:
+    case JavaMethod::Void:
+      duk_error(ctx, DUK_ERR_API_ERROR, "unsupported argument type: %d", type);
```
swankjesse: need this break or return?

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
...
+      throw new UnsupportedOperationException("Bound interfaces cannot extend other interfaces.");
+    }
+    if (!type.isInstance(object)) {
+      throw new IllegalArgumentException("object is not an instance of the given type.");
+    }
+    HashMap<String, Method> methods = new HashMap<>();
+    for (Method method : type.getMethods()) {
+      checkSignatureSupported(method);
+      if (methods.put(method.getName(), method) != null) {
+        throw new UnsupportedOperationException("Method overloads are not supported.");
```
JakeWharton: nit: include method name in exception

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -0,0 +1,132 @@
...
+ */
+duk_ret_t pushJavaScriptType(JNIEnv* env, duk_context *ctx, JavaMethod::Type type, jvalue value) {
+  switch (type) {
+    case JavaMethod::Void:
+      return 0;
+
+    case JavaMethod::String: {
+      const JString result(env, static_cast<jstring>(value.l));
+      duk_push_string(ctx, result);
+      return 1;
```
swankjesse: what does this mean?

swankjesse: ie. what’s the rule? Number of pushes?

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    default:
+      duk_error(ctx, DUK_ERR_API_ERROR, "unsupported return type: %d", type);
```
swankjesse: break?!

tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015 Square, Inc.
+ * Copyright (C) 2016 Square, Inc.
```
swankjesse: For what it’s worth, we typically set these at file-creation time and leave them as-is til the end of time.

tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
```diff
@@ -86,4 +86,121 @@
...
+      assertThat(expected).hasMessage("Only interfaces can be bound.");
+    }
+  }
+
+  @Test
+  public void callMethodOnJavaObject() {
+    duktape.bind(TestInterface.class, "value", new TestInterface() {
+      @Override
+      public String getValue() {
+        return "8675309";
```
swankjesse: 📞

swankjesse:
Really awesome. I’ve got lots of comments, but it’s all nitpicks and questions about how it all works.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Agreed.  Done.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Done.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Done.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Done.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Done.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Works for me. Done.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Nice catch.  Fixed.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Yes, it's supposed to be called when duktape's GC runs and there are no more JS references to our bound Java object.  It's hooked up below (see duk_set_finalizer).

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Yes, I'd still have to forward declare the javaObjectFinalizer above the use.  If it's all the same, I'd personally rather keep it as is.

duktape/src/main/jni/DuktapeContext.h
szurbrigg: I initially had Duktape.h, but that collides with duktape.h from the duktape project.   Renaming the class to DuktapeContext is a little annoying too, but I guess it makes sense.

duktape/src/main/jni/JavaMethod.h
szurbrigg: Agreed.  I'm still thinking about how to handle this.  I think I want to create a JavaValue class (maybe with concrete subclasses like JavaVoid, JavaString, etc.) to complement JavaMethod.  I'll use that to abstract away marshalling/unmarshalling different types.

I'm open to suggestions. :-)

duktape/src/main/jni/JavaMethod.cpp
szurbrigg: duk_error never returns.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Done.

duktape/src/main/jni/JavaMethod.cpp
szurbrigg: Yes, the return value to duktape is the number of entries pushed to the stack.  I'll add a comment.

duktape/src/main/jni/JavaMethod.cpp
szurbrigg: duk_error never returns.  But I'll add a comment in both places to say so.

tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
szurbrigg: Noted and reverted.

duktape/src/main/jni/JavaMethod.h
szurbrigg: Will improve in a follow up when I add more supported types.

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -63,6 +65,30 @@ public String evaluate(String script) {
   }
 
   /**
+   * Binds {@code object} to {@code name} for use in JavaScript as a global object. {@type} defines
+   * the interface implemented by {@code object} that will be accessible to JavaScript.
```
swankjesse: Please document the other limitations: no overloading, no extending interfaces

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -0,0 +1,184 @@
...
+  duk_pop_2(ctx);
+
+  JNIEnv* env;
+  javaVM->AttachCurrentThread(&env, nullptr);
+  return env;
+}
+
+jobject getJavaThis(duk_context* ctx) {
+  duk_push_this(ctx);
+  if (!duk_get_prop_string(ctx, -1, JAVA_THIS_PROP_NAME)) {
```
swankjesse: will this ever happen ?

swankjesse: errr.  . . will we ever not have a this pointer?

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Done.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Not possible at the moment, no.  Might be a bit over-defensive, I'll undo these checks.

swankjesse:
LGTM

PR #33, opened 2016-02-12
Support boolean, int, and double Java types.

szurbrigg:
Update the bind function to allow these primitive types when calling Java methods.

### DIFF
--- a/duktape/src/main/java/com/squareup/duktape/Duktape.java
+++ b/duktape/src/main/java/com/squareup/duktape/Duktape.java
@@ -109,13 +109,14 @@ public final class Duktape implements Closeable {
   }
 
   private static void checkSignatureSupported(Method method) {
-    if (!isSupportedType(method.getReturnType())) {
+    if (!isSupportedParameterType(method.getReturnType())
+        && !void.class.equals(method.getReturnType())) {
       throw new UnsupportedOperationException(
           String.format("Return type %s on %s is not supported",
               method.getReturnType().toString(), method.getName()));
     }
     for (Class<?> parameterType : method.getParameterTypes()) {
-      if (!isSupportedType(parameterType)) {
+      if (!isSupportedParameterType(parameterType)) {
         throw new UnsupportedOperationException(
             String.format("Parameter type %s on %s is not supported",
                 parameterType.toString(), method.getName()));
@@ -123,10 +124,12 @@ public final class Duktape implements Closeable {
     }
   }
 
-  /** Returns true if we support {@code: type} in calls from JavaScript. */
-  private static boolean isSupportedType(Class<?> type) {
-    // TODO: support some more types.
-    return void.class.equals(type) || String.class.equals(type);
+  /** Returns true if we support {@code: type} as parameters in calls from JavaScript. */
+  private static boolean isSupportedParameterType(Class<?> type) {
+    return boolean.class.equals(type)
+        || int.class.equals(type)
+        || double.class.equals(type)
+        || String.class.equals(type);
   }
 
   private static native long createContext();
--- a/duktape/src/main/jni/DuktapeContext.cpp
+++ b/duktape/src/main/jni/DuktapeContext.cpp
@@ -155,7 +155,8 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring instance, jobject object,
 
   const jsize numMethods = env->GetArrayLength(methods);
   for (jsize i = 0; i < numMethods; ++i) {
-    JavaMethod* javaMethod = new JavaMethod(env, env->GetObjectArrayElement(methods, i));
+    jobject method = env->GetObjectArrayElement(methods, i);
+    JavaMethod* javaMethod = new JavaMethod(env, method);
 
     // Use VARARGS here to allow us to manually validate that the proper number of arguments are
     // given in the call.  If we specify the actual number of arguments needed, duktape will try to
@@ -165,7 +166,11 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring instance, jobject object,
     duk_push_pointer(m_context, javaMethod);
     duk_put_prop_string(m_context, func, JAVA_METHOD_PROP_NAME);
 
-    duk_put_prop_string(m_context, objIndex, javaMethod->getName());
+    const jmethodID getName =
+        env->GetMethodID(env->GetObjectClass(method), "getName", "()Ljava/lang/String;");
+    const JString methodName(env, static_cast<jstring>(env->CallObjectMethod(method, getName)));
+
+    duk_put_prop_string(m_context, objIndex, methodName);
   }
 
   const JString instanceName(env, instance);
--- a/duktape/src/main/jni/JavaMethod.cpp
+++ b/duktape/src/main/jni/JavaMethod.cpp
@@ -14,124 +14,168 @@
  * limitations under the License.
  */
 #include "JavaMethod.h"
-#include <algorithm>
 #include "JString.h"
 
 namespace {
 
-JavaMethod::Type getType(JNIEnv* env, jobject typeObject) {
-  if (env->IsSameObject(typeObject, env->FindClass("java/lang/String"))) {
-    return JavaMethod::String;
-  }
-  // TODO: support some more types.
-  return JavaMethod::Void;
-}
-
-/**
- * Convert the duktape value at the top of the stack to the given Java type.
- * throws a JavaScript TypeError if the JS type is not {@code type}.
- */
-jvalue popJavaType(JNIEnv* env, duk_context* ctx, JavaMethod::Type type) {
-  const duk_idx_t index = -1;
-  jvalue value;
-  switch (type) {
-    case JavaMethod::String:
-      // Check if the caller passed in a null string.
-      value.l = duk_get_type(ctx, index) != DUK_TYPE_NULL
-                ? env->NewStringUTF(duk_require_string(ctx, index))
-                : nullptr;
-      break;
-
-    case JavaMethod::Void:
-    default:
-      duk_error(ctx, DUK_ERR_API_ERROR, "unsupported argument type: %d", type);
-      // unreachable - duk_error never returns.
-      break;
-  }
-  duk_pop(ctx);
-  return value;
-}
-
-/**
- * Convert the given {@code value} to a duktape type and push it onto the stack (for a function
- * call return).
- * Returns the number of entries pushed to the duktape stack.
- */
-duk_ret_t pushJavaScriptType(JNIEnv* env, duk_context *ctx, JavaMethod::Type type, jvalue value) {
-  switch (type) {
-    case JavaMethod::Void:
-      return 0;
-
-    case JavaMethod::String: {
-      const JString result(env, static_cast<jstring>(value.l));
-      duk_push_string(ctx, result);
-      return 1;
-    }
-
-    default:
-      duk_error(ctx, DUK_ERR_API_ERROR, "unsupported return type: %d", type);
-      // unreachable - duk_error never returns.
-      return DUK_RET_INTERNAL_ERROR;
-  }
-}
+JavaMethod::ArgumentLoader getArgumentLoader(JNIEnv *env, jobject typeObject);
+JavaMethod::MethodBody getMethodBody(JNIEnv* env, jmethodID methodId, jobject returnType);
 
 } // anonymous namespace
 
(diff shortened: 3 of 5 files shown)


duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -125,8 +125,11 @@ private static void checkSignatureSupported(Method method) {
 
   /** Returns true if we support {@code: type} in calls from JavaScript. */
   private static boolean isSupportedType(Class<?> type) {
-    // TODO: support some more types.
-    return void.class.equals(type) || String.class.equals(type);
+    return void.class.equals(type)
+        || boolean.class.equals(type)
+        || int.class.equals(type)
+        || double.class.equals(type)
```
JakeWharton: We'll probably also want to do the boxed versions of these 3 in a follow-up.

JakeWharton: Actually we probably want to get `long`, `float, `short`, and `char` in here as well, right? At the very least the first two. Also could be a follow-up. Maybe I'll even attempt it!

szurbrigg: For sure.

szurbrigg: Yeah, we could support them all.  I'm not a JS expert, but we'd probably have to do our own limit checking.

e.g. IIRC JavaScript tops out at 53 bit integers or something, so a Java long might be truncated.

swankjesse: Yeah. If a number can be represented exactly as a Java double, the conversion is lossless. Otherwise it’s lossy. The easiest way to check is to round trip cast & compare.

JakeWharton:
Very cool. Don't know enough to review the important bits!

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -165,7 +166,11 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring instance, jobject object,
     duk_push_pointer(m_context, javaMethod);
     duk_put_prop_string(m_context, func, JAVA_METHOD_PROP_NAME);
 
-    duk_put_prop_string(m_context, objIndex, javaMethod->getName());
```
swankjesse: Just curious: why this change?

duktape/src/main/jni/JavaMethod.h
```diff
@@ -33,20 +37,23 @@ class JavaMethod {
...
-  const char* getName() const {
-    return m_name.c_str();
-  }
-
-  enum Type {
-    Void,
-    String
-  };
+  /**
+   * Defines a functor to use to pop a value off the duktape stack and convert it to the required
```
swankjesse: Nice.

duktape/src/main/jni/JavaMethod.h
```diff
@@ -33,20 +37,23 @@ class JavaMethod {
...
+   * Java type.
+   */
+  typedef std::function<jvalue(duk_context*, JNIEnv*)> ArgumentLoader;
+  /**
+   * Defines a functor to invoke the correct JNI method that will return the required Java type,
+   * convert the return value to a JavaScript type and push it to the duktape stack.  Returns the
+   * number of entries pushed to the stack.
+   * If the Java method throws an exception, the functor will throw a duktape error with the
```
swankjesse: Interesting!

duktape/src/main/jni/JavaMethod.h
```diff
@@ -33,20 +37,23 @@ class JavaMethod {
...
-  };
+  /**
+   * Defines a functor to use to pop a value off the duktape stack and convert it to the required
+   * Java type.
+   */
+  typedef std::function<jvalue(duk_context*, JNIEnv*)> ArgumentLoader;
+  /**
+   * Defines a functor to invoke the correct JNI method that will return the required Java type,
+   * convert the return value to a JavaScript type and push it to the duktape stack.  Returns the
+   * number of entries pushed to the stack.
```
swankjesse: The lack of a return type in the typedef confuses my Java-generics eyes

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -19,119 +19,160 @@
...
+
+  jmethodID methodId = env->FromReflectedMethod(method);
+  m_methodBody = getMethodBody(env, methodId, returnType);
 }
 
-duk_ret_t JavaMethod::invoke(duk_context *ctx, JNIEnv *env, jobject javaThis) const {
-  if (duk_get_top(ctx) != m_parameterTypes.size()) {
+duk_ret_t JavaMethod::invoke(duk_context* ctx, JNIEnv* env, jobject javaThis) const {
+  if (duk_get_top(ctx) != m_argumentLoaders.size()) {
     duk_error(ctx, DUK_ERR_API_ERROR, "wrong number of arguments");
```
swankjesse: does this pop the top? Should it?

swankjesse: should it also return?

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -19,119 +19,160 @@
...
-      returnValue.l = env->CallObjectMethodA(javaThis, m_methodId, args.data());
-      break;
-
-    default:
-      return DUK_RET_INTERNAL_ERROR;
+
+  std::vector<jvalue> args(m_argumentLoaders.size());
+  // Load the arguments off the stack and convert to java types.
+  // Note we're going backwards since the last argument is at the top of the stack.
+  std::transform(m_argumentLoaders.rbegin(), m_argumentLoaders.rend(), args.rbegin(),
```
swankjesse: If this was a Java transform I’d complain that a loop would be simpler. I’m in no place to say what’s idiomatic or not in C++, but this seems like it’s unnecessarily fancy

swankjesse: (I think the threshold for me is that you need to define a signature and a block . . . if you could pass the method reference alone I think the transform call would win)

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -19,119 +19,160 @@
...
+                 });
+
+  return m_methodBody(ctx, env, javaThis, args.data());
+}
+
+namespace {
+
+/**
+ * Loads the (primitive) TYPE member of {@code boxedClassName}.
+ * For example, given java/lang/Integer, this function will return int.class.
```
swankjesse: Nice docs.

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -19,119 +19,160 @@
...
+ * Get a functor that pops the object off the top of the stack and converts it to an instance of
+ * {@code typeObject}.  Duktape may throw if the argument is not the correct type.
+ */
+JavaMethod::ArgumentLoader getArgumentLoader(JNIEnv *env, jobject typeObject) {
+  if (env->IsSameObject(typeObject, env->FindClass("java/lang/String"))) {
+    return [](duk_context* ctx, JNIEnv* jniEnv) {
+      jvalue value;
+      // Check if the caller passed in a null string.
+      value.l = duk_get_type(ctx, -1) != DUK_TYPE_NULL
+                ? jniEnv->NewStringUTF(duk_require_string(ctx, -1))
```
swankjesse: this is a great pattern

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -19,119 +19,160 @@
...
+      jvalue value;
+      value.z = duk_require_boolean(ctx, -1);
+      duk_pop(ctx);
+      return value;
+    };
   }
+  if (env->IsSameObject(typeObject, getPrimitiveType(env, "java/lang/Integer"))) {
+    return [](duk_context* ctx, JNIEnv*) {
+      jvalue value;
+      value.i = duk_require_int(ctx, -1);
```
swankjesse: wow, duktape really helps here!

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -19,119 +19,160 @@
...
+      const JString result(jniEnv, static_cast<jstring>(returnValue));
+      duk_push_string(ctx, result);
+      return 1;
+    };
+  }
+  if (env->IsSameObject(returnType, getPrimitiveType(env, "java/lang/Boolean"))) {
+    return [methodId](duk_context* ctx, JNIEnv* jniEnv, jobject javaThis, jvalue* args) {
+      jboolean returnValue = jniEnv->CallBooleanMethodA(javaThis, methodId, args);
+      checkRethrowException(ctx, jniEnv);
+      duk_push_boolean(ctx, returnValue == JNI_TRUE);
```
swankjesse: cute

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -19,119 +19,160 @@
...
+  }
+  if (env->IsSameObject(returnType, getPrimitiveType(env, "java/lang/Double"))) {
+    return [methodId](duk_context* ctx, JNIEnv* jniEnv, jobject javaThis, jvalue* args) {
+      jdouble returnValue = jniEnv->CallDoubleMethodA(javaThis, methodId, args);
+      checkRethrowException(ctx, jniEnv);
+      duk_push_number(ctx, returnValue);
+      return 1;
+    };
+  }
+  return [methodId](duk_context* ctx, JNIEnv* jniEnv, jobject javaThis, jvalue* args) {
```
swankjesse: should this guard for void? Seems safer

tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
```diff
@@ -192,4 +194,46 @@ public String getValue() {
...
+      }
+      @Override public double d(double d) {
+        return d / 2.0;
+      }
+    });
+
+    // TODO: add an evaluate interface that supports other types.
+    assertThat(duktape.evaluate("value.b(false).toString()")).isEqualTo("true");
+    assertThat(duktape.evaluate("value.i(4).toString()")).isEqualTo("16");
+    assertThat(duktape.evaluate("value.d(6.28318).toString()")).isEqualTo("3.14159");
```
swankjesse: Tau!

swankjesse:
I like the code structure here. Straightforward & obvious.

I’d like you to add test cases for exceptions thrown by Java.

swankjesse:
LGTM

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: The name was only being used here.  I don't want to store an unused copy of it forever in the JavaMethod object.

duktape/src/main/jni/JavaMethod.cpp
szurbrigg: `duk_error` causes a duktape JS error object to be thrown, so the stack will unwind automatically. The function never returns, but I'll add a return statement for clarity (like I've done elsewhere).

duktape/src/main/jni/JavaMethod.cpp
szurbrigg: Yeah, the trick is that we're iterating over the vector backwards, since we're popping the last java argument first, and using `rbegin` in a transform call seemed more straightforward at the time than an indexed loop counting down.  I agree, though, it's a little cumbersome.  I can make it a for loop with a numeric index.

duktape/src/main/jni/JavaMethod.cpp
szurbrigg: Sure, will do.

szurbrigg:
There's already test for Java exceptions now: `DuktapeTest.callMethodOnJavaObjectThrowsJavaException()`.  Was there something else to do with Java exceptions that needs a test?

duktape/src/main/jni/JavaMethod.cpp
swankjesse: would it work with a loop and an iterator?

szurbrigg: We'd need two iterators.  We're reading from m_argumentLoaders and writing to args.   I'm thinking of something like this:
```c++
for (ssize_t i = m_argumentLoaders.size() - 1; i >= 0; --i) {
  args[i] = m_argumentLoaders[i](ctx, env);
}
```

PR #34, opened 2016-02-22
Support boxed primitive types in Java calls.

szurbrigg:


duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -76,6 +86,46 @@ jobject getPrimitiveType(JNIEnv* env, const char* boxedClassName) {
...
+  // Check if the caller passed in a null value.
+  if (duk_get_type(ctx, -1) == DUK_TYPE_NULL) {
+    duk_pop(ctx);
+    jvalue value;
+    value.l = nullptr;
+    return value;
+  }
+
+  jvalue value = popPrimitive(ctx, env);
+  value.l = env->NewObjectA(type, constructor, &value);
```
JakeWharton: Can we call the static `valueOf` method that all boxed types have here? That way we leverage their individual caches for small/some values instead of actually allocating each time.

JakeWharton:
Very cool

duktape/src/main/jni/JavaMethod.cpp
szurbrigg: Yes, excellent idea.  Will do.

duktape/src/main/jni/GlobalRef.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+GlobalRef::GlobalRef(const GlobalRef& other)
+    : m_javaVM(other.m_javaVM)
+    , m_object(getEnvFromJavaVM(m_javaVM)->NewGlobalRef(other.m_object)) {
+}
+
+GlobalRef& GlobalRef::operator=(GlobalRef& other) {
+  if (&other != this) {
+    // Increment the new reference before decrementing the old.
```
swankjesse: Good comment!

duktape/src/main/jni/GlobalRef.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+JNIEnv* getEnvFromJavaVM(JavaVM* javaVM) {
+  if (javaVM == nullptr) {
+    return nullptr;
+  }
+
+  JNIEnv* env;
+  javaVM->AttachCurrentThread(
+#ifdef __ANDROID__
```
swankjesse: ew, right?

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -65,6 +67,14 @@ duk_ret_t JavaMethod::invoke(duk_context* ctx, JNIEnv* env, jobject javaThis) co
 
 namespace {
 
+/** Calls toString() on the given object and returns a copy of the result. */
```
swankjesse: cute

duktape/src/main/jni/JavaMethod.cpp
```diff
@@ -91,28 +141,57 @@ JavaMethod::ArgumentLoader getArgumentLoader(JNIEnv *env, jobject typeObject) {
...
   }
-  assert(env->IsSameObject(typeObject, getPrimitiveType(env, "java/lang/Double")));
-  return [](duk_context* ctx, JNIEnv*) {
-    jvalue value;
-    value.d = duk_require_number(ctx, -1);
-    duk_pop(ctx);
-    return value;
+
+  if (env->IsSameObject(typeObject, getPrimitiveType(env, "java/lang/Double"))) {
+    return popDouble;
```
swankjesse: where's the box?

swankjesse: ahh, it's `getPrimitiveType` not `FindClass()`. All good.

swankjesse:
LGTM!

duktape/src/main/jni/GlobalRef.cpp
szurbrigg: Yes.  Both the Mac and Linux jni.h declare that argument as a void**, which is disappointing.

PR #35, opened 2016-02-23
Throw C++ exceptions for binding errors.

szurbrigg:
- Exceptions are translated to Java exceptions and rethrown to the Java caller.
- Make sure objects and memory are freed up when bad things happen.

### DIFF
--- a/duktape/src/main/jni/DuktapeContext.cpp
+++ b/duktape/src/main/jni/DuktapeContext.cpp
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 #include "DuktapeContext.h"
+#include <memory>
+#include <string>
+#include <stdexcept>
 #include "JString.h"
 #include "JavaMethod.h"
 #include "GlobalRef.h"
@@ -72,11 +75,12 @@ duk_ret_t javaObjectFinalizer(duk_context *ctx) {
   }
 
   // Remove the global reference from the bound java object.
-  jobject thisObject = static_cast<jobject>(duk_require_pointer(ctx, -1));
-  getJNIEnv(ctx)->DeleteGlobalRef(thisObject);
+  getJNIEnv(ctx)->DeleteGlobalRef(static_cast<jobject>(duk_require_pointer(ctx, -1)));
+  duk_pop(ctx);
+  duk_del_prop_string(ctx, -1, JAVA_METHOD_PROP_NAME);
 
   // Iterate over all of the properties, deleting all the JavaMethod objects we attached.
-  duk_enum(ctx, -2, DUK_ENUM_OWN_PROPERTIES_ONLY);
+  duk_enum(ctx, -1, DUK_ENUM_OWN_PROPERTIES_ONLY);
   while (duk_next(ctx, -1, true)) {
     if (!duk_get_prop_string(ctx, -1, JAVA_METHOD_PROP_NAME)) {
       duk_pop_2(ctx);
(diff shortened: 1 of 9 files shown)


tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
```diff
@@ -127,6 +127,24 @@ public String getValue() {
...
+    });
+    try {
+      duktape.bind("value", TestInterface.class, new TestInterface() {
+        @Override public String getValue() {
+          return "bar";
+        }
+      });
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected.getMessage()).isEqualTo("A global object called value already exists");
```
JakeWharton: nit: can do `assertThat(expected).hasMessage` and save some chars

szurbrigg: Done.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -141,39 +150,61 @@ jstring DuktapeContext::evaluate(JNIEnv *env, jstring code, jstring fname) {
...
-    JavaMethod* javaMethod = new JavaMethod(env, method);
+
+    const jmethodID getName =
+        env->GetMethodID(env->GetObjectClass(method), "getName", "()Ljava/lang/String;");
+    const JString methodName(env, static_cast<jstring>(env->CallObjectMethod(method, getName)));
+
+    std::unique_ptr<JavaMethod> javaMethod;
+    try {
+      javaMethod.reset(new JavaMethod(env, method));
+    } catch (std::invalid_argument& invalidArgument) {
```
swankjesse: neat! Never seen C++ exceptions before

swankjesse: this one surprises me though . . . why try/catch to test the argument rather than using an if? In Java code catching an IllegalArgumentException is almost always a bad idea

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -141,39 +150,61 @@ jstring DuktapeContext::evaluate(JNIEnv *env, jstring code, jstring fname) {
...
-    const jmethodID getName =
-        env->GetMethodID(env->GetObjectClass(method), "getName", "()Ljava/lang/String;");
-    const JString methodName(env, static_cast<jstring>(env->CallObjectMethod(method, getName)));
-
+    // Add this method to the bound object.
     duk_put_prop_string(m_context, objIndex, methodName);
   }
 
-  const JString instanceName(env, instance);
+  // Make our bound java object a property of the duktape global object (so it's a JS global).
```
swankjesse: supernit: java –> Java

swankjesse: Also I think duktape wants to be Duktape in comments

duktape/src/main/jni/DuktapeContext.cpp
swankjesse: ahhh, I think this is one of those cases where it does make sense. You’re not catching the exception, you're just converting it.

swankjesse:
LGTM!

swankjesse:
One other thing to consider for follow up - you’ve got really nice exception messages when the types aren't supported. You should validate that those messages are formatted as expected in your tests.

(Otherwise sometimes exceptions are missing spaces like `Unexpected methodtoString()` or similar

PR #37, opened 2016-03-06
Include JavaScript stack when Java throws.

szurbrigg:
- If a call to a bound Java method throws an exception, include intermediate JavaScript stack trace elements in the exception.

### DIFF
--- a/duktape/src/main/java/com/squareup/duktape/DuktapeException.java
+++ b/duktape/src/main/java/com/squareup/duktape/DuktapeException.java
@@ -34,21 +34,40 @@ public final class DuktapeException extends RuntimeException {
 
   public DuktapeException(String detailMessage) {
     super(getErrorMessage(detailMessage));
+    addDuktapeStack(this, detailMessage);
+  }
+
+  /**
+   * Parses {@code StackTraceElement}s from {@code detailMessage} and adds them to the proper place
+   * in {@code throwable}'s stack trace.  Note: this method is also called from native code.
+   */
+  static void addDuktapeStack(Throwable throwable, String detailMessage) {
     String[] lines = detailMessage.split("\n", -1);
     if (lines.length <= 1) {
       return;
     }
-    // We have a stacktrace following the message.  Add it to this exception.
+    // We have a stacktrace following the message.  Add it to the exception.
     List<StackTraceElement> elements = new ArrayList<>();
-    for (int i = 1; i < lines.length; ++i) {
-      StackTraceElement stackTraceElement = toStackTraceElement(lines[i]);
-      if (stackTraceElement == null) {
-        continue;
+
+    // Splice the JavaScript stack in right above the call to Duktape.evaluate.
+    boolean spliced = false;
+    for (StackTraceElement stackTraceElement : throwable.getStackTrace()) {
+      if (!spliced
+          && stackTraceElement.isNativeMethod()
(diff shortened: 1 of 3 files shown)


duktape/src/main/java/com/squareup/duktape/DuktapeException.java
```diff
@@ -34,21 +34,36 @@
 
   public DuktapeException(String detailMessage) {
     super(getErrorMessage(detailMessage));
+    addDuktapeStack(this, detailMessage);
+  }
+
+  public static void addDuktapeStack(Throwable throwable, String detailMessage) {
```
swankjesse: package-private ?

tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
```diff
@@ -146,6 +146,57 @@ public String getValue() {
...
+          + "  value.getValue();\n"       // Line 10.
+          + "}\n", "test.js");
+      fail();
+    } catch (UnsupportedOperationException e) {
+      assertThat(e).hasMessage("Cannot getValue");
+
+      StackTraceElement[] stackTrace = e.getStackTrace();
+
+      // Top entry is what threw - the TestInterface implementation.
+      assertThat(stackTrace[0].getClassName()).isEqualTo(boundObject.getClass().getName());
```
swankjesse: might be more readable to just dump the stacktrace to a string & compare that? At a minimum that format is very familiar. 

```
  static String toStacktrace(Throwable t) {
    StringWriter writer = new StringWriter();
    t.printStackTrace(new PrintWriter(writer));
    return writer.toString();
  }
```

swankjesse: (though I guess the drawback of that is you also get the JUnit runtime stuff. So um, nevermind!)

tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
```diff
@@ -146,6 +146,57 @@ public String getValue() {
...
+      assertThat(stackTrace[4].getClassName()).isEqualTo(Duktape.class.getName());
+      assertThat(stackTrace[4].getMethodName()).isEqualTo("evaluate");
+      assertThat(stackTrace[4].isNativeMethod()).isTrue();
+
+      // Then the Java method.
+      assertThat(stackTrace[5].getClassName()).isEqualTo(Duktape.class.getName());
+      assertThat(stackTrace[5].getMethodName()).isEqualTo("evaluate");
+      assertThat(stackTrace[5].isNativeMethod()).isFalse();
+
+      // Then this test method.
```
swankjesse: Inception!

swankjesse:
LGTM

duktape/src/main/java/com/squareup/duktape/DuktapeException.java
szurbrigg: Good call.  Done.

tests/src/androidTest/java/com/squareup/duktape/DuktapeTest.java
szurbrigg: I had initially done that, and ran into just that.  There's also the ugliness/brittleness of the line numbers that show up in the printed stack trace.

PR #46, opened 2016-04-10
Initial implementation of proxy support.

szurbrigg:
- support a proxy method that allows Java to call a global JavaScript
  object.
- still a work in progress: no argument support, only strings for return
  values.
- misc: split up the tests into separate fixtures by functional group.


duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -95,6 +97,33 @@ public String evaluate(String script) {
...
   /**
+   * Creates a proxy to a global JavaScript object called {@code name} that implements {@code type}.
+   * {@code type} defines the interface implemented in JavaScript that will be accessible to Java.
+   * {@code type} must be an interface that does not extend any other interfaces, and cannot define
+   * any overloaded methods.
+   * <p>
+   * Methods of the interface may return {@code void} or any of the following supported argument
+   * types: {@code boolean}, {@link Boolean}, {@code int}, {@link Integer}, {@code double},
+   * {@link Double}, {@link String}.
+   * </p>
```
swankjesse: you don’t need to close `<p>` tags. That’s an XHTML thing, and XHTML is dead

swankjesse: also, if you put the `<p>` on the same line as `Methods`, then IntelliJ is able to reformat automatically. That’s what we do in our other HTML Javadocs.

duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -95,6 +97,33 @@ public String evaluate(String script) {
...
+    if (!type.isInterface()) {
+      throw new UnsupportedOperationException("Only interfaces can be proxied. Received: " + type);
+    }
+    if (type.getInterfaces().length > 0) {
+      throw new UnsupportedOperationException(type + " must not extend other interfaces");
+    }
+    return (T) Proxy.newProxyInstance(type.getClassLoader(), new Class<?>[]{ type },
+        new InvocationHandler() {
+      @Override
+      public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
```
swankjesse: Can you check if `method`’s declaring class is `type`? All interfaces inherit the methods of Object.class and when that happens your best bet is to do say `return method.invoke(InvocationHandler.this, args)` instead of calling out to JavaScript.

(and then if you also want, you can override `toString()` in this invocation handler to return something like `"proxy(value, FooInterface.class)"`

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -93,6 +93,45 @@ duk_ret_t javaObjectFinalizer(duk_context *ctx) {
...
+    }
+    // Pop the stack text.
+    duk_pop(ctx);
+  } else {
+    // Not an error or no stacktrace, just convert to a string.
+    env->ThrowNew(exceptionClass, duk_safe_to_string(ctx, -1));
+  }
+}
+
+void throwNoSuchMethod(JNIEnv* env, const std::string& message) {
```
swankjesse: Nice. Java defines two exceptions that are almost identical here:

NoSuchMethodException: reflection failures
NoSuchMethodError: mismatch between compiler and code

I think you picked the right one. NoSuchMethodException is more common, but it’s a checked exception and you definitely need an unchecked exception.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -93,6 +93,45 @@ duk_ret_t javaObjectFinalizer(duk_context *ctx) {
   return 0;
 }
 
+void handleJavaScriptError(JNIEnv* env, duk_context* ctx) {
```
swankjesse: Can the name of this method be changed to suggest its side effect? Like `queueJavaExceptionForJavaScriptError` ? Hard to figure out a word that means `env->Throw` where the exception _will_ throw, but not til JNI returns.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -219,3 +228,52 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring name, jobject object, job
...
   duk_pop(m_context);
 }
+
+jobject DuktapeContext::call(JNIEnv *env, jstring name, jobject method, jobjectArray args) {
+  duk_push_global_object(m_context);
+  const JString instanceName(env, name);
+  if (!duk_get_prop_string(m_context, -1, instanceName)) {
+    duk_pop(m_context);
+    const jclass illegalArgumentException = env->FindClass("java/lang/IllegalArgumentException");
+    const std::string message = "A global object called " + instanceName.str() + " was not found";
```
swankjesse: I’ve got a pass-by-value / pass-by-reference question for you. Suppose I do the following strange sequence of calls:

1. JavaScript:
```
var x = {
  greet: function() { return "hello"; }
}
```

2. Java:
```
Greeter greeter = duktape.proxy("x", Greeter.class);
```

3. JavaScript
```
x = {
  greet: function() { return "howdy"; }
}
```

4. Java:
```
System.out.println(greeter.greet());
```

Does it print `hello` or `howdy` ? Effectively I’m asking whether we should bind `x` at the time of the call to `proxy()`, or if we should bind it upon each method call.

Doing it eagerly means we have a persistent reference across heaps that will need to be cleared, either manually, or by a side effect of the Java finalizer (which has concurrency problems). This is more difficult technically!

But doing it eagerly _also_ means that we can validate that `x` exists eagerly, and even if `x` is reassigned you get the value at the time of the `proxy()` call. There’s a tiny potential performance benefit because we will will look up `x` by something stronger than its name.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -219,3 +228,52 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring name, jobject object, job
...
+    env->ThrowNew(illegalArgumentException, message.c_str());
+    return nullptr;
+  }
+
+  // TODO: cache the call wrapper.
```
swankjesse: And I guess there’s a similar, weirder problem with the method references also:

1. JavaScript:
    ```
    var x = {
      greet: function() { return "hello"; }
    };
    ```
2. Java:
    ```
    Greeter greeter = duktape.proxy("x", Greeter.class);
    ```
3. JavaScript
    ```
    x.greet = function() { return "howdy"; };
    ```

swankjesse: Probably if we lock in the `x`, we should also lock in the `greet`, and vice-versa.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -219,3 +228,52 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring name, jobject object, job
...
+  jclass methodClass = env->GetObjectClass(method);
+
+  const jmethodID getName = env->GetMethodID(methodClass, "getName", "()Ljava/lang/String;");
+  const JString methodName(env, static_cast<jstring>(env->CallObjectMethod(method, getName)));
+
+  // Push the function (property) onto the stack.
+  if (!duk_get_prop_string(m_context, -1, methodName)) {
+    // Pop the global object and our proxied instance.
+    duk_pop_2(m_context);
+    throwNoSuchMethod(env, "Method " + instanceName.str() + "." + methodName.str() + " not found");
```
swankjesse: Maybe shove the word Duktape in here somewhere? Years from now Java programmers will look at their FooInterface and be like, "yes, the foo() method is totally found" not realizing that their interfaces are being implemented by JavaScript behind the scenes.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -219,3 +228,52 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring name, jobject object, job
...
+    return nullptr;
+  }
+
+  // Push 'this' onto the stack.
+  duk_push_null(m_context);
```
swankjesse: null?!

swankjesse: Ahh, all good. Push null, then clobber it.

duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -219,3 +228,52 @@ void DuktapeContext::bindInstance(JNIEnv *env, jstring name, jobject object, job
...
+    // TODO: handle other return types.
+    result = duk_is_null_or_undefined(m_context, -1)
+        ? nullptr
+        : env->NewStringUTF(duk_safe_to_string(m_context, -1));
+  } else {
+    handleJavaScriptError(env, m_context);
+    result = nullptr;
+  }
+
+  // Pop the global object, our instance, and the call result.
```
swankjesse:
```
 // Pop the global object, our instance, and the call’s result or error.
```

tests/src/androidTest/java/com/squareup/duktape/DuktapeBindTest.java
```diff
@@ -0,0 +1,383 @@
...
+    duktape.bind("value", TestInterface.class, new TestInterface() {
+      @Override public String getValue() {
+        return "foo";
+      }
+    });
+    try {
+      duktape.bind("value", TestInterface.class, new TestInterface() {
+        @Override
+        public String getValue() {
+          return "bar";
```
swankjesse: supernit: having code like `throw new AssertionError()` here is preferable. If we ever want to run code coverage on this project, this line should never be reached and lines like `throw new AssertionError()` can be safely ignored.

duktape/src/main/jni/DuktapeContext.cpp
swankjesse: Or should this just be DuktapeException, for consistency with `bind()` ? Hmm.

swankjesse: Having a rule like _“We always throw DuktapeException when your Java and JavaScript disagree”_ is easy to work with.

tests/src/androidTest/java/com/squareup/duktape/DuktapeBindTest.java
```diff
@@ -0,0 +1,383 @@
...
+    duktape.bind("value", TestInterface.class, new TestInterface() {
+      @Override
+      public String getValue() {
+        return "bar";
+      }
+    });
+    try {
+      duktape.evaluate("value.increment()");
+      fail();
+    } catch (DuktapeException expected) {
```
swankjesse: I think that with proxy() we’re doing a NoSuchMethodError, and this seems inconsistent? 

I think all we really want to do is make up some hard rules and follow ’em.

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+          .hasMessage("Only interfaces can be proxied. Received: class java.lang.String");
+    }
+  }
+
+  interface TestInterface {
+    String getValue();
+  }
+
+  @Test public void proxy() {
+    duktape.evaluate("var value = { getValue: function() { return '8675309'; } }");
```
swankjesse: nit: needs `;` at the end

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+    try {
+      proxy.getValue();
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("A global object called DoesNotExist was not found");
+    }
+  }
+
+  @Test public void proxyMissingMethodThrows() {
+    duktape.evaluate("var value = { getOtherValue: function() { return '8675309'; } }");
```
swankjesse: `:`

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  @Test public void proxyMethodNotCallableThrows() {
+    duktape.evaluate("var value = { getValue: '8675309' }");
+    TestInterface proxy = duktape.proxy("value", TestInterface.class);
+    try {
+      proxy.getValue();
+      fail();
+    } catch (DuktapeException expected) {
+      assertThat(expected).hasMessage("TypeError: '8675309' not callable");
```
swankjesse: nice

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+    TestInterface proxy = duktape.proxy("value", TestInterface.class);
+
+    // Close the context - proxy can no longer be used.
+    duktape.close();
+
+    try {
+      proxy.getValue();
+      fail();
+    } catch (NullPointerException expected) {
+      assertThat(expected).hasMessage("Null Duktape context");
```
swankjesse: Maybe add to the message: “Is your duktape closed?”

swankjesse: (Helps programmers figure out where their race condition is)

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  @Test public void proxyCallThrows() {
+    duktape.evaluate("var value = { getValue: function() { throw 'nope'; } }");
```
swankjesse: `;`

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+    TestInterface proxy = duktape.proxy("value", TestInterface.class);
+
+    try {
+      proxy.getValue();
+      fail();
+    } catch (DuktapeException expected) {
+      assertThat(expected).hasMessage("nope");
```
swankjesse: Nice!

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+    try {
+      proxy.getValue();
+      fail();
+    } catch (NoSuchMethodError expected) {
+      assertThat(expected).hasMessage("Method value.getValue not found");
+    }
+  }
+
+  @Test public void proxyMethodNotCallableThrows() {
+    duktape.evaluate("var value = { getValue: '8675309' }");
```
swankjesse: `:`

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -0,0 +1,120 @@
...
+    try {
+      proxy.getValue();
+      fail();
+    } catch (DuktapeException expected) {
+      assertThat(expected).hasMessage("TypeError: '8675309' not callable");
+    }
+  }
+
+  @Test public void proxyCalledAfterDuktapeClosed() {
+    duktape.evaluate("var value = { getValue: function() { return '8675309'; } }");
```
swankjesse: `;`

swankjesse:
# LGTM

This is wonderful. Lots of comments because I can’t help myself, but it’s all pedantic.

swankjesse:
I also wonder about better names for `proxy` and `bind`. Would it be absurd to have names that are closer to their utility rather than their implementation?

```
duktape.bind("a", TestInterface.class, new TestInterface() {
  ...
});
TestInterface b = duktape.proxy("b TestInterface.class);
```

–>

```
duktape.set("a", TestInterface.class, new TestInterface() {
  ...
});
TestInterface b = duktape.get("b TestInterface.class);
```

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: I think I agree.  My take is that in your example above, if you proxy something to x but then reassign x in JS, the Java proxy still points to the previous instance. (I am still working out how to look up the object/property with something stronger than the name).  

This also applies for another case where you change step 3 above to `x.greet = function() { return "howdy"; };`.  The Java proxy should still be connected to the previous function property that was replaced.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Agreed.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Yeah, I like that.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: Neat!  Will do.

duktape/src/main/java/com/squareup/duktape/Duktape.java
szurbrigg: TIL :-)

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Works for me.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: > Years from now

:+1: 

Will do.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Yeah, I think I really wanted `duk_dup(m_context, -2);` here instead of those two, but I had spaced on the function name at the time.  Will change.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Sure thing.

tests/src/androidTest/java/com/squareup/duktape/DuktapeBindTest.java
szurbrigg: Yeah, the difference is here the `DuktapeException` is wrapping something thrown by Duktape.

I guess that's another possibility: `DuktapeException` is something thrown by Duktape and we continue to use `NoSuchMethodError` when proxying, since it's sort of a Java-side thing?  I don't really feel strongly either way, though.

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
szurbrigg: Fixed x5.

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
szurbrigg: Good call.

tests/src/androidTest/java/com/squareup/duktape/DuktapeBindTest.java
szurbrigg: Sounds good.

tests/src/androidTest/java/com/squareup/duktape/DuktapeBindTest.java
swankjesse: Is it useful to differentiate between linking errors and runtime errors, when linking happens at runtime?

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: After more discussion, I think we've agreed that we'll track the instance of the object, but look up the method by name each time.  Reassigning `x` will print `hello`, but reassigning `x.greet` will print `howdy`.

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: Actually, I just noticed another Duktape shorthand for what I want: `duk_pcall_prop` which will throw for me if the property is not found or callable.

tests/src/androidTest/java/com/squareup/duktape/DuktapeBindTest.java
szurbrigg: Sounds like a trick question... :-)

PR #49, opened 2016-04-21
Create JavaScriptObject to proxy a JS global.

szurbrigg:
- Allows us to (hopefully) bind to the instance, rather than just the name (still a WIP).
- Allows us to wrap methods early & cache them like we do when binding a Java object.

### DIFF
--- a/duktape/src/main/jni/DuktapeContext.cpp
+++ b/duktape/src/main/jni/DuktapeContext.cpp
@@ -17,9 +17,12 @@
 #include <memory>
 #include <string>
 #include <stdexcept>
+#include <functional>
 #include "JString.h"
 #include "JavaMethod.h"
 #include "GlobalRef.h"
+#include "JavaScriptObject.h"
+#include "JavaExceptions.h"
 
 namespace {
 
@@ -93,44 +96,6 @@ duk_ret_t javaObjectFinalizer(duk_context *ctx) {
   return 0;
 }
 
-/**
- * Sets up a Java {@code DuktapeException} based on the Duktape JavaScript error at the top of the
- * Duktape stack. The exception will be thrown to the Java caller when the current JNI call returns.
- */
-void queueJavaExceptionForJavaScriptError(JNIEnv *env, duk_context *ctx) {
-  jclass exceptionClass = env->FindClass("com/squareup/duktape/DuktapeException");
-
-  // If it's a Duktape error object, try to pull out the full stacktrace.
-  if (duk_is_error(ctx, -1) && duk_get_prop_string(ctx, -1, "stack")) {
-    const char* stack = duk_safe_to_string(ctx, -1);
-
-    // Is there an exception thrown from a Java method?
-    if (duk_get_prop_string(ctx, -2, JavaMethod::JAVA_EXCEPTION_PROP_NAME)) {
-      jthrowable ex = static_cast<jthrowable>(duk_get_pointer(ctx, -1));
-
-      // add the Duktape JavaScript stack to this exception.
-      const jmethodID addDuktapeStack =
-          env->GetStaticMethodID(exceptionClass,
-                                 "addDuktapeStack",
-                                 "(Ljava/lang/Throwable;Ljava/lang/String;)V");
-      env->CallStaticVoidMethod(exceptionClass, addDuktapeStack, ex, env->NewStringUTF(stack));
-
-      // Rethrow the Java exception.
-      env->Throw(ex);
-
-      // Pop the Java throwable.
-      duk_pop(ctx);
-    } else {
-      env->ThrowNew(exceptionClass, stack);
-    }
-    // Pop the stack text.
-    duk_pop(ctx);
-  } else {
-    // Not an error or no stacktrace, just convert to a string.
-    env->ThrowNew(exceptionClass, duk_safe_to_string(ctx, -1));
-  }
-}
-
 } // anonymous namespace
 
 DuktapeContext::DuktapeContext(JavaVM* javaVM)
@@ -147,6 +112,8 @@ DuktapeContext::DuktapeContext(JavaVM* javaVM)
 }
 
 DuktapeContext::~DuktapeContext() {
+  std::for_each(m_proxiedObjects.begin(), m_proxiedObjects.end(),
+                std::default_delete<JavaScriptObject>());
   duk_destroy_heap(m_context);
 }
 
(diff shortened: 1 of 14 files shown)


duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -228,44 +192,26 @@ void DuktapeContext::bind(JNIEnv *env, jstring name, jobject object, jobjectArra
...
-  duk_push_global_object(m_context);
-  // TODO: attach to something stronger than the name in proxy(), and use that here.
-  const JString instanceName(env, static_cast<jstring>(target));
-  if (!duk_get_prop_string(m_context, -1, instanceName)) {
-    duk_pop(m_context);
-    const jclass illegalArgumentException = env->FindClass("java/lang/IllegalArgumentException");
-    const std::string message = "A global object called " + instanceName.str() + " was not found";
-    env->ThrowNew(illegalArgumentException, message.c_str());
+jobject DuktapeContext::call(JNIEnv *env, jlong target, jobject method, jobjectArray args) {
+  JavaScriptObject *object = target < m_proxiedObjects.size()
```
baoti: target >= 0

duktape/src/main/jni/duktape-jni.cpp
```diff
@@ -79,33 +79,33 @@ Java_com_squareup_duktape_Duktape_evaluate__JLjava_lang_String_2Ljava_lang_Strin
...
 }
 
-JNIEXPORT jobject JNICALL
+JNIEXPORT jlong JNICALL
 Java_com_squareup_duktape_Duktape_proxy(JNIEnv *env, jclass type, jlong context, jstring name,
                                         jobjectArray methods) {
   DuktapeContext* duktape = reinterpret_cast<DuktapeContext*>(context);
   if (throwIfNull(env, duktape)) {
-    return nullptr;
+    return 0L;
```
baoti: 0 is a valid proxy instance, because `target(is 0) < m_proxiedObjects.size()`

duktape/src/main/jni/DuktapeContext.cpp
szurbrigg: It's implicit since `m_proxiedObjects.size()` is unsigned and `target` will be changed to unsigned for the compare, but I'll add the check to make it more apparent.

duktape/src/main/jni/duktape-jni.cpp
szurbrigg: Sure, but we've already thrown a NullPointerException :-).   I'll change it to return -1L to make it clear.

duktape/src/main/jni/JavaScriptObject.cpp
```diff
@@ -0,0 +1,178 @@
...
+  m_instance = duk_get_heapptr(m_context, -1);
+  if (m_instance == nullptr) {
+#endif
+    duk_pop_2(m_context);
+    throw std::invalid_argument("JavaScript global called " + m_name + " is not an object");
+  }
+
+  if (duk_has_prop_string(m_context, -1, WRAPPER_THIS_PROP_NAME)) {
+    duk_pop_2(m_context);
+    throw std::invalid_argument("JavaScript object " + m_name + " is already proxied");
```
swankjesse: should this Just Work?

swankjesse: (I’m imagining cases for and against)

duktape/src/main/jni/JavaScriptObject.cpp
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  // Make sure that the object has all of the methods we want.
+  jmethodID getName = nullptr;
+  const jsize numArgs = env->GetArrayLength(methods);
+  for (jsize i = 0; i < numArgs; ++i) {
+    auto method = env->GetObjectArrayElement(methods, i);
+    if (getName == nullptr) {
```
swankjesse: is this an optimization to avoid looking up the class for java.lang.Method ?  A little bizarre

swankjesse: (but fine!)

duktape/src/main/jni/JavaScriptObject.cpp
```diff
@@ -0,0 +1,178 @@
...
+#ifdef NO_INSTANCE_VAR
+  duk_push_global_object(m_context);
+  if (!duk_get_prop_string(m_context, -1, m_name.c_str())) {
+    duk_pop(m_context);
+    queueNullPointerException(env, "JavaScript object called " + m_name + " was not found");
+    return nullptr;
+  }
+#else
+  if (m_instance == nullptr) {
+    queueNullPointerException(env, "JavaScript object " + m_name + " has been garbage collected");
```
swankjesse: This error is odd . . . doesn't this happen only if the user deliberately assigns the value to null? Or is this if the Duktape has been closed?

duktape/src/main/jni/JavaScriptObject.cpp
```diff
@@ -0,0 +1,178 @@
...
+  JavaScriptObject* obj = reinterpret_cast<JavaScriptObject*>(duk_require_pointer(ctx, -1));
+  duk_pop(ctx);
+
+  duk_del_prop_string(ctx, -1, WRAPPER_THIS_PROP_NAME);
+
+  // Null out the instance pointer - it's been garbage collected!
+  obj->m_instance = nullptr;
+
+  if (obj->m_nextFinalizer) {
+    // Continue with the next finalizer in the chain.
```
swankjesse: neat!

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
```diff
@@ -117,4 +116,36 @@
...
+    duktape.evaluate("var value = { getValue: function() { return '8675309'; } };");
+
+    TestInterface proxy = duktape.proxy("value", TestInterface.class);
+    duktape.evaluate("value.getValue = function() { return '7471111'; };");
+
+    String v = proxy.getValue();
+    assertThat(v).isEqualTo("7471111");
+  }
+
+  @Test public void proxyNonObjectThrows() {
```
swankjesse: test a garbage collected value?

swankjesse:
LGTM. Just a handful of questions

duktape/src/main/jni/JavaScriptObject.cpp
szurbrigg: Yeah, I was having trouble deciding if we should forbid it or re-use proxies.  I think it's better (and simpler) if we just create a new proxy every time the method is called.

duktape/src/main/jni/JavaScriptObject.cpp
szurbrigg: That's exactly what it is. :-)

duktape/src/main/jni/JavaScriptObject.cpp
szurbrigg: `m_instance` points at the object that we're proxying in the Duktape VM.  I've set up the `finalizer` method to have Duktape let us know when that object is being GC'd so we can null out this pointer.    This could happen if the JavaScript code deletes/replaces the proxied variable.

Use of the proxy after closing Duktape will be detected before calling `DuktapeContext::call` in duktape-jni.cpp.

tests/src/androidTest/java/com/squareup/duktape/DuktapeProxyTest.java
szurbrigg: Good idea.  Will do.

PR #51, opened 2016-04-28
Create a stack checker used in debug builds.

szurbrigg:
- throws/aborts when the stack isn't what we expect
- shows the C++ callstack and what was on the duktape stack in logcat

### DIFF
--- a/duktape/build.gradle
+++ b/duktape/build.gradle
@@ -18,8 +18,15 @@ model {
     buildTypes {
       release {
         ndk {
-          CFlags.addAll(['-Os', '-fomit-frame-pointer'])
-          cppFlags.addAll(['-Os', '-fomit-frame-pointer'])
+          CFlags.addAll(['-Os', '-fomit-frame-pointer', '-DNDEBUG'])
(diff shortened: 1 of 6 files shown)


duktape/build.gradle
```diff
@@ -18,8 +18,15 @@ model {
     buildTypes {
       release {
         ndk {
-          CFlags.addAll(['-Os', '-fomit-frame-pointer'])
-          cppFlags.addAll(['-Os', '-fomit-frame-pointer'])
+          CFlags.addAll(['-Os', '-fomit-frame-pointer', '-DNDEBUG'])
+          cppFlags.addAll(['-Os', '-fomit-frame-pointer', '-DNDEBUG'])
+        }
+      }
+      debug {
```
szurbrigg: My gradle knowledge is pretty limited.  Does anyone know how I can get the unit tests to use a debug build of libduktape.so so we can hit assertions in the C++ code?

JakeWharton: Not with the experimental plugin that we're currently using. It may not even be supported with it. I'll have to dig into it more.

swankjesse:
Dig.

PR #59, opened 2016-08-09
Update README with JS proxy example.

szurbrigg:


README.md
```diff
@@ -63,14 +63,56 @@ try {
...
+
+```java
+try {
+  // Connect our interface to a JavaScript object called Utf8.
+  Utf8 utf8 = duktape.proxy("Utf8", Utf8.class);
+
+  String greeting = utf8.fromHex("EC9588EB8595ED9598EC84B8EC9A9421");
+  Log.d("Greeting", greeting);
+} finally {
+  duktape.close();
```
JakeWharton: It's weird to see this without the `create()` call but it would be weird to see the `create()` call without calling `evaluate`. Maybe omit the `create()` / `try` / `finally` / `close()` dance in both of these examples since it's outlined above?

swankjesse:
LGTM. Agree with Jake’s comment.

README.md
szurbrigg: Agreed.  Will do.

PR #68, opened 2016-09-11
Allow arguments of type Object in function calls.

szurbrigg:
- Type checking for these is done at runtime.

### DIFF
--- a/tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
+++ b/tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
@@ -276,4 +276,34 @@ public class DuktapeGetTest {
     assertThat(printer.call(true, 42, 2.718281828459))
         .isEqualTo("boolean: true, int: 42, double: 2.718281828459");
   }
+
+  interface TestMultipleObjectArgs {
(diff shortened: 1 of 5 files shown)


tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -276,4 +276,34 @@
...
+
+  @Test public void passUnsupportedTypeAsObjectFails() {
+    duktape.evaluate("var printer = {\n" +
+        "  print: function(b, i, d) {\n" +
+        "    return 'boolean: ' + b + ', int: ' + i + ', double: ' + d;\n" +
+        "  }\n" +
+        "};");
+    TestMultipleObjectArgs printer = duktape.get("printer", TestMultipleObjectArgs.class);
+    try {
+      printer.print(true, 42, new Date());
```
swankjesse: Nice.

swankjesse:
I like it. Curious what the motivation is though. As a rule I don’t like dynamic stuff that isn’t strictly necessary!

szurbrigg:
I want to add support for variadic functions next, so we can expose access to `Console.log` from Java, or provide a Java logger object to JavaScript:

PR #70, opened 2016-09-14
Fix type checking when returning values from JS.

szurbrigg:
- duk_require_XXX only really works inside of a script execution,
  so when we call out to JS and pop the result off the stack from
  outside the execution, we need to check the type.
- Make all results Object, and let Java take care of it.  The only
  exception is that JS has no Integer, so we have to truncate
  Doubles as needed.

### DIFF
--- a/duktape/src/main/jni/java/JavaType.h
+++ b/duktape/src/main/jni/java/JavaType.h
@@ -27,8 +27,13 @@
 class JavaType {
 public:
   virtual ~JavaType() = default;
-  /** Pops a {@code jvalue} from the Duktape stack in {@code ctx}. */
-  virtual jvalue pop(duk_context* ctx, JNIEnv*) const = 0;
+
+  /**
+   * Pops a {@code jvalue} from the Duktape stack in {@code ctx}. if {@code inScript} is true,
+   * caller is inside a JavaScript execution so JavaScript exceptions can be used. If false, type
+   * errors will throw C++ exceptions.
+   */
+  virtual jvalue pop(duk_context* ctx, JNIEnv* env, bool inScript) const = 0;
   /**
    * Pushes {@code value} to the Duktape stack in {@code ctx}. Returns the number of entries pushed.
    */
@@ -42,6 +47,8 @@ public:
    * Return true if this is a primitive (int, boolean, etc.), false if not (String, Integer, etc.).
    */
   virtual bool isPrimitive() const { return false; }
+  /** Return true if this type is a java.lang.Integer. */
+  virtual bool isInteger() const { return false; }
 };
 
 /** Manages the {@code JavaType} instances for a particular {@code DuktapeContext}. */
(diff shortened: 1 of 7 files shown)


duktape/src/main/jni/javascript/JavaScriptObject.cpp
```diff
@@ -206,14 +207,48 @@ duk_ret_t JavaScriptObject::finalizer(duk_context* ctx) {
...
+
 JavaScriptObject::MethodBody buildMethodBody(JavaTypeMap& typeMap, JNIEnv* env, jobject method,
                                              const std::string& methodName) {
   const jclass methodClass = env->GetObjectClass(method);
 
   const jmethodID getReturnType =
       env->GetMethodID(methodClass, "getReturnType", "()Ljava/lang/Class;");
   jclass returnTypeClass = static_cast<jclass>(env->CallObjectMethod(method, getReturnType));
-  const JavaType* returnType = typeMap.getBoxed(env, static_cast<jclass>(returnTypeClass));
+  // JavaScript always gives us doubles, so we convert to Integer if needed.
```
szurbrigg: This seems like overkill. I'm considering forbidding int/Integer results from JS calls as a simpler solution.  Thoughts?

swankjesse: I like that a lot better than this. It’s truer to the form.

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -261,6 +261,19 @@
     assertThat(proxy.d(null)).isNull();
   }
 
+  interface IntegerGetter {
+    int get();
+  }
+
+  @Test public void proxyCallTruncatesDoubleToInteger() {
+    duktape.evaluate("var value = {\n" +
+        "  get: function() { return 2.718281828459; }\n" +
```
swankjesse: and it avoids this weirdness.

duktape/src/main/jni/java/JavaType.h
```diff
@@ -27,8 +27,13 @@
 class JavaType {
 public:
   virtual ~JavaType() = default;
-  /** Pops a {@code jvalue} from the Duktape stack in {@code ctx}. */
-  virtual jvalue pop(duk_context* ctx, JNIEnv*) const = 0;
+
+  /**
+   * Pops a {@code jvalue} from the Duktape stack in {@code ctx}. if {@code inScript} is true,
+   * caller is inside a JavaScript execution so JavaScript exceptions can be used. If false, type
```
swankjesse: nice docs

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -240,27 +237,40 @@
...
+    int get();
+  }
+
+  @Test public void proxyMethodCannotReturnInteger() {
+    duktape.evaluate("var value = {\n" +
+        "  get: function() { return 2; }\n" +
+        "};");
+
+    try {
+      duktape.get("value", IntegerGetter.class);
```
swankjesse: fail

Issue #63, opened 2016-09-08
Get and set values to context


pietrocaselani:
Are there any plans to implement methods to get and set values (variables and functions) on a duktape context?
I need to get a javascript function from Java code and invoke it with arguments?
Do you guys plan to do this soon? Otherwise I could open a pull request.

szurbrigg:
We weren't planning anything like that.   Can you accomplish what you need with `Duktape.evaluate()`?  It can be used to set/get JS variables.

pietrocaselani:
I need to do something like that.
```javascript
function hello(name) {
  return "Hello " + name;
}
```

And then in Java code

```java
helloFunction = duktape.get("hello");
helloFunction.call("World");
```

I can do this on iOS/JavaScriptCore with this [method](https://developer.apple.com/reference/javascriptcore/jsvalue/1451648-call).

szurbrigg:
I've added a test that demonstrates how to do this with `Duktape.set` in https://github.com/square/duktape-android/pull/65.

PR #71, opened 2016-09-16
Load and manipulate a script

pietrocaselani:
The proposed solution on issue #63 was not exactly what I was looking for. I wish I could load a script, insert and modify the JavaScript values from Java, also, call functions in a simple way. I really liked the JavaScriptCore API for iOS, so I tried to do something like trying not to break the API that you already have.


swankjesse:
Thanks for the submission, but I don’t think this is a good fit for our API. I believe you might be able to accomplish something quite similar on top of this API.

PR #72, opened 2016-09-18
Store context in a ThreadLocal to explicitly forbid sharing.

szurbrigg:
https://github.com/square/duktape-android/issues/69


swankjesse:
I don’t really like this. It’s not consistent with how thread-unsafe things typically work in Java.

In particular, a reasonable programmer should be able to defend a `Duktape` instance with `synchronized` blocks and things should just work from any thread. That’s how `ArrayList` works! We’re better off just being explicit that `Duktape` instances aren’t thread safe.

szurbrigg:
Sure thing.

PR #74, opened 2016-09-24
Change the methods of Duktape to be synchronized.

szurbrigg:
https://github.com/square/duktape-android/issues/69


szurbrigg:
How about this, then?

PR #75, opened 2016-09-26
Support VarArg method calls between Java and JS.

szurbrigg:


duktape/src/main/jni/java/JavaMethod.cpp
```diff
@@ -59,7 +72,10 @@ duk_ret_t JavaMethod::invoke(duk_context* ctx, JNIEnv* env, jobject javaThis) co
   std::vector<jvalue> args(m_argumentLoaders.size());
   // Load the arguments off the stack and convert to Java types.
   // Note we're going backwards since the last argument is at the top of the stack.
-  for (ssize_t i = m_argumentLoaders.size() - 1; i >= 0; --i) {
+  if (m_isVarArgs) {
+    args.back().l = m_argumentLoaders.back()->popArray(ctx, env, argCount - minArgs, true);
```
swankjesse: neat

duktape/src/main/jni/java/JavaType.cpp
```diff
@@ -16,7 +16,29 @@
 #include "JavaType.h"
 #include "JString.h"
 #include "JavaExceptions.h"
-#include "GlobalRef.h"
+
+jarray JavaType::popArray(duk_context* ctx, JNIEnv* env, uint32_t count, bool inScript) const {
+  jobjectArray array = env->NewObjectArray(count, static_cast<jclass>(m_classRef.get()), nullptr);
+  for (auto i = count; i > 0u; --i) {
```
swankjesse: oh you

duktape/src/main/jni/java/JavaType.cpp
```diff
@@ -16,7 +16,29 @@
...
+    env->SetObjectArrayElement(array, i - 1, pop(ctx, env, inScript).l);
+  }
+  return array;
+}
+
+duk_ret_t JavaType::pushArray(duk_context* ctx, JNIEnv* env, const jarray& values) const {
+  const auto size = env->GetArrayLength(values);
+  jvalue element;
```
swankjesse: off topic: why is this declared outside of the loop?

swankjesse: (is it because I don’t think in C++? I think it is)

duktape/src/main/jni/java/JavaType.cpp
```diff
@@ -135,11 +164,41 @@ struct Boolean : public Primitive {
...
+    if (size == 0) {
+      return 0;
+    }
+    jvalue element;
+    jboolean* elements = env->GetBooleanArrayElements(static_cast<jbooleanArray>(values), nullptr);
+    for (jsize i = 0; i < size; ++i) {
+      element.z = elements[i];
+      try {
+        push(ctx, env, element);
+      } catch (std::invalid_argument& e) {
```
swankjesse: why would this happen?

duktape/src/main/jni/java/JavaType.cpp
```diff
@@ -135,11 +164,41 @@ struct Boolean : public Primitive {
...
+        env->ReleaseBooleanArrayElements(static_cast<jbooleanArray>(values), elements, JNI_ABORT);
+        duk_pop_n(ctx, i);
+        throw e;
+      }
+    }
+    env->ReleaseBooleanArrayElements(static_cast<jbooleanArray>(values), elements, JNI_ABORT);
```
swankjesse: off topic: not having `finally` kinda sucks

duktape/src/main/jni/java/JavaType.cpp
```diff
@@ -178,11 +237,41 @@ struct Integer : public Primitive {
     return value;
   }
 
+  jarray popArray(duk_context* ctx, JNIEnv* env, uint32_t count, bool inScript) const {
+    jintArray array = env->NewIntArray(count);
+    for (auto i = count; i > 0u; --i) {
```
swankjesse: why iterate over [count..1] and insert at [count-1..0] ? Wouldn’t it be simpler to iterate [count-1..0] ?

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -330,4 +331,50 @@
...
     }
   }
+
+  interface TestVarArgs {
+    String call(String separator, Object... args);
+  }
+
+  @Test public void callVarArgMethod() {
+    duktape.evaluate(""
+        + "function joiner() {\n"
```
swankjesse: cute

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -330,4 +331,50 @@
...
+    try {
+      joiner.call(", ", "Test", new Date(), 1.0);
+      fail();
+    } catch (Exception expected) {
+      assertThat(expected).hasMessage("Unsupported Java type class java.util.Date");
+    }
+  }
+
+  interface Summer {
+    double sum(double... args);
```
swankjesse: wanna do the boilerplate of repeating with booleans and ints? You’re already duplicating code in the implementation so it’s prudent to cover it all

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -330,4 +331,50 @@
...
+        + "  sum: function() {\n"
+        + "    var result = 0;\n"
+        + "    for (i = 0; i < arguments.length; i++) {\n"
+        + "      result = result + arguments[i];\n"
+        + "    }\n"
+        + "    return result;"
+        + "  }\n"
+        + "};");
+    Summer summer = duktape.get("summer", Summer.class);
+    assertThat(summer.sum()).isWithin(0.000001).of(0.0);
```
swankjesse: it’s not exactly 0 ?

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -330,4 +331,50 @@
...
+    assertThat(summer.sum(1, 2, 3, 4)).isWithin(0.000001).of(10.0);
```
swankjesse: it’s not exactly 0?

duktape/src/main/jni/java/JavaType.cpp
szurbrigg: Good eye, it can't here.  `Object.push` can, and I roughly copied the method from there.  Will remove from the `Primitives` where it can't.

duktape/src/main/jni/java/JavaType.cpp
szurbrigg: Habit/premature optimization. Some mystics believe that if you declare it inside the loop, the compiler will keep allocating and deallocating it on each iteration, when we really just want a place to update our `.l` each time.  It's not a big struct, though, so it's probably not worth the hassle.

duktape/src/main/jni/java/JavaType.cpp
szurbrigg: I *should* write an RAII wrapper, but doing the release in two places seemed like less typing.

duktape/src/main/jni/java/JavaType.cpp
szurbrigg: `count` and therefore `i` are unsigned, so `i >= 0u` is always true.

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: Good call.  I was considering this, too.

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: com.google.common.truth.Truth.DoubleSubject:
> isEqualTo(Double other)
> Deprecated. 
> Use isWithin(double) instead. Double comparison should always have a tolerance.

Is there a better way?

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: It is, but `isEqualTo` is deprecated (see above).

duktape/src/main/jni/java/JavaType.cpp
szurbrigg: Actually, without that extra catch we're only calling free in one spot.

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: meh, I'll just use isEqualTo until I can't.

swankjesse: We’ll see if they bite.

PR #76, opened 2016-10-05
evaluate returns null if result type unsupported

szurbrigg:
- The last statement of a script ends up being the return value, which causes problems in some cases when we try to marshal it back to Java. e.g. evaluating "foo.bar = function() { return 3; }" will return a function object and cause an IllegalArgumentException.


szurbrigg:
An alternative to this might be follow the Duktape API, and create another method called Duktape.evaluateNoResult that returns void.

swankjesse:
LGTM. It's too bad we can't differentiate between an implicit return and an explicit one.

PR #78, opened 2016-10-10
Add custom date provider to support "2001/01/21"

zenanhu:


szurbrigg:
Thanks for the submission, but this is changing code in [Duktape](https://github.com/svaarala/duktape) - not this project.  If we were to support a custom date formatter/parser, I'd rather it be configurable.  For example, have the user specify a `DateFormat` for the instance returned from `Duktape.create()`.

PR #85, opened 2017-01-24
Basic (and limited) support for arrays in arguments and return values.

szurbrigg:
- Uses the JS stack to marshal, so currently limited to a few hundred
  elements.
- Will rework some marshaling code to support more, and
  multi-dimensional arrays.

### DIFF
--- a/duktape/src/main/jni/DuktapeContext.cpp
+++ b/duktape/src/main/jni/DuktapeContext.cpp
@@ -98,7 +98,13 @@ duk_ret_t javaObjectFinalizer(duk_context *ctx) {
 }
 
 void fatalErrorHandler(duk_context* ctx, duk_errcode_t code, const char* msg) {
+#ifndef NDEBUG
+  duk_push_context_dump(ctx);
+  const char* debugContext = duk_get_string(ctx, -1);
+  throw std::runtime_error(std::string(msg) + " (" + std::to_string(code) + ") - " + debugContext);
+#else
   throw std::runtime_error(msg);
+#endif
 }
 
 } // anonymous namespace
@@ -134,13 +140,36 @@ jobject DuktapeContext::evaluate(JNIEnv* env, jstring code, jstring fname) const
   }
 
   const int supportedTypeMask = DUK_TYPE_MASK_BOOLEAN | DUK_TYPE_MASK_NUMBER | DUK_TYPE_MASK_STRING;
-  if (!duk_check_type_mask(m_context, -1, supportedTypeMask)) {
+  if (duk_check_type_mask(m_context, -1, supportedTypeMask)) {
+    // The result is a supported scalar type - return it.
+    return m_objectType->pop(m_context, env, false).l;
+  } else if (!duk_is_array(m_context, -1)) {
     // The result is an unsupported type, undefined, or null.
     duk_pop(m_context);
     return nullptr;
   }
 
-  return m_objectType->pop(m_context, env, false).l;
+  const auto stackTop = duk_get_top_index(m_context);
+
+  // It's an array - load values onto the stack.
+  const auto arraySize = duk_get_length(m_context, -1);
+  for (duk_size_t i = 0; i < arraySize; ++i) {
+    duk_get_prop_index(m_context, -1 - i, i);
+  }
+
+  try {
+    // Create a Java array from the values.
+    jobject result = m_objectType->popArray(m_context, env, arraySize, false);
+
+    // Pop the JS array off the stack.
+    duk_pop(m_context);
+
+    return result;
+  } catch (std::invalid_argument& e) {
+    // Failed to marshal an array element - clean up the stack.
+    duk_set_top(m_context, stackTop);
+    throw e;
+  }
 }
 
 void DuktapeContext::set(JNIEnv *env, jstring name, jobject object, jobjectArray methods) {
--- a/tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
+++ b/tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
@@ -26,6 +26,7 @@ import org.junit.runner.RunWith;
 import java.util.Date;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.fail;
 
 @RunWith(AndroidJUnit4.class)
@@ -215,7 +216,7 @@ public class DuktapeGetTest {
 
     TestPrimitiveTypes proxy = duktape.get("value", TestPrimitiveTypes.class);
     assertThat(proxy.b(true)).isEqualTo(false);
-    assertThat(proxy.d(2, 6.28318)).isWithin(0.0001).of(3.14159);
+    assertThat(proxy.d(2, 6.28318)).isEqualTo(3.14159);
   }
 
   interface TestMultipleArgTypes {
@@ -248,7 +249,7 @@ public class DuktapeGetTest {
 
     TestBoxedPrimitiveArgTypes proxy = duktape.get("value", TestBoxedPrimitiveArgTypes.class);
     assertThat(proxy.b(false)).isEqualTo(true);
-    assertThat(proxy.d(6.28318)).isWithin(0.0001).of(3.14159);
+    assertThat(proxy.d(6.28318)).isEqualTo(3.14159);
 
     assertThat(proxy.b(null)).isNull();
     assertThat(proxy.d(null)).isNull();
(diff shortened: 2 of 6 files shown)


duktape/src/main/jni/DuktapeContext.cpp
```diff
@@ -98,7 +98,13 @@ duk_ret_t javaObjectFinalizer(duk_context *ctx) {
 }
 
 void fatalErrorHandler(duk_context* ctx, duk_errcode_t code, const char* msg) {
+#ifndef NDEBUG
+  duk_push_context_dump(ctx);
```
swankjesse: neat

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -388,4 +388,189 @@
...
+      + "    if (v) {\n"
+      + "      v.sort();\n"
+      + "    }\n"
+      + "    return v;\n"
+      + "  },\n"
+      + "  sortNullable: function(v) { return this.sort(v); }\n"
+      + "};";
+
+  @Ignore // TODO: rework array marshalling to not use the Duktape stack.
+  @Test public void longArrays() {
```
swankjesse: supernit: i first assumed this was `long` like `uint64`

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -388,4 +388,189 @@
...
+    duktape.evaluate(SORTER_FUNCTOR);
+
+    ObjectSorter sorter = duktape.get("Sorter", ObjectSorter.class);
+
+    int length = 10000; // Current safe limit is ~249.
```
swankjesse: curious: found by experimentation?

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -388,4 +388,189 @@
...
+  @Test public void arraysOfObjects() {
+    duktape.evaluate(SORTER_FUNCTOR);
+
+    ObjectSorter sorter = duktape.get("Sorter", ObjectSorter.class);
+
+    assertThat(sorter.sort(null)).isNull();
+
+    Object[] original = new Object[]{2, 4, 3, 1};
+    Object[] sorted = sorter.sort(original);
+    assertThat(sorted).isEqualTo(new Object[]{1.0, 2.0, 3.0, 4.0});
```
swankjesse: the double conversion is interesting. Makes sense because it’s `Object[]`

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -388,4 +388,189 @@
...
+    // Replace the sorter with something broken.
+    duktape.evaluate(""
+        + "Sorter.sort = function(v) {"
+        + "  return [ 'a', 'b', 3, 'd' ];\n"
+        + "};");
+    try {
+      sorter.sort(new String[]{ "b", "d", "c", "a" });
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("Cannot convert return value 3 to String");
```
swankjesse: nice! Love stuff like this

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -388,4 +388,189 @@
...
+        + "};");
+    try {
+      sorter.sort(new double[0]);
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("Cannot convert return value null to double");
+    }
+  }
+
+  // Note, we return double[] and Double[] since we can't return ints from JavaScript.
```
swankjesse: This is different from our behavior for scalars, right?

swankjesse: I’m curious how much pain it causes to support `int[]` ?
(no pressing need though)

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
```diff
@@ -494,4 +497,191 @@ public String getValue() {
...
+  }
+
+  @Test public void arraysOfObjects() {
+    duktape.set("Sorter", ObjectSorter.class, new ObjectSorter() {
+      @Override
+      public Object[] sort(Object[] args) {
+        if (args == null) return null;
+        Arrays.sort(args, new Comparator<Object>() {
+          @Override
+          public int compare(Object l, Object r) {
```
swankjesse: nit: `l` is a hostile variable name.

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
```diff
@@ -494,4 +497,191 @@ public String getValue() {
...
+            if (l == null) return -1;
+            if (r == null) return 1;
```
swankjesse: very much looks like `return l`

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
```diff
@@ -494,4 +497,191 @@ public String getValue() {
...
+        .isEqualTo(new String[]{null, "a", "b", "d"});
+
+    TimeZone original = TimeZone.getDefault();
+    try {
+      TimeZone.setDefault(TimeZone.getTimeZone("GMT+0:00"));
+      duktape.evaluate("Sorter.sort([ 1, 2, 3, new Date(0) ])");
+      fail();
+    } catch (DuktapeException expected) {
+      assertThat(expected)
+          .hasMessage("Error: Cannot marshal return value 1970-01-01 00:00:00.000+00:00 to Java");
```
swankjesse: fun.

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
```diff
@@ -494,4 +497,191 @@ public String getValue() {
...
+    });
+
+    assertThat((Object[]) duktape.evaluate("Sorter.sort(['b', 'd', 'c', 'a'])"))
+        .isEqualTo(new String[]{ "a", "b", "c", "d" });
+
+    try {
+      duktape.evaluate("Sorter.sort(['b', 'd', 3, 'a'])");
+      fail();
+    } catch (DuktapeException expected) {
+      assertThat(expected).hasMessage("TypeError: string required, found 3 (stack index -1)");
```
swankjesse: `stack index -1` is leaking implementation details. Will probably change if you build another mechanism for types

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
```diff
@@ -494,4 +497,191 @@ public String getValue() {
...
+      duktape.evaluate("Sorter.sort([2, 4, null, 1])");
+      fail();
+    } catch (DuktapeException expected) {
+      assertThat(expected).hasMessage("TypeError: number required, found null (stack index -1)");
+    }
+
+    assertThat((Object[]) duktape.evaluate("Sorter.sort([2, 4, 3.14, 1])"))
+        .isEqualTo(new Object[]{ 1.0, 2.0, 3.0, 4.0 });
+
+    assertThat((Object[]) duktape.evaluate("Sorter.sortNullsFirst([2, null, 3, 1])"))
```
swankjesse: this stuff is pretty cool

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
```diff
@@ -494,4 +497,191 @@ public String getValue() {
...
+        .isEqualTo(new Object[]{ null, 1.0, 2.0, 3.0 });
```
swankjesse: curious that your comparator puts null first, whereas on DuktapeGetTest:501 puts it last

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
```diff
@@ -494,4 +497,191 @@ public String getValue() {
...
+  }
+
+  interface BoolSorter {
+    boolean[] sort(boolean[] args);
+    Boolean[] sortNullsFirst(Boolean[] args);
+  }
+
+  @Test public void arraysOfBooleans() {
+    duktape.set("Sorter", BoolSorter.class, new BoolSorter() {
+      @Override public boolean[] sort(boolean[] args) {
```
swankjesse: O(N) sorting!

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: renamed to `marshalArrayWithManyElements`.

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: Yep!

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: We can't tell the difference between an int and double in JS, so we treat everything as a double (this is related to why we chose not to support int as a return value to Java).

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
szurbrigg: This is the same as the scalar case - we forbid int as a return type.

We totally could support int[] (and int), but we would end up silently truncating floating point numbers when the return type is declared int - same as a the scalar case.

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
szurbrigg: renamed to lhs and rhs.

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
szurbrigg: That message is from the Duktape VM, but yes, it will go away with the new system (since I won't be using the stack).

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
szurbrigg: I wrote these ones first, and had decided nulls first was reasonable.  I can change it to nulls last if you feel strongly.

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
swankjesse: SGTM!

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
swankjesse: Works for me. I’ve already forgetten all the details like this.

tests/src/androidTest/java/com/squareup/duktape/DuktapeSetTest.java
swankjesse: Nope, no strong preference.

misael20000:
> - Uses the JS stack to marshal, so currently limited to a few hundred
>   elements.
> - Will rework some marshaling code to support more, and
>   multi-dimensional arrays.

2b53832ff796b5cd2dc2d11c38741e3ee0f35094
Pathward A.N
124085244
3534760680001

PR #87, opened 2017-04-03
Support longer and multidimensional arrays as arguments.

szurbrigg:


duktape/src/main/jni/java/JavaType.h
```diff
@@ -41,18 +41,22 @@ class JavaType {
...
+   * false, the top entry on the JavaScript stack should be an array that will be converted.
    */
-  virtual jarray popArray(duk_context*, JNIEnv*, uint32_t count, bool inScript) const;
+  virtual jarray popArray(duk_context*, JNIEnv*, uint32_t count, bool expanded, bool inScript) const;
   /**
    * Pushes {@code value} to the Duktape stack in {@code ctx}. Returns the number of entries pushed.
    */
   virtual duk_ret_t push(duk_context* ctx, JNIEnv*, const jvalue& value) const = 0;
   /**
    * Pushes the elements of {@code values} to the Duktape stack in {@code ctx} individually.
```
swankjesse: tiny nit – this says “individually” but that’s now configurable

szurbrigg: Good catch.  Will fix in follow-up.

PR #97, opened 2017-09-05
Allocate a local JVM frame when binding interface methods.

szurbrigg:
https://github.com/square/duktape-android/issues/95


szurbrigg:
😄 Thanks Jake!

swankjesse:
Nice test!

PR #100, opened 2017-11-06
Change build config to reduce AAR size.

szurbrigg:


JakeWharton:
What's the savings?

szurbrigg:
A little disappointing - maybe half?  The arm64 version of the library went from over 7MB to 3.3.  Pretty far from the ~650kB we had with the old toolchain.  

It looks like the library still has debug information in it.  I wonder if gradle will strip it when it gets packaged into an APK (saw mention of that somewhere)?

PR #101, opened 2017-12-06
At support for 'float' 32-bit floating point types.

vii:
While increasingly rare, these are still frequently used in OpenGL APIs.

### DIFF
--- a/tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
+++ b/tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
@@ -54,11 +54,27 @@ public class DuktapeGetTest {
     String getValue();
   }
 
+  interface TestInterfaceMoreTypes {
+    String getValue();
+    double getValueDouble();
+    float getValueFloat();
+    boolean getValueBoolean();
+  }
+
   @Test public void get() {
-    duktape.evaluate("var value = { getValue: function() { return '8675309'; } };");
-    TestInterface proxy = duktape.get("value", TestInterface.class);
+    duktape.evaluate("var value = { "+
+                     "getValue: function() { return '8675309'; }," +
+                     "getValueDouble: function() { return 357; }," +
+                     "getValueFloat: function() { return 499.5; }," +
+                     "getValueBoolean: function() { return true; }," +
+                     "};");
+    TestInterfaceMoreTypes proxy = duktape.get("value", TestInterfaceMoreTypes.class);
     String v = proxy.getValue();
     assertThat(v).isEqualTo("8675309");
+
+    assertThat(proxy.getValueDouble()).isEqualTo(357.0);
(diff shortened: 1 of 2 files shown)


szurbrigg:
I'm curious what the motivation for this is?  The JS side won't be able to distinguish between floats and doubles, so why not just use double on the Java interfaces and handle the conversion in Java?

vii:
@szurbrigg there are some interfaces which already exist, which expect float[] arrays and parameters. For example, many OpenGL API calls use float[]

https://developer.android.com/reference/android/opengl/GLES20.html#glGetFloatv(int,%20float[],%20int)

It is obviously possible to handle floats by converting them in the Java code, but the same could be said for ints . . .

vii:
the cost of copying large float arrays back and forth to doubles for these OpenGL calls, is very high - and doubles use twice the memory!

szurbrigg:
Sure, although they're going to be copied from double arrays when we marshal them in/out of JS anyway.  And now I'm thinking we should remove ints from the API as well...  

Can you add some tests that pass/return `Float`, `float[]` and `Float[]`, then?  And some tests in `DuktapeSetTest`s to make sure that marshalling these types works when calling from JS to Java?

What should happen if a double coming from JS doesn't fit in a float?  Should it be silently truncated?  Should it throw?

vii:
@szurbrigg it would be consistent to forbid both ints and floats.

However, that imposes boilerplate conversions on the users. It is convenient to allow JavaScript to implement existing interfaces. Later on, if more types are added to Duktape (for example, V8 supports many more types, like Float32Array), then these types can be naturally translated.

The rules for floating point conversion from 64-bit to 32-bit are (I believe) that overly precise values are rounded to nearest. It's not pedantic and rounding will occur, but the default is sensible and I believe users of 32-bit floats (which are unusual as this pull request notes) are typically aware of such issues . . .

tests/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
```diff
@@ -541,6 +560,35 @@
     }
   }
 
+  interface FloatSorter {
+    float[] sort(float[] args);
+    Float[] sortNullable(Float[] args);
+  }
+
+  @Test public void arraysOfFloats() {
```
vii: here are the array tests @szurbrigg

vii:
@szurbrigg I added the tests you requested. This makes it slightly more convenient to use the Duktape bridge with existing Java interfaces. Let me know if I can make any more changes?

szurbrigg:
@vii Cool.  Can you please also add some overloads in DuktapeSetTest.java to sanity check float support?  `TestPrimitiveTypes` and `TestBoxedPrimitiveArgTypes` should have methods for float and Float respectively.   Also, please update the README to list float/Float as supported types.

PR #106, opened 2018-03-16
Fix 1.1.0 version in changelog

ChaitanyaPramod:
1.1.0 version was mistyped as 1.0.1. There is no maven artifact for 1.0.1


szurbrigg:
Thanks!

PR #107, opened 2018-03-25
Remove getLocalTimeZoneOffset from Duktape.java.

szurbrigg:


duktape/src/main/jni/duktape-jni.cpp
```diff
@@ -23,27 +24,31 @@
...
-  getLocalTimeZoneOffset = env->GetStaticMethodID(static_cast<jclass>(duktapeClass->get()),
-                                                  "getLocalTimeZoneOffset",
-                                                  "(D)I");
+
+  auto tzClass = env->FindClass("java/util/TimeZone");
+  auto getDefaultTimeZone = env->GetStaticMethodID(tzClass, "getDefault", "()Ljava/util/TimeZone;");
+  auto getOffset = env->GetMethodID(tzClass, "getOffset", "(J)I");
+
+  const GlobalRef timeZoneClass(env, tzClass);
+  getTimeZoneOffset = [timeZoneClass, getDefaultTimeZone, getOffset](double time) {
```
swankjesse: Explicitly declaring what's captured is interesting!

szurbrigg:
Hmm, you think it'll make a big performance difference?  I just didn't like having that weird method hanging out in Java when I could completely hide it in the lib.

duktape/src/main/jni/duktape-jni.cpp
szurbrigg: Safety first and no surprises.

swankjesse:
Gotcha. No I definitely don't care about the performance cost here either way. It's completely negligible.

But I was trying to infer your motivation and the branch name _one-less-hop_ made me guess it was intended as an optimization. 

Keeping the Java code short & direct is great, and this helps.

LGTM!

PR #114, opened 2018-07-16
Update to Duktape 2.2.1.

szurbrigg:
https://github.com/square/duktape-android/issues/108


duktape/src/main/jni/duktape/duk_config.h
```diff
@@ -3562,21 +3022,32 @@ typedef FILE duk_file;
...
 #else
 #error no provider for DUK_USE_DATE_GET_NOW()
 #endif
 
+/** Duktape-Android modification: Hack in date parsing and timezone support for Android */
+extern duk_int_t android__get_local_tzoffset(duk_double_t time);
+extern duk_bool_t android__date_parse_string(duk_context* ctx, const char* str);
+#define DUK_USE_DATE_GET_LOCAL_TZOFFSET android__get_local_tzoffset
+#define DUK_USE_DATE_PARSE_STRING android__date_parse_string
+/** End Duktape-Android. */
```
JakeWharton: This is unfortunate. Can we file an issue or something to allow easier external customization? Or is there info as to why that isn't allowed easily?

szurbrigg: Agreed.  I spent a bit of time trying to avoid this, but didn't want to hold the rest of the PR any longer.   I got to this point by looking at the example [here](https://github.com/svaarala/duktape/blob/master/examples/dummy-date-provider/dummy_date_provider.c).   I'll see if I can't find a better way (or request one) in a follow up.

PR #135, opened 2019-07-23
Migrate to AndroidX, shared dependencies

JakeWharton:


duktape/src/main/jni/StackChecker.h
```diff
@@ -54,7 +54,6 @@ class StackChecker {
   const duk_idx_t m_top;
 };
 
-// TODO: always run unit tests with a debug version of this library.
```
JakeWharton: AGP 3.x did this automatically when they lived in the sample, but now they live in this module where they run on the debug variant.

PR #138, opened 2019-07-24
Add error-prone and lint

JakeWharton:
Correct and enforce public API is entirely annotated for nullability. Correct and enforce some error-prone warnings.

### DIFF
--- a/duktape/src/main/java/com/squareup/duktape/Duktape.java
+++ b/duktape/src/main/java/com/squareup/duktape/Duktape.java
@@ -15,6 +15,8 @@
  */
 package com.squareup.duktape;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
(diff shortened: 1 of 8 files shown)


duktape/src/main/java/com/squareup/duktape/Duktape.java
```diff
@@ -154,16 +160,16 @@ public String toString() {
...
 
-  @Override protected synchronized void finalize() throws Throwable {
+  @Override protected synchronized void finalize() {
     if (context != 0) {
       Logger.getLogger(getClass().getName()).warning("Duktape instance leaked!");
     }
   }
 
   private static native long createContext();
-  private static native void destroyContext(long context);
```
JakeWharton: https://errorprone.info/bugpattern/UnsafeFinalization

JakeWharton:
Friendly ping!

PR #141, opened 2019-08-05
Splice QuickJS stack traces into QuickJsExceptions.

szurbrigg:


quickjs/src/main/jni/quickjs-jni.c
```diff
@@ -3,6 +3,7 @@
 #include <malloc.h>
 #include "quickjs/quickjs.h"
 
+// TODO(szurbrigg): delete these global refs on teardown.
```
JakeWharton: Should we store these in a struct next to the `JSRuntime` and return the struct instance pointer to the Java?

I originally had that setup for storing the `JSContext` and `JSRuntime` until I figured out that you can get one from the other.

quickjs/src/main/jni/quickjs-jni.c
```diff
@@ -56,22 +54,34 @@ Java_com_squareup_quickjs_QuickJs_evaluate__JLjava_lang_String_2Ljava_lang_Strin
...
       JS_FreeValue(context, messageValue);
 
+      const char *stack = JS_ToCString(context, stackValue);
+      JS_FreeValue(context, stackValue);
+
       jclass Exception = (*env)->FindClass(env, "com/squareup/quickjs/QuickJsException");
-      (*env)->ThrowNew(env, Exception, message);
+      char *messageAndStack = (char*) malloc(strlen(message) + strlen(stack) + 2);
+      strcpy(messageAndStack, message);
+      strcat(messageAndStack, "\n");
```
JakeWharton: Thinking that we should just create a two-argument constructor instead of doing a concatenation only to tear it back apart. Can do in a follow-up.

quickjs/src/main/jni/quickjs-jni.c
szurbrigg: Definitely. We'll have other pointers to track as well when we start binding Java and JS objects.

quickjs/src/main/jni/quickjs-jni.c
szurbrigg: Good call, this does feel a little clunky.

quickjs/src/main/java/com/squareup/quickjs/QuickJsException.java
```diff
@@ -18,10 +18,75 @@
...
+  public QuickJsException(@NonNull String detailMessage, @NonNull String jsStackTrace) {
+    super(detailMessage);
+    addJavaScriptStack(this, jsStackTrace);
+  }
+
+  /**
+   * Parses {@code StackTraceElement}s from {@code detailMessage} and adds them to the proper place
+   * in {@code throwable}'s stack trace.  Note: this method is also called from native code.
+   */
+  // TODO(szurbrigg): share this code with the DuktapeException implementation.
```
swankjesse: I don’t think there’s _that_ much utility to sharing the code here?

swankjesse: (in that it requires a new abstraction to share the code, and such an abstraction might be fragile)

szurbrigg: Yeah, probably not.  It started as a straight copy-paste, but it's already diverged a little.  Will delete the TODO in my next PR.

PR #150, opened 2019-09-03
Convert quickjs-jni to C++.

szurbrigg:


quickjs/src/main/jni/quickjs-jni.cpp
```diff
@@ -0,0 +1,329 @@
...
+};
+
+
+extern "C" JNIEXPORT jlong JNICALL
+Java_com_squareup_quickjs_QuickJs_createContext(JNIEnv *env, jclass type) {
+  Context *c = new(std::nothrow) Context(env);
+  if (!c) {
+    return 0;
+  }
+  booleanValueOf = env->GetStaticMethodID(c->booleanClass, "valueOf",
```
JakeWharton: I don't understand why this is stored in a static field whereas the class reference is stored in an instance field of the context. Shouldn't they be kept either both static or both instance?

quickjs/src/main/jni/quickjs-jni.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+static void throwJsException(JNIEnv *env, Context *context, JSValue value) {
+  JSContext *jsContext = context->jsContext;
+
+  JSValue exceptionValue = JS_GetException(jsContext);
+
+  JSValue messageValue = JS_GetPropertyStr(jsContext, exceptionValue, "message");
+  JSValue stackValue = JS_GetPropertyStr(jsContext, exceptionValue, "stack");
+
+// If the JS does a `throw 2;`, there won't be a message property.
```
JakeWharton:
```suggestion
  // If the JS does a `throw 2;`, there won't be a message property.
```

quickjs/src/main/jni/quickjs-jni.cpp
szurbrigg: The method ID is managed by the VM - there's no method to explicitly release it like there is for the global class references, which is why I left them out of the context structure.  After your comment and looking at it some more though, I agree it feels weird, so I'll move them into the Context class.

Issue #120, opened 2018-12-06
JNI ERROR (app bug): local reference table overflow (max=512)


resengupta:
JNI ERROR (app bug): local reference table overflow (max=512)
    art/runtime/indirect_reference_table.cc:132] local reference table dump:
    art/runtime/indirect_reference_table.cc:132]   Last 10 entries (of 512):
    art/runtime/indirect_reference_table.cc:132]       511: 0x134f6250 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       510: 0x134f6220 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       509: 0x134f61f0 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       508: 0x134f61c0 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       507: 0x134f6190 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       506: 0x134f6160 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       505: 0x134f6130 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       504: 0x134f6100 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       503: 0x134f60d0 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]       502: 0x134f60a0 libcore.util.ZoneInfo
    art/runtime/indirect_reference_table.cc:132]   Summary:
    art/runtime/indirect_reference_table.cc:132]         2 of $Proxy21 (1 unique instances)
    art/runtime/indirect_reference_table.cc:132]         2 of $Proxy20 (1 unique instances)
    art/runtime/indirect_reference_table.cc:132]         1 of java.lang.Thread
    art/runtime/indirect_reference_table.cc:132]         2 of java.lang.reflect.Method (2 unique instances)
    art/runtime/indirect_reference_table.cc:132]         4 of java.lang.String (2 unique instances)
    art/runtime/indirect_reference_table.cc:132]         1 of java.util.ArrayList
    art/runtime/indirect_reference_table.cc:132]         1 of java.lang.Object[] (1 elements)
    art/runtime/indirect_reference_table.cc:132]         1 of java.lang.Object[] (2 elements)
[...]

PR #152, opened 2019-09-11
Fix JNI local ref issues

bwalter:
While fixing all the possible JNI issues would require to go through all JNI local refs used in the project, we can at least fix the most frequent issues:
- when using the custom Duktape Data parser (e.g.:  `for (var i = 0; i < 1000; ++i) { new Date(i).toString(); }`)
- when pushing large object arrays


bwalter:
Fixes #120

PR #153, opened 2019-09-14
Support passing arguments to JsMethodProxies.

szurbrigg:


bwalter:
If you are interested, I will try to help with the reviewing from time to time as we have been doing very similar changes in our duktape-android-based project.

quickjs/src/main/jni/JsMethodProxy.cpp
```diff
@@ -18,53 +18,119 @@
 #include <algorithm>
 #include "Context.h"
 
-JsMethodProxy::JsMethodProxy(JNIEnv *env, const char *name, jobject method)
-    : name(name), methodId(env->FromReflectedMethod(method)) {
+namespace {
+
+std::string getName(JNIEnv *env, jclass javaClass) {
+  const jmethodID method =
+      env->GetMethodID(env->GetObjectClass(javaClass), "getName", "()Ljava/lang/String;");
```
bwalter: As this function is likely to get called quite often in the future, it would be a also a good idea to release the local ref returned by GetObjectClass(), too.

quickjs/src/main/jni/JsMethodProxy.cpp
```diff
@@ -18,53 +18,119 @@
...
+  } else if (typeName == "int") {
+    return [](const Context* c, jvalue v) {
+      return JS_NewInt32(c->jsContext, v.i);
+    };
+  } else if (typeName == "boolean") {
+    return [](const Context* c, jvalue v) {
+      return JS_NewBool(c->jsContext, v.z);
+    };
+  } else {
+    return [](const Context*, jvalue) { return JS_NULL; };
```
bwalter: Is it going to be handled via an exception in the future?

quickjs/src/main/jni/JsMethodProxy.cpp
```diff
@@ -18,53 +18,119 @@
...
+    arg.l = context->env->GetObjectArrayElement(args, numArgs);
     arguments[numArgs] = argumentLoaders[numArgs](context, arg);
+    context->env->DeleteLocalRef(arg.l);
   }
 
   jobject result;
   if (!context->env->ExceptionCheck()) {
-    JSValue callResult = JS_Call(context->jsContext, function, thisPointer, numArgs, arguments);
+    auto property = JS_NewAtom(context->jsContext, name.c_str());
+    JSValue callResult = JS_Invoke(context->jsContext, thisPointer, property, numArgs, arguments);
```
bwalter: I have just noticed that we are not doing that either but the JSValue of the argumentLoaders array might have to be free'd after JS_Invoke(). Also (even it probably doesn't matter much) maybe the ArgumentLoader typedef should return JSValue (instead of JSValueConst).

quickjs/src/main/jni/JsMethodProxy.cpp
szurbrigg: Good idea

quickjs/src/main/jni/JsMethodProxy.cpp
szurbrigg: yeah, I'll add a TODO

quickjs/src/main/jni/JsMethodProxy.cpp
szurbrigg: It is... line 140?

bwalter: No, what I mean is that the JSValue instances created by the argument loaders need to be eventually free'd (not relevant for primitives but the reference count needs to be decremented for strings, arrays, objects, ...).

In other words, you probably need to iterate through the "arguments" array after JS_Invoke() and "JS_FreeValue()" them otherwise the JS objects will be leaked.

Also the argument loader lambda should probably return a JSValue (instead of JSValueConst) because it is newly created and it feels weird to free a const value afterwards.

szurbrigg:
```  for (int i = 0; i < numArgs; i++) {
    JS_FreeValue(context->jsContext, arguments[i]);	    JS_FreeValue(context->jsContext, arguments[i]);
  }	  }```

Comment (szurbrigg) [reply quickjs/src/main/jni/JsMethodProxy.cpp]: I have the arguments as JSValueConst since that's the type the invoke API accepts:  
```
JSValue JS_Invoke(JSContext *ctx, JSValueConst this_val, JSAtom atom,
                  int argc, JSValueConst *argv);
```

PR #157, opened 2019-09-28
 Support array args and variadic java methods.

szurbrigg:


quickjs/src/main/java/com/squareup/quickjs/QuickJsException.java
```diff
@@ -64,8 +64,7 @@ private static void addJavaScriptStack(Throwable throwable, String detailMessage
     for (StackTraceElement stackTraceElement : throwable.getStackTrace()) {
       if (!spliced
           && stackTraceElement.isNativeMethod()
-          && stackTraceElement.getClassName().equals(QuickJs.class.getName())
-          && stackTraceElement.getMethodName().equals("evaluate")) {
```
JakeWharton: What makes this conditional fail?

quickjs/src/main/jni/JsMethodProxy.cpp
```diff
@@ -37,6 +37,62 @@ std::string getName(JNIEnv *env, jclass javaClass) {
...
+        if (!v.l) return JS_NULL;
+        JSValue result = JS_NewArray(c->jsContext);
+        auto elements = c->env->GetIntArrayElements(static_cast<jintArray >(v.l), nullptr);
+        for (jsize i = 0, e = c->env->GetArrayLength(static_cast<jarray>(v.l)); i < e; i++) {
+          JS_SetPropertyUint32(c->jsContext, result, i, JS_NewInt32(c->jsContext, elements[i]));
+        }
+        c->env->ReleaseIntArrayElements(static_cast<jintArray>(v.l), elements, JNI_ABORT);
+        return result;
+      };
+    } else if (elementTypeName == "boolean") {
```
JakeWharton: Feels like there's missing test cases for these new types

szurbrigg: QuickJsGetTest.proxyFunctionObject()

quickjs/src/main/java/com/squareup/quickjs/QuickJsException.java
szurbrigg: Whoops, this shouldn't have made it in - was just playing. Thanks.

PR #160, opened 2019-10-06
Include JS trace when code called through proxy

### LINKED ISSUE #159: JS stack trace not added to DuktapeException when a proxy is used
Java stack trace gets populated properly if the exception is thrown as the result of a call to `Duktape.evaluate()`:
```kotlin
duktape.evaluate("""
    1 + 1
    1 + 2
    1 + z
    """.trimIndent(), "test.js")
```
produces:
```java
com.squareup.duktape.DuktapeException: ReferenceError: identifier 'z' undefined
    at JavaScript.eval(test.js:3)
    at com.squareup.duktape.Duktape.evaluate(Native Method)
    at com.squareup.duktape.Duktape.evaluate(Duktape.java:60)
    at com.squareup.duktapestacktrace.DuktapeCaller.call(DuktapeCaller.kt:25)
```
However, if the code is executed through a proxy, the JS part of the trace is missing.
```kotlin
duktape.evaluate("""
    function ad(a, b) {
      return a + b;
    }
    """.trimIndent())
duktape.evaluate("""
    var Adder = {
      add: function(a, b) {
        return add(a, b);
      }
    };
    """.trimIndent(), "test.js")
val adder = duktape.get("Adder", Adder::class.java)
return adder.add(1, 2)
```
produces:
```java
com.squareup.duktape.DuktapeException: ReferenceError: identifier 'add' undefined
    at com.squareup.duktape.Duktape.call(Native Method)
    at com.squareup.duktape.Duktape.access$100(Duktape.java:28)
    at com.squareup.duktape.Duktape$1.invoke(Duktape.java:139)
    at java.lang.reflect.Proxy.invoke(Proxy.java:1006)
    at $Proxy0.add(Unknown Source)
    at com.squareup.duktapestacktrace.DuktapeCaller.call(DuktapeCaller.kt:24)
```
The reason is likely that the code in `DuktapeException` is looking for the `evaluate` call in the trace, which is missing from the second example.

Egorand:
Fixes #159

### DIFF
--- a/duktape/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
+++ b/duktape/src/androidTest/java/com/squareup/duktape/DuktapeGetTest.java
@@ -122,6 +122,20 @@ public final class DuktapeGetTest {
     }
   }
 
+  @Test public void proxyCallThrowsIncludeStacktrace() {
+    duktape.evaluate("function nop() { return 1; }");
+    duktape.evaluate("var value = { getValue: function() { return nope(); } };", "test.js");
+    TestInterface proxy = duktape.get("value", TestInterface.class);
+
+    try {
+      proxy.getValue();
+      fail();
+    } catch (DuktapeException expected) {
+      assertThat(expected.getStackTrace()).asList()
+          .contains(new StackTraceElement("JavaScript", "[anon]", "test.js", 1));
+    }
+  }
+
   @Test public void replaceProxiedObjectProxyReferencesOld() {
     duktape.evaluate("var value = { getValue: function() { return '8675309'; } };");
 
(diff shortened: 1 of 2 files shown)


Egorand:
Failing test is related to the Regex pattern change, which now lets a line from a `*.cpp` file through - I guess this is undesirable. I'll look for another option.

Egorand:
I changed the regex pattern to disallow lines coming from `.cpp` files. Is this solution robust enough?

szurbrigg:
LGTM! Do we need the same thing in the QuickJS version?

Egorand:
I can take a look later and maybe submit a followup PR.

Egorand:
Can someone merge please? Don't have rights. Also, any plans for a release?

JakeWharton:
We would need to disable QuickJS so it wasn't included in a release because it's not ready.

PR #161, opened 2019-10-11
Update quickjs-jni.cpp

webfolderio:
Added missing macro


JakeWharton:
Hmm why wasn't this failing in the tests? How can we prevent this in the future?

webfolderio:
It might be compiler issue. I use msvc instad of gcc.

JakeWharton:
Thankfully these methods shouldn't change much so I guess it's not worth worrying about figuring out some kind of validation of this at this time.

PR #164, opened 2019-10-21
QuickJs.set object support.

szurbrigg:


quickjs/src/androidTest/java/com/squareup/quickjs/QuickJsSetTest.java
```diff
@@ -0,0 +1,1149 @@
+/*
```
szurbrigg: Copy of DuktapeSetTest, with a couple tests @Ignored for exception propagation issues (will do in follow-up)

quickjs/src/main/jni/JavaObjectProxy.cpp
```diff
@@ -0,0 +1,59 @@
...
+    env->DeleteLocalRef(method);
+  }
+
+  if (!env->ExceptionCheck()) {
+    JS_SetPropertyFunctionList(context->jsContext, proxy, functions, numMethods);
+  }
+}
+
+JavaObjectProxy::~JavaObjectProxy() {
+  for (int i = 0; i < proxies.size(); i++) {
```
JakeWharton: This seems like it should use the functions' size?

szurbrigg: I updated so that JavaMethodProxy manages the memory for the method name, so now I don't have to loop and free them here.  Also changed functions to a vector<> so it automatically gets cleaned up as well.

PR #168, opened 2019-11-21
Don't store the JNIEnv in the QuickJs Context.

szurbrigg:


quickjs/src/main/jni/Context.cpp
```diff
@@ -754,6 +766,26 @@ JSValue Context::throwJavaExceptionFromJs() const {
...
+
+  javaVm->AttachCurrentThread(
+#ifdef __ANDROID__
+      &env,
+#else
+      reinterpret_cast<void**>(&env),
+#endif
+      nullptr);
+  if (env) {
+    thread_local jniEnvDetacher detacher(javaVm);
```
JakeWharton: Is this cleared if the thread goes away? This is the only part I don't think I understand.

szurbrigg: Yeah, `thread_local` basically keeps the object alive until the thread exit, and then the destructor for `jniEnvDetacher` will get called and disconnect it from the JVM.

PR #169, opened 2019-11-21
Don't leak global class references in QuickJs.

szurbrigg:


szurbrigg:
I had to take the `const` qualifier off all the methods/Context pointers since we might need to modify the globalReferences map.

PR #171, opened 2019-12-10
QuickJS function proxy fails when there are no arguments

alastaircoote:
If I make the following interface:

```
interface What {
    fun value(): Int
}
```
then try create a proxy to a JS object that matches the interface:

```
val context = QuickJs.create()
context.evaluate("var test_value = {value: function() {return 10 }};")
val gotInstance = context.get("test_value",What::class.java)
val gotValue = gotInstance.value()
```

That last line fails, because the args array is null. However the interface instead looks like:

```
interface What {
    fun value(i:Int): Int
}
```

and I run `gotInstance.value(10)` it works fine, because an array has been passed.

I've attached a pull request that fixes the issue by creating an issue if there isn't one. There's almost definitely a better way to do this, but the C++ code is a little out of my comfort zone so I've left it like this for now.

### DIFF
--- a/quickjs/src/main/java/com/squareup/quickjs/QuickJs.java
+++ b/quickjs/src/main/java/com/squareup/quickjs/QuickJs.java
@@ -133,12 +133,18 @@ public final class QuickJs implements Closeable {
         new InvocationHandler() {
           @Override
           public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+
+            Object[] argsToSend = args;
+            if (argsToSend == null) {
+              argsToSend = new Object[0];
+            }
+
             // If the method is a method from Object then defer to normal invocation.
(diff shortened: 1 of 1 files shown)


JakeWharton:
What's the failure of the existing code? I suspect we can fix this on the native side so that we're not forced to allocate.

alastaircoote:
D'oh, not sure how I forgot to include that. I get:

```
A/zygote: java_vm_ext.cc:504] JNI DETECTED ERROR IN APPLICATION: jarray was NULL
    java_vm_ext.cc:504]     in call to GetArrayLength
    java_vm_ext.cc:504]     from java.lang.Object com.squareup.quickjs.QuickJs.call(long, long, java.lang.Object, java.lang.Object[])
```

JakeWharton:
Cool. I think we can fix that on the native side to avoid the allocation.

PR #183, opened 2021-06-10
New release process and snapshot publishing

JakeWharton:
Based on the same thing we use everywhere else.


gradle.properties
```diff
@@ -1,3 +1,21 @@
+GROUP=app.cash.duktape
```
mattprecious, line 1: Intentional change?
```suggestion
GROUP=com.squareup.duktape
```

mattprecious (approved):

gradle.properties
JakeWharton: Hmm yeah let's leave duktape in the old and quickjs in the new

gradle.properties
```diff
@@ -1,3 +1,21 @@
...
+POM_URL=https://github.com/cashapp/duktape-android/
+POM_SCM_URL=https://github.com/cashapp/duktape-android/
+POM_SCM_CONNECTION=scm:git:git://github.com/cashapp/duktape-android.git
+POM_SCM_DEV_CONNECTION=scm:git:ssh://git@github.com/cashapp/duktape-android.git
+
+POM_LICENCE_NAME=Apache-2.0
+POM_LICENCE_URL=https://www.apache.org/licenses/LICENSE-2.0
+POM_LICENCE_DIST=repo
+
+POM_DEVELOPER_ID=cashapp
```
mattprecious, line 15: All this has changed, too. Intentional?

JakeWharton: Yeah I mean the repo moved and this is what we use on other cashapp projects. The values are mostly irrelevant but are required.

PR #186, opened 2021-06-10
Update QuickJS to 2021-03-27

JakeWharton:


quickjs/src/main/jni/Context.cpp
```diff
@@ -471,7 +471,7 @@ Context::getJsToJavaConverter(JNIEnv* env, jclass type, bool boxed) {
             for (int i = 0; i < length && !env->ExceptionCheck(); i++) {
               int element;
               auto jsElement = JS_GetPropertyUint32(c->jsContext, v, i);
-              if (!JS_IsInteger(jsElement)) {
+              if (JS_VALUE_GET_TAG(jsElement) != JS_TAG_INT) {
```
JakeWharton, line 474: This function was removed. It previously checked for values which were either `JS_TAG_INT` or `JS_TAG_BIG_INT`. I've changed it to check only for `JS_TAG_INT` because calling `length` on an array calls `JS_NewInt32` which creates a value of this specific type.

quickjs/src/main/jni/Context.cpp
```diff
@@ -658,7 +658,7 @@ Context::getJsToJavaConverter(JNIEnv* env, jclass type, bool boxed) {
   } else if (typeName == "int") {
     return [](Context* c, JNIEnv* env, const JSValueConst& v) {
       jvalue result;
-      if (!JS_IsInteger(v)) {
+      if (JS_VALUE_GET_TAG(v) != JS_TAG_INT) {
```
JakeWharton, line 661: Similar justification here except it's based on your own expectation of the type

Egorand (approved):
Everything under `src/main/jni/quickjs` are just bundled sources?

JakeWharton:
Yep!

PR #218, opened 2021-08-04
Fix UTF-8 encoding of inputs and outputs

### LINKED ISSUE #214: Emoji roundtrip fails

swankjesse:
Use the Java APIs rather than the JNI APIs to convert char* to UTF-8
and back. The Java APIs use UTF-8, not modified UTF-8.

Also include a trailing '\u0000', which QuickJS expects (even though
it accepts the string length as a parameter).

Also fix a bug where the test case had an unexpected unprintable
character.

Closes: https://github.com/cashapp/quickjs-java/issues/214

### DIFF
--- a/quickjs/common/test/app/cash/quickjs/QuickJsTest.java
+++ b/quickjs/common/test/app/cash/quickjs/QuickJsTest.java
@@ -17,7 +17,6 @@ package app.cash.quickjs;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertThat;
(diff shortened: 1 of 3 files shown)


quickjs/common/test/app/cash/quickjs/QuickJsTest.java
```diff
@@ -85,9 +84,8 @@
     }
   }
 
-  @Ignore("https://github.com/cashapp/quickjs-java/issues/214")
   @Test public void emojiRoundTrip() {
-    assertThat(quickjs.evaluate("\"\uD83D\uDC1D️\""))
-        .isEqualTo("\uD83D\uDC1D️️");
+    assertThat(quickjs.evaluate("\"a\uD83D\uDC1Dcdefg\""))
+        .isEqualTo("a\uD83D\uDC1Dcdefg");
```
swankjesse, line 89: on trunk this fails in two ways. The emoji is encoded wrong, and the remainder of the string is truncated.

```
expected: a🐝cdefg
but was : aðcd
```

JakeWharton (approved):

PR #221, opened 2021-08-04
Support non-ASCII characters everywhere.

swankjesse:


quickjs/common/native/Context.cpp
```diff
@@ -336,18 +327,11 @@ Context::toJavaObject(JNIEnv* env, const JSValueConst& value, bool throwOnUnsupp
 }
 
 jobject Context::eval(JNIEnv* env, jstring source, jstring file) {
-  // Convert to UTF-8 and add a trailing '\u0000'.
-  const jbyteArray sourceCodeUtf8BytesObject = static_cast<jbyteArray>(env->CallObjectMethod(source, stringGetBytes, stringUtf8));
-  size_t sourceLength = env->GetArrayLength(sourceCodeUtf8BytesObject);
-  jbyte* sourceCodeUtf8Bytes = env->GetByteArrayElements(sourceCodeUtf8BytesObject, NULL);
-  std::string sourceCodeString = std::string(reinterpret_cast<char*>(sourceCodeUtf8Bytes), sourceLength);
-  sourceCodeString += "\u0000";
```
JakeWharton, line 344: Is this safe to lose? I thought QuickJS required it or something.

JakeWharton (approved):

quickjs/common/native/Context.cpp
swankjesse: In an earlier iteration of my code it crashed without it. But at a certain point I switched to `std::string` and its `c_str()` function adds a trailing '\u0000'.

PR #223, opened 2021-08-05
Support creating UInt8Arrays from Java byte[] arrays

swankjesse:
Previously we could decode a user-provided Uint8Array but we couldn't create one.


JakeWharton (approved):
I don't think there's any test that actually asserts a UInt8Array is created in JS. They all roundtrip back to Java or use null. I think just one with a `typeof array === "UInt8Array"` would ensure we are producing the expected type in JS and not some other representation.

swankjesse:
Good call. I used `array.constructor.name` because `typeof array` returned `"object"`.

PR #224, opened 2021-08-08
KtBridge: call Kotlin/JS from Kotlin/JVM

swankjesse:
This is the first step in a more ambitious project.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -42,6 +42,9 @@ jobs:
 
       - run: ./gradlew :quickjs:jvm:build
 
+      - name: KtBridge
+        run: ./gradlew :ktbridge:check
+
       - uses: actions/upload-artifact@v2
         with:
           name: jvm-native-libraries
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,7 @@ buildscript {
     classpath 'com.android.tools.build:gradle:4.2.1'
     classpath 'net.ltgt.gradle:gradle-errorprone-plugin:2.0.1'
     classpath 'com.vanniktech:gradle-maven-publish-plugin:0.14.2'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.20"
   }
 }
 
@@ -27,7 +28,8 @@ ext {
       'duktape': 'com.squareup.duktape:duktape-android:1.4.0',
       'junit': 'junit:junit:4.13.2',
       'truth': 'com.google.truth:truth:1.0',
-      'okio': 'com.squareup.okio:okio:1.14.1',
+      'okio': 'com.squareup.okio:okio:2.10.0',
+      'okioMultiplatform': 'com.squareup.okio:okio-multiplatform:2.10.0',
   ]
 }
 
--- a/ktbridge/build.gradle
+++ b/ktbridge/build.gradle
@@ -1,0 +1,40 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+apply plugin: 'com.vanniktech.maven.publish'
+
+kotlin {
+  jvm {
+    withJava()
+  }
+
+  js {
+    browser {
+    }
+  }
+
+  sourceSets {
+    commonMain {
+      dependencies {
+        api deps.okioMultiplatform
+      }
+    }
+    jvmMain {
+      dependencies {
+        api deps.androidx.annotation
+        api project(":quickjs:jvm")
+      }
+    }
+    jvmTest {
+      dependencies {
+        implementation project(":ktbridge:testing")
+        implementation deps.junit
+        implementation deps.truth
+      }
+    }
+  }
+}
+
+tasks {
+  jvmTest {
+    dependsOn(":ktbridge:testing:jsBrowserProductionWebpack")
+  }
+}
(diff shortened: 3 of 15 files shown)


swankjesse:
Next steps:
 * [x] Generate Java so we don't need a dynamic proxy
 * [x] Generate JS so we don't need to use `@JsName` and so that functions can be overloaded
 * [ ] Integrate either kotlinx.serialization or Wire for a built-in JsAdapter
 * [ ] Support suspend functions
 * [ ] Bridge from JVM back to JS

build.gradle
```diff
@@ -5,11 +5,15 @@ buildscript {
     mavenCentral()
     google()
     gradlePluginPortal()
+    maven {
+      url "https://plugins.gradle.org/m2/"
```
JakeWharton: We already have this one line up

ktbridge/build.gradle
```diff
@@ -0,0 +1,40 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+apply plugin: 'com.vanniktech.maven.publish'
+
+kotlin {
+  jvm {
+    withJava()
+  }
```
JakeWharton, line 7:
```suggestion
  jvm()
```

ktbridge/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+
+  js {
+    browser {
+    }
+  }
+
+  sourceSets {
+    commonMain {
+      dependencies {
+        implementation deps.okioMultiplatform
```
JakeWharton:
```suggestion
        api deps.okioMultiplatform
```

ktbridge/testing/build.gradle
```diff
@@ -0,0 +1,33 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+
+kotlin {
+  jvm {
+    withJava()
+  }
```
JakeWharton:
```suggestion
  jvm()
```

ktbridge/testing/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+  js {
+    browser {
+    }
+  }
+
+  sourceSets {
+    commonMain {
+      dependencies {
+        implementation deps.okioMultiplatform
```
JakeWharton:
```suggestion
```
Not needed after the switch to `api` above

ktbridge/testing/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+        implementation project(":ktbridge")
+      }
+    }
+    jsMain {
+      dependencies {
+        implementation project(":ktbridge")
+      }
+    }
```
JakeWharton:
```suggestion
```
commonMain dep will be available to all platforms

JakeWharton (commented):
Too hard to review on mobile

build.gradle
swankjesse: Ooooh ouch. Fixed.

ktbridge/build.gradle
swankjesse: Fixed.

ktbridge/build.gradle
swankjesse: Fixed!

ktbridge/testing/build.gradle
swankjesse: Fixed.

ktbridge/testing/build.gradle
swankjesse: Fixed.

ktbridge/testing/build.gradle
swankjesse: Nice, yes of course. Fixed.

ktbridge/build.gradle
swankjesse: Hmmm... that didn’t work.

```
./gradlew clean ktbridge:tasks

> Task :ktbridge:tasks FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':ktbridge:tasks'.
> Could not create task ':ktbridge:javadocsJar'.
   > Could not create task of type 'JavadocsJar'.
      > Task with name 'javadoc' not found in project ':ktbridge'.
```

.github/workflows/release.yaml
```diff
@@ -42,6 +42,9 @@ jobs:
 
       - run: ./gradlew :quickjs:jvm:build
 
+      - name: KtBridge
+        run: ./gradlew :ktbridge:check
```
swankjesse, line 46: I added this. I don’t have opinions on how these are organized so if you’d prefer a separate job that works for me.

JakeWharton:
Just FYI I didn't actually review anything but the build files yet.

swankjesse:
Sounds good. Lots of this is bootstrapping, and I expect the API to change significantly before anything safe to release. Also I'm open to splitting this off into another project if you'd prefer that.

JakeWharton:
(Still having not reviewed...)

I'm tempted to rewrite the existing Java into Kotlin and move this into the core. Make it one of the core value propositions. It also opens up the ability to support `suspend` functions there as well as types like UByteArray, UInt, etc.

swankjesse:
One more thing that'd unlock: we could expand this project to integrate QuickJS on iOS via Kotlin/Native. Seems like a big win to me!

ktbridge/build.gradle
JakeWharton: Need to apply the Dokka plugin since this is Kotlin

PR #227, opened 2021-08-16
Convert Gradle builds to KTS

swankjesse:
I'm about to move the plugin into its own subproject and it's
way easier for me to work in KTS.


oldergod (approved):

build.gradle.kts
```diff
@@ -0,0 +1,90 @@
...
+  dependencies {
+    classpath("com.android.tools.build:gradle:4.2.1")
+    classpath("net.ltgt.gradle:gradle-errorprone-plugin:2.0.1")
+    classpath("com.vanniktech:gradle-maven-publish-plugin:0.14.2")
+    classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.20")
+  }
+}
+
+plugins {
+  id("com.gradle.plugin-publish") version "0.12.0" apply false
```
JakeWharton, line 22:
```suggestion
```
Same as other PR, we don't use this

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -0,0 +1,9 @@
+object Dependencies {
```
JakeWharton, line 1: Ewwwww buildSrc for dependencies is 👎

quickjs/android/build.gradle.kts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+dependencies {
+  api(Dependencies.androidxAnnotation)
+
+  androidTestImplementation(Dependencies.androidxTestRunner)
+  androidTestImplementation(Dependencies.truth)
+}
+
+fun quickJsVersion(): File {
```
JakeWharton, line 72: This needs to return a String

JakeWharton (approved):
Screen Shot 2021-08-18 at 11 02 59 AM

swankjesse:
Merged manually.

quickjs/android/build.gradle.kts
swankjesse: Oooh yuck, I replaced a file’s contents with its path. Ugh. Fixed.

build.gradle.kts
swankjesse: Fixed.

buildSrc/src/main/kotlin/Dependencies.kt
swankjesse: Yeah, let’s switch to the Gradle 7 thing?

PR #229, opened 2021-08-17
Use expect/actual to expose APIs

swankjesse:
Rename BridgedCall to InboundCall

Emulate the JS API on the JVM more explicitly so the plugin test works
more naturally

Move the name of the webpack module name into the JVM stub declaration

Split the InternalBridge and BridgeToJs interfaces

### DIFF
--- a/ktbridge/README.md
+++ b/ktbridge/README.md
@@ -19,7 +19,8 @@ kotlin.js.compiler=ir
 Define a Shared API
 -------------------
 
-Create an interface and value types that'll be shared between platforms.
+In your `commonMain` directory, create an interface and value types that'll be shared between
+platforms. Use an `expect` val to declare a shared global instance.
 
 ```kotlin
 interface EchoService {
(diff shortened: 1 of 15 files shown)


ktbridge/README.md
```diff
@@ -19,7 +19,8 @@ kotlin.js.compiler=ir
 Define a Shared API
 -------------------
 
-Create an interface and value types that'll be shared between platforms.
+In your `commonMain` directory, create an interface and value types that'll be shared between
+platforms. Use an `expect` val to declare a shared global instance.
```
JakeWharton, line 23: A global instance will not work for us. We have multiple VMs and can reload a VM based on feature flags or new copies of the JS / bytecode being downloaded without restarting the app.

JakeWharton: Even on the JS-side I don't think it's safe to assume a single, global instance in practice.

JakeWharton (approved):

PR #234, opened 2021-08-20
Convert two test classes to Kotlin

JakeWharton:
Also drop Truth in (long term) anticipation of running on Kotlin/Native.


quickjs/common/test/app/cash/quickjs/Utf8Test.kt
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @Test fun nonAsciiInProxyInputAndOutput() {
+    quickjs.evaluate("""
+      |var formatter = {
+      |  format: function(message) {
+      |    return '(' + message + ', ' + message + ')';
+      |  }
+      |};
+      |""".trimMargin())
```
swankjesse, line 63: Yay multiline strings

swankjesse (approved):

PR #243, opened 2021-08-24
Merge ktbridge into QuickJS

JakeWharton:
No ~real~ big changes, just moving source files around.


quickjs/src/jniTest/kotlin/app/cash/quickjs/ktbridge/KtBridgeTest.kt
```diff
@@ -40,7 +40,7 @@ class KtBridgeTest {
     quickjs.close()
   }
   
-  @Test fun `jvm call js service`() {
+  @Test fun jvmCallJsService() {
```
swankjesse, line 36: Shall we follow this style for JVM tests? I haven’t been but am happy to

swankjesse (approved):

quickjs/src/jniTest/kotlin/app/cash/quickjs/ktbridge/KtBridgeTest.kt
JakeWharton: Doesn't matter. I personally find it easier to just not use spaces everywhere so that I don't have to worry about where the test is running and whether it supports spaces.

JakeWharton:
The failure here is how the JS file is loaded in the test. It's relying on relative path on the filesystem. I need to change it to load from the jar and embed it in the jar/apk so that it also works on Android.

PR #245, opened 2021-08-25
Support null parameter values and results

swankjesse:


ktbridge/build.gradle.kts
```diff
@@ -37,6 +37,10 @@ kotlin {
   }
 }
 
+kotlin.sourceSets.all {
+  languageSettings.useExperimentalAnnotation("kotlin.ExperimentalStdlibApi")
```
JakeWharton: We need to propagate this as an annotation on our APIs rather than suppress it globally for our sources. Otherwise we hide our usage and someone upgrading their Kotlin to a newer version may see broken behavior.

JakeWharton (approved):

ktbridge/build.gradle.kts
swankjesse: This is for `typeOf<T>()` . . . . is it worth the grind of making everything in our universe experimental? I think this thing is pretty stable, and if it breaks it’ll break at compile-time.

JakeWharton: Ah, is it for our own internal usage only? In that case I would actually prefer if we did targeted `@OptIn`s at the call-sites just so we don't accidentally use something that might actually break.

swankjesse: Done. Yep, that’s what I really wanted.

PR #246, opened 2021-08-25
Read testing.js from resources

JakeWharton:
Rather than directly from the filesystem relying on working directory and relative paths.

### DIFF
--- a/ktbridge/build.gradle.kts
+++ b/ktbridge/build.gradle.kts
@@ -6,13 +6,17 @@ plugins {
   id("org.jetbrains.dokka")
 }
 
+val copyTestingJs = tasks.register<Copy>("copyTestingJs") {
+  destinationDir = buildDir.resolve("generated/testingJs")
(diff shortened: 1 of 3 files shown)


ktbridge/build.gradle.kts
```diff
@@ -4,13 +4,17 @@ plugins {
   id("org.jetbrains.dokka")
 }
 
+val copyTestingJs = tasks.register<Copy>("copyTestingJs") {
+  destinationDir = buildDir.resolve("generated/testingJs")
+  from("testing/build/distributions/testing.js")
+  dependsOn(":ktbridge:testing:jsBrowserProductionWebpack")
```
JakeWharton, line 12: Still not perfect since we're depending on implementation detail paths, task names, and doing both across project boundaries, but we can "fix" (i.e., adhere to Gradle's ridiculous ways of doing things "correctly") later.

JakeWharton:
Merging to unblock myself, although this is mostly extracted from #243

PR #250, opened 2021-08-27
Support generic parameters and return types

swankjesse:


swankjesse:
Motivation is so we can do callbacks in implementing coroutines.

JakeWharton (approved):

PR #251, opened 2021-08-29
Implement support for suspending calls

swankjesse:
This is pretty basic. It doesn't yet support:
 - calls that fail with an exception
 - jobs that are canceled
 - execution on JavaScript

### DIFF
--- a/quickjs/build.gradle.kts
+++ b/quickjs/build.gradle.kts
@@ -31,8 +31,15 @@ val versionWriterTaskProvider = tasks.register("writeVersion", VersionWriterTask
 
 val copyTestingJs = tasks.register<Copy>("copyTestingJs") {
   destinationDir = buildDir.resolve("generated/testingJs")
-  from(projectDir.resolve("testing/build/distributions/testing.js"))
-  dependsOn(":quickjs:testing:jsBrowserProductionWebpack")
+  if (true) {
+    // Production, which is minified JavaScript.
+    from(projectDir.resolve("testing/build/distributions/testing.js"))
+    dependsOn(":quickjs:testing:jsBrowserProductionWebpack")
+  } else {
+    // Development, which is not minified and has useful stack traces.
+    from(projectDir.resolve("testing/build/developmentExecutable/testing.js"))
+    dependsOn(":quickjs:testing:jsBrowserDevelopmentWebpack")
+  }
 }
 
 kotlin {
@@ -46,6 +53,7 @@ kotlin {
     val commonMain by getting {
       kotlin.srcDir(versionWriterTaskProvider)
       dependencies {
+        api(Dependencies.kotlinxCoroutines)
         api(Dependencies.okioMultiplatform)
       }
     }
--- a/quickjs/src/commonMain/kotlin/app/cash/quickjs/KtBridge.kt
+++ b/quickjs/src/commonMain/kotlin/app/cash/quickjs/KtBridge.kt
@@ -15,20 +15,41 @@
  */
 package app.cash.quickjs
 
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+
 // TODO(jwilson): merge with QuickJs?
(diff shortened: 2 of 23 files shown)


quickjs/src/commonMain/kotlin/app/cash/quickjs/KtBridge.kt
```diff
@@ -15,21 +15,42 @@
...
+
+    override fun invokeSuspending(
+      instanceName: String,
+      funName: String,
+      encodedArguments: ByteArray,
+      callbackName: String
+    ) {
+      val handler = inboundHandlers[instanceName] ?: error("no handler for $instanceName")
+      // TODO(jwilson): confirm how we bootstrap a coroutine on an inbound call.
+      GlobalScope.launch {
```
swankjesse, line 46: This is almost certainly wrong but I don’t know what’s right.

JakeWharton: We will likely have to accept a CoroutineScope when creating the QuickJs instance.

JakeWharton:
Still reviewing. I think you're getting lucky through the use of GlobalScope that we don't invoke a continuation in the same stack frame as the suspend function was called.

swankjesse:
Agreed on the GlobalScope / Scope thing. Accepting a Scope when we create the `QuickJs` is nice – can we automatically close when the scope completes?

JakeWharton:
If the CoroutineScope has a Job you can attach a completion listener

quickjs/build.gradle.kts
```diff
@@ -31,8 +31,15 @@ val versionWriterTaskProvider = tasks.register("writeVersion", VersionWriterTask
 
 val copyTestingJs = tasks.register<Copy>("copyTestingJs") {
   destinationDir = buildDir.resolve("generated/testingJs")
-  from(projectDir.resolve("testing/build/distributions/testing.js"))
-  dependsOn(":quickjs:testing:jsBrowserProductionWebpack")
+  if (true) {
+    // Production, which is minified JavaScript.
+    from(projectDir.resolve("testing/build/distributions/testing.js"))
+    dependsOn(":quickjs:testing:jsBrowserProductionWebpack")
```
JakeWharton, line 37: Do we ever actually care about this for tests?

quickjs/testing/src/jvmMain/kotlin/app/cash/quickjs/testing/suspendingJvm.kt
```diff
@@ -0,0 +1,37 @@
...
+import app.cash.quickjs.KtBridge
+
+val KtBridge.jsSuspendingEchoService: SuspendingEchoService
+  get() = get("jsSuspendingEchoService", EchoJsAdapter)
+
+class JvmSuspendingEchoService(
+  private val greeting: String
+) : SuspendingEchoService {
+  override suspend fun suspendingEcho(request: EchoRequest): EchoResponse {
+    return EchoResponse("$greeting from the suspending JVM, ${request.message}")
```
JakeWharton, line 27: I'm a bit worried that none of the tests actually suspend. This means that the result is delivered synchronously through the return value rather than through the Continuation parameter. Once we get `setTimeout`, all of these should add a `yield()` or `delay(1)` or something.

JakeWharton (approved):
Mostly just unblocking. I expect a lot of this to evolve.

quickjs/build.gradle.kts
swankjesse: I’m undecided on whether I want tests to exercise production JS or development JS. Having development JS is really nice when things crash!

quickjs/testing/src/jvmMain/kotlin/app/cash/quickjs/testing/suspendingJvm.kt
swankjesse: 100% agree.

PR #253, opened 2021-08-29
Implement setTimeout()

swankjesse:
This is wired up right but requires changes to QuickJs to either
post all work to a single-threaded CoroutineDispatcher or introduce
synchronized statements.

### DIFF
--- a/quickjs/src/jsMain/kotlin/app/cash/quickjs/platformJs.kt
+++ b/quickjs/src/jsMain/kotlin/app/cash/quickjs/platformJs.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


quickjs/src/jsMain/kotlin/app/cash/quickjs/platformJs.kt
```diff
@@ -0,0 +1,70 @@
...
+  val jsPlatform = RealJsPlatform(hostPlatform)
+
+  // Create global functions for JavaScript callers.
+  js("""
+    globalThis.setTimeout = function(handler, delay, args) {
+      return jsPlatform.setTimeout(handler, delay, args);
+    };
+    globalThis.clearTimeout = function(timeoutID) {
+      return jsPlatform.clearTimeout(timeoutID);
+    };
```
JakeWharton, line 64:
```suggestion
    globalThis.setTimeout = jsPlatform.setTimeout;
    globalThis.clearTimeout = jsPlatform.clearTimeout;
```
Does this work? It should.

JakeWharton (approved):

quickjs/src/jsMain/kotlin/app/cash/quickjs/platformJs.kt
swankjesse: It didn’t work for me when I attempted it. I think the issue is the `this` reference which doesn’t come through when I just assign the functions. It fails like so:
```
cannot read property '_nextTimeoutId' of undefined
```

PR #254, opened 2021-08-30
Debug API to get JS memory usage

swankjesse:


quickjs/src/native/Context.cpp
```diff
@@ -77,6 +78,7 @@ Context::Context(JNIEnv* env)
       doubleGetValue(env->GetMethodID(doubleClass, "doubleValue", "()D")),
       stringGetBytes(env->GetMethodID(stringClass, "getBytes", "(Ljava/lang/String;)[B")),
       stringConstructor(env->GetMethodID(stringClass, "<init>", "([BLjava/lang/String;)V")),
+      memoryUsageConstructor(env->GetMethodID(memoryUsageClass, "<init>", "(JJJJJJJJJJJJJJJJJJJJJJJJJJ)V")),
```
swankjesse, line 81: Gotta scrub some J’s

quickjs/src/jniTest/kotlin/app/cash/quickjs/MemoryUsageTest.kt
```diff
@@ -0,0 +1,116 @@
...
+
+  @Test fun definePropertyIncreasesPropertiesCount() {
+    val diff = diffMemoryUsage {
+      quickjs.evaluate(
+        """
+        globalThis.hello = 'hello';
+        """
+      )
+    }
+    assertEquals(diff.stringsCount, 0L) // Why isn't this 1?
```
JakeWharton, line 47: Maybe it's like const pool vs. heap? Try `String('hello')`

quickjs/src/native/Context.cpp
JakeWharton: 🙈

quickjs/src/jniMain/kotlin/app/cash/quickjs/MemoryUsage.kt
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.quickjs
+
+/** Introspect QuickJS for its current memory usage. */
+class MemoryUsage(
```
JakeWharton:
```suggestion
import androidx.annotation.Keep

/** Introspect QuickJS for its current memory usage. */
@Keep // Fully-qualified type and constructor used from native code.
class MemoryUsage(
```

quickjs/src/native/Context.cpp
```diff
@@ -66,6 +66,7 @@ Context::Context(JNIEnv* env)
       doubleClass(static_cast<jclass>(env->NewGlobalRef(env->FindClass("java/lang/Double")))),
       objectClass(static_cast<jclass>(env->NewGlobalRef(env->FindClass("java/lang/Object")))),
       stringClass(static_cast<jclass>(env->NewGlobalRef(env->FindClass("java/lang/String")))),
+      memoryUsageClass(static_cast<jclass>(env->NewGlobalRef(env->FindClass("app/cash/quickjs/MemoryUsage")))),
```
JakeWharton, line 69: Is there a way to make this lookup fault tolerant? Like return null if type cannot be found?

That would allow us to use a more targeted shrinker rule and allow this class to be removed if unused.

quickjs/src/native/Context.cpp
```diff
@@ -163,6 +165,42 @@ jbyteArray Context::compile(JNIEnv* env, jstring source, jstring file) {
   return result;
 }
 
+jobject Context::memoryUsage(JNIEnv* env) {
+  JSMemoryUsage jsMemoryUsage;
```
JakeWharton, line 169: I'm assuming this is stack allocated and doesn't need any kind of explicit free-ing.

JakeWharton (approved):

quickjs/src/native/Context.cpp
swankjesse: That’s a great idea. Lemme make a note of it!

quickjs/src/native/Context.cpp
swankjesse: You got it.

quickjs/src/native/Context.cpp
swankjesse: https://github.com/cashapp/quickjs-java/issues/255

swankjesse: One really simple strategy for this – just defer looking up these types until we need ’em.

PR #257, opened 2021-09-02
Make Zipline the primary public API of this project

swankjesse:
This is a big broad refactoring that changes where types are and
how they're hooked up.

The biggest behavior change is introducing a Dispatcher as a core
part of the API, and adopting that where necessary.


swankjesse:
This PR is already huge so I wanted to call out the obvious loose ends I need to do in follow-up. I added these to our tracking issue.

 * Make `Zipline.close()`. Right now closing the QuickJS and the dispatcher is super awkward.
 * Rename packages and modules
 * Shrink the QuickJS API significantly!
 * Update documentation!
 * Add Console is in the base JavaScript environment (I believe Coroutines uses this!)

quickjs/src/jniMain/kotlin/app/cash/quickjs/ZiplineJvm.kt
```diff
@@ -0,0 +1,101 @@
...
+
+internal class ZiplineJvm(
+  override val quickJs: QuickJs,
+  override val dispatcher: CoroutineDispatcher,
+) : Zipline(), HostPlatform {
+  private val ktBridge: KtBridge
+  private val jsPlatform: JsPlatform
+
+  init {
+    val jsInboundBridge = object : InternalBridge {
```
JakeWharton: Implement this on the `ZipelineJvm` directly

quickjs/src/jniTest/kotlin/app/cash/quickjs/KtBridgeSuspendingTest.kt
```diff
@@ -17,63 +17,45 @@ package app.cash.quickjs
...
 import app.cash.quickjs.testing.prepareSuspendingJvmBridges
 import com.google.common.truth.Truth.assertThat
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.ExecutorCoroutineDispatcher
 import org.junit.After
 import org.junit.Before
-import org.junit.Ignore
 import org.junit.Test
 
-@Ignore("Need to coordinate mutual exclusion")
```
swankjesse, line 31: Happy-path suspending really works now!

quickjs/src/jniTest/kotlin/app/cash/quickjs/SetTimeoutTest.kt
```diff
@@ -16,36 +16,45 @@
...
 
 import kotlin.test.assertEquals
+import kotlinx.coroutines.ExecutorCoroutineDispatcher
+import kotlinx.coroutines.delay
 import org.junit.After
 import org.junit.Before
-import org.junit.Ignore
 import org.junit.Test
 
-@Ignore("Need to coordinate mutual exclusion")
```
swankjesse, line 24: Happy-path setTimeout() really works now!

quickjs/src/jniMain/kotlin/app/cash/quickjs/Zipline.kt
```diff
@@ -0,0 +1,76 @@
...
+    block()
+  }
+}
+
+/** Enqueue [block] to run on the JavaScript dispatcher and return immediately. */
+inline fun Zipline.launch(crossinline block: suspend Zipline.() -> Unit) {
+  CoroutineScope(EmptyCoroutineContext).launch(dispatcher) {
+    block()
+  }
+}
```
JakeWharton: I find both of these highly suspect and don't think either is something I want to expose in public API. Either the caller is in a coroutine scope in which they can `launch`/`async`, they're in a suspend function where they can `withContext`, or they're not in which case they can `runBlocking`.

These feel like helpers for our tests.

quickjs/src/jniMain/kotlin/app/cash/quickjs/Zipline.kt
```diff
@@ -0,0 +1,76 @@
...
+
+  @PublishedApi
+  internal abstract fun set(name: String, service: InboundService<*>)
+
+  companion object {
+    fun create(
+      dispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher(),
+    ): Zipline {
+      // TODO(jwilson): avoid runBlocking if I’m already on our dispatcher??
+      return kotlinx.coroutines.runBlocking(dispatcher) {
```
JakeWharton: I don't see why it's needed at all. No one can interact with the instance until it's returned so we're safe to create and initialize on the caller thread.

quickjs/src/jniMain/kotlin/app/cash/quickjs/Zipline.kt
```diff
@@ -0,0 +1,76 @@
...
+import app.cash.quickjs.internal.bridge.InboundService
+import app.cash.quickjs.internal.bridge.OutboundClientFactory
+import java.util.concurrent.Executors
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.launch
+
+actual abstract class Zipline {
```
JakeWharton: Wondering if we should use this name in the API or just have it be the library and package name and use "InterfaceBridge" or something here instead.

quickjs/src/jniMain/kotlin/app/cash/quickjs/ZiplineJvm.kt
```diff
@@ -0,0 +1,101 @@
...
+  ): T {
+    return ktBridge.get(name, outboundClientFactory)
+  }
+
+  override fun set(name: String, handler: InboundService<*>) {
+    ktBridge.set(name, handler)
+  }
+
+  override fun setTimeout(timeoutId: Int, delayMillis: Int) {
+    launch {
```
JakeWharton: I'm wondering if this should have `start=UNDISPATCHED` so that we get to the delay before the function returns. But I'm not sure what the expected behavior is around a 0 delayMillis for `setTimeout` and what the `delay(0)` actual behavior is.

quickjs/src/jsMain/kotlin/app/cash/quickjs/ZiplineJs.kt
```diff
@@ -0,0 +1,133 @@
...
+internal val THE_ONLY_ZIPLINE: Zipline = ZiplineJs()
+
+private class ZiplineJs : Zipline(), JsPlatform {
+  private val ktBridge: KtBridge
+  private val hostPlatform: HostPlatform
+  private var nextTimeoutId = 1
+  private val jobs = mutableMapOf<Int, Job>()
+
+  init {
+    val jsOutboundBridge = object : InternalBridge {
```
JakeWharton: Implement on `ZiplineJs` directly

quickjs/src/jsMain/kotlin/app/cash/quickjs/ZiplineJs.kt
```diff
@@ -0,0 +1,133 @@
...
+import app.cash.quickjs.internal.bridge.InternalBridge
+import app.cash.quickjs.internal.bridge.KtBridge
+import app.cash.quickjs.internal.bridge.OutboundClientFactory
+import kotlinx.coroutines.Dispatchers
+
+/**
+ * We declare this as a top-level property because that causes this object to be constructed when
+ * the code is loaded, and we rely on that for the side-effects performed in the ZiplineJs
+ * constructor.
+ */
```
JakeWharton: You can put side-effects into a `fun main()` which is what I did when I implemented QuickJs on its own companion. That allows elimination of this property and changing `ZiplineJs` to an `object` (or moving it to be directly implemented in `Zipline`'s `companion object`).

JakeWharton (approved):

quickjs/src/jniMain/kotlin/app/cash/quickjs/ZiplineJvm.kt
swankjesse: Good call. Done.

quickjs/src/jsMain/kotlin/app/cash/quickjs/ZiplineJs.kt
swankjesse: Good call. Done.

quickjs/src/jniMain/kotlin/app/cash/quickjs/Zipline.kt
swankjesse: Lemme move it to our testUtil. I think it probably goes away when I add the test dispatcher.

quickjs/src/jniMain/kotlin/app/cash/quickjs/Zipline.kt
swankjesse: Yeah, good call. Done.

quickjs/src/jniMain/kotlin/app/cash/quickjs/Zipline.kt
swankjesse: I really like brand names for entry points. Retrofit, Moshi, Timber, Guice.

quickjs/src/jniMain/kotlin/app/cash/quickjs/ZiplineJvm.kt
swankjesse: I think you’re on to something. Lemme do, but in follow-up with new tests.

quickjs/src/jsMain/kotlin/app/cash/quickjs/ZiplineJs.kt
swankjesse: I couldn’t figure out how to make this work. In particular I don’t know how to get a main() inside the zipline runtime library to load when it’s just a dependency of the testing project.

JakeWharton: `fun main` code should run when the library is loaded because it winds up being top-level in the JS closure. It _should_ behave same as the code which performs this field initialization.

swankjesse: Might be different if there’s a fun main in an included library?

PR #261, opened 2021-09-02
Unbreak tests which have been flaky

swankjesse:
Tests have been failing like this:

    app.cash.quickjs.QuickJsException: stack overflow
        at app.cash.quickjs.QuickJs.evaluate(Native Method)

The initial fix is pretty heavy-handed; I'm just setting the stack size
to 0 which ultimately removes the stack size limit. But this motivates
further investigation.

See: https://github.com/cashapp/quickjs-java/issues/260

Also adopt TestCoroutineDispatcher for more predictable tests.

### DIFF
--- a/quickjs/build.gradle.kts
+++ b/quickjs/build.gradle.kts
@@ -80,6 +80,7 @@ kotlin {
       resources.srcDir(copyTestingJs)
       dependencies {
         implementation(Dependencies.truth)
+        implementation(Dependencies.kotlinxCoroutinesTest)
         implementation(project(":quickjs:testing"))
       }
     }
@@ -91,6 +92,7 @@ android {
 
   defaultConfig {
     minSdkVersion(18)
+    multiDexEnabled = true
 
     testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
@@ -104,6 +106,13 @@ android {
         cFlags("-fstrict-aliasing", "-DCONFIG_VERSION=\\\"${quickJsVersion()}\\\"")
         cppFlags("-fstrict-aliasing", "-DCONFIG_VERSION=\\\"${quickJsVersion()}\\\"")
       }
+    }
+
+    packagingOptions {
+      // We get multiple copies of some license files via JNA, which is a transitive dependency of
+      // kotlinx-coroutines-test. Don't fail the build on these duplicates.
+      exclude("META-INF/AL2.0")
+      exclude("META-INF/LGPL2.1")
     }
   }
 
@@ -159,6 +168,7 @@ dependencies {
   androidTestImplementation(Dependencies.junit)
   androidTestImplementation(Dependencies.androidxTestRunner)
   androidTestImplementation(Dependencies.truth)
+  androidTestImplementation(Dependencies.kotlinxCoroutinesTest)
   androidTestImplementation(project(":quickjs:testing"))
 
   add(PLUGIN_CLASSPATH_CONFIGURATION_NAME, project(":quickjs-kotlin-plugin"))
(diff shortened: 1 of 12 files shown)


quickjs/build.gradle.kts
```diff
@@ -105,6 +107,13 @@ android {
         cppFlags("-fstrict-aliasing", "-DCONFIG_VERSION=\\\"${quickJsVersion()}\\\"")
       }
     }
+
+    packagingOptions {
+      // We get multiple copies of some license files via JNA, which is a transitive dependency of
+      // kotlinx-coroutines-test. Don't fail the build on these duplicates.
+      pickFirst("META-INF/AL2.0")
+      pickFirst("META-INF/LGPL2.1")
```
JakeWharton: We generally use `exclude` since these are useless files

quickjs/src/jniTest/kotlin/app/cash/quickjs/testUtil.kt
```diff
@@ -39,17 +36,3 @@ fun Zipline.loadTestingJs() {
     .use(BufferedReader::readText)
   quickJs.evaluate(testingJs, "testing.js")
 }
-
-/** Execute [block] on the JavaScript dispatcher and wait for it to return. */
-inline fun <T> Zipline.runBlocking(crossinline block: suspend Zipline.() -> T): T {
```
swankjesse, line 44: # 🎉

quickjs/src/jniTest/kotlin/app/cash/quickjs/CallBridgesTest.kt
```diff
@@ -110,7 +105,7 @@ internal class CallBridgesTest {
   }
 
   @Test
-  fun suspendingRequestAndResponse() {
+  fun suspendingRequestAndResponse() = dispatcher.runBlockingTest {
```
JakeWharton: We've banned runBlockingTest internally because it has some really, really bad behavior. Like it's 100% broken for some coroutine shapes. Normally I recommend `runBlocking(dispatcher)` in place.

If it works for now we can leave it. On the first problem I am going to replace every single usage. It's a bad API because it's just a timebomb waiting to happen.

quickjs/src/jniMain/kotlin/app/cash/quickjs/ziplineJni.kt
```diff
@@ -54,9 +52,11 @@ actual abstract class Zipline {
 
   companion object {
     fun create(
-      dispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher(),
+      dispatcher: CoroutineDispatcher,
```
JakeWharton, line 55: And just like that you've made this whole file compatible with Kotlin/Native. Saved me an expect/actual function to produce a default argument value.

JakeWharton (approved):

quickjs/build.gradle.kts
swankjesse: Done.

quickjs/src/jniMain/kotlin/app/cash/quickjs/ziplineJni.kt
swankjesse: Nice.

quickjs/src/jniTest/kotlin/app/cash/quickjs/CallBridgesTest.kt
swankjesse: Good call. Switched.

JakeWharton:
Need to rebase on this

PR #262, opened 2021-09-02
Zipline.close()

swankjesse:


quickjs/src/jniMain/kotlin/app/cash/quickjs/ziplineJni.kt
```diff
@@ -50,25 +51,34 @@ actual abstract class Zipline {
...
       // TODO(jwilson): figure out a 512 KiB limit caused intermittent stack overflow failures.
       quickJs.maxStackSize = 0L
-      return ZiplineJvm(quickJs, dispatcher)
+      return ZiplineJni(quickJs, dispatcher)
     }
   }
 }
 
-private class ZiplineJvm(
+private class ZiplineJni(
```
JakeWharton, line 74: I think this is actually a `QuickJsZipline`. With a small tweak to extract the call to the logger, we can reuse the entirety of this implementation on Kotlin/Native.

JakeWharton (approved):

PR #263, opened 2021-09-02
Start preparing for native targets

JakeWharton:
Split the JNI implementation from the existence of a JS engine.

Just wanted to get this started. It will require a few steps so might as well try to land them as incrementally as possible.


quickjs/README.md
```diff
@@ -0,0 +1,21 @@
+# Zipline
+
+## Multiplatform structure
```
swankjesse, line 3: 👨🏻‍🍳 👌🏻

swankjesse (approved):

PR #267, opened 2021-09-03
Rename packages, modules, and artifacts to Zipline

swankjesse:
QuickJS is now a component of the Zipline library.


swankjesse:
This PR is intended to be purely mechanical. I’d like to follow up with more careful PRs to fix docs, etc.

I also need to do the follow-up work to rename the repo.

README.md
```diff
@@ -124,7 +124,7 @@ repositories {
   mavenCentral()
 }
 dependencies {
-  implementation 'app.cash.quickjs:quickjs-android:0.9.2'
+  implementation 'app.cash.zipline:zipline-android:0.9.2'
```
JakeWharton: I would not update the README and let it track the latest release until closer to the first zipline release. These Maven coordinates were already broken after the switch to multiplatform, for example, but I was going to wait until release to fix them.

JakeWharton (approved):

JakeWharton:
The great tragedy of GitHub is that the display branch and the PR integration branch are forced to be the same. I would just leave docs as-is for now.

swankjesse:
Yeah good call on docs. Fixed.

PR #270, opened 2021-09-07
Use kotlinx.serialization to encode and decode

swankjesse:
This makes the API work pretty nice, especially since most types just
work out of the box.


zipline/src/commonMain/kotlin/app/cash/zipline/DefaultZiplineSerializersModule.kt
```diff
@@ -0,0 +1,56 @@
...
+ * val EchoSerializersModule: SerializersModule = SerializersModule {
+ *   include(DefaultZiplineSerializersModule)
+ * }
+ * ```
+ */
+val DefaultZiplineSerializersModule: SerializersModule = SerializersModule {
+  contextual(Throwable::class, ThrowableSerializer)
+}
+
+internal object ThrowableSerializer : KSerializer<Throwable> {
```
swankjesse, line 39: this is a good candidate for polymorphic serialization...

zipline/src/commonMain/kotlin/app/cash/zipline/DefaultZiplineSerializersModule.kt
```diff
@@ -0,0 +1,56 @@
...
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.modules.SerializersModule
+
+/**
+ * All Zipline [SerializersModule] should include this to get serialization coverage of core types
```
swankjesse, line 26: Thinking about this... rather than requiring callers to remember to install our SerializerModule in their SerializerModule, I think the nicer API is to take the caller’s possibly-empty module, and install _that_ in a module that we define that has Throwable.

For follow-up.

zipline/src/commonMain/kotlin/app/cash/zipline/Zipline.kt
```diff
@@ -15,10 +15,12 @@
  */
 package app.cash.zipline
 
+import kotlinx.serialization.modules.SerializersModule
+
 expect abstract class Zipline {
   abstract val engineVersion: String
 
-  fun <T : Any> get(name: String, jsAdapter: JsAdapter): T
+  fun <T : Any> get(name: String, serializersModule: SerializersModule): T
```
swankjesse, line 23: TODO: default this to EmptySerializersModule. Might Just Work with our IR plugin?

adrw (approved):
LG

PR #272, opened 2021-09-10
Upgrade the publishing plugin

swankjesse:
Also rename KtBridgeGradlePlugin to ZiplinePlugin

Also move the KtBridge internals to app.cash.zipline.kotlinir

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,8 +1,11 @@
 import com.android.build.gradle.BaseExtension
+import com.vanniktech.maven.publish.MavenPublishBaseExtension
+import com.vanniktech.maven.publish.SonatypeHost
 import net.ltgt.gradle.errorprone.CheckSeverity
 import net.ltgt.gradle.errorprone.errorprone
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
+import org.jetbrains.dokka.gradle.DokkaTask
 
 buildscript {
   repositories {
@@ -13,7 +16,7 @@ buildscript {
   dependencies {
     classpath("com.android.tools.build:gradle:4.2.1")
     classpath("net.ltgt.gradle:gradle-errorprone-plugin:2.0.1")
-    classpath("com.vanniktech:gradle-maven-publish-plugin:0.14.2")
+    classpath("com.vanniktech:gradle-maven-publish-plugin:0.17.0")
     classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.20")
     classpath("org.jetbrains.kotlin:kotlin-serialization:1.5.20")
     classpath("org.jetbrains.dokka:dokka-gradle-plugin:1.5.0")
@@ -53,14 +56,6 @@ subprojects {
     }
   }
 
-  plugins.withType<SigningPlugin> {
-    extensions.configure<SigningExtension> {
-      val signingKey = findProperty("signingKey") as String?
-      val signingPassword = ""
-      useInMemoryPgpKeys(signingKey, signingPassword)
-    }
-  }
-
   tasks.withType(Test::class).configureEach {
     testLogging {
       if (System.getenv("CI") == "true") {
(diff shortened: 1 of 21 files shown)


JakeWharton:
I would prefer to stay on 0.14.2 until a version newer than 0.17.0 is released with a fix for the hostile env var check.

swankjesse:
Is the hostile env var relevant if we get Github Actions to do all the publishing?

JakeWharton:
Yes unfortunately it fails at configuration time meaning I can't do any builds because I have the "old" vars. Of course I need those for other projects which use an older version or alternative means of releasing to Sonatype.

swankjesse:
Ooooh failing at configuration time is a nonstarter. Got it.

JakeWharton:
Let me inquire for a new release in classic _user_ fashion.

JakeWharton:
Asked

zipline-gradle-plugin/build.gradle.kts
```diff
@@ -21,11 +25,27 @@ buildConfig {
...
+    create("zipline") {
+      id = "app.cash.zipline"
+      displayName = "zipline"
+      description = "Compiler plugin to generate bridges between platforms"
+      implementationClass = "app.cash.zipline.gradle.ZiplinePlugin"
     }
   }
 }
+
+// The plugin marker artifacts don't conform to Maven Central's requirements. Disable publishing
```
JakeWharton: This isn't a problem in licensee. Why is it here?

buildSrc/src/main/kotlin/Ext.kt
```diff
@@ -1,5 +1,5 @@
 object Ext {
   val compileSdkVersion = 29
   val ndkAbiFilters = listOf("x86", "x86_64", "armeabi-v7a", "arm64-v8a")
-  val kotlinPluginId = "app.cash.zipline.ktbridge.plugin"
+  val kotlinPluginId = "app.cash.zipline.kotlinir"
```
JakeWharton: Don't think using "ir" provides much value. In 6 months there will only be IR anyway. Can we do just .kotlin?

build.gradle.kts
```diff
@@ -89,3 +84,50 @@ subprojects {
...
+        licenses {
+          license {
+            name.set("The Apache Software License, Version 2.0")
+            url.set("http://www.apache.org/licenses/LICENSE-2.0.txt")
+            distribution.set("repo")
+          }
+        }
+        developers {
+          developer {
+            id.set("square")
```
JakeWharton:
```suggestion
            id.set("cashapp")
```

build.gradle.kts
```diff
@@ -89,3 +84,50 @@ subprojects {
...
+            name.set("Square, Inc.")
```
JakeWharton:
```suggestion
            name.set("Cash App")
```

JakeWharton (approved):
Approve based on the fact they said there'd be a new plugin version this week.

buildSrc/src/main/kotlin/Ext.kt
swankjesse: I wanted `ir` to convey ASTs and code rewriting more than the IR API in particular. Maybe just saying `kotlin` felt unclear, and `kotlincompilerplugin` too verbose. I’ll go with `kotlin` unless you have other recommendations!

build.gradle.kts
swankjesse: I hope we’re not doing this .. . . so far as I know there’s no corporate entity for `cashapp` and we shouldn’t assign ownership to it?

build.gradle.kts
swankjesse: Same same.

zipline-gradle-plugin/build.gradle.kts
swankjesse: I get it this in Wire:

> Invalid POM: /com/squareup/wire/com.squareup.wire.gradle.plugin/4.0.0-alpha.12/com.squareup.wire.gradle.plugin-4.0.0-alpha.12.pom: Project name missing, Project description missing

Most likely cause is the publish plugin doesn’t create a complete POM for these.

swankjesse: Found it!
https://github.com/square/wire/pull/2066/files

build.gradle.kts
JakeWharton: We do this on almost every cashapp project. There's no legal requirement here or anything. I think it ties it more closely to the GitHub org than anything.

swankjesse: Works for me.

PR #274, opened 2021-09-13
Fix the published version

swankjesse:
This was publishing as 'unspecified' and failing

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,7 +27,12 @@ plugins {
   id("com.github.gmazzo.buildconfig") version "2.1.0" apply false
 }
 
-subprojects {
+apply(plugin = "com.vanniktech.maven.publish.base")
+
+allprojects {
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -27,7 +27,12 @@ plugins {
   id("com.github.gmazzo.buildconfig") version "2.1.0" apply false
 }
 
-subprojects {
+apply(plugin = "com.vanniktech.maven.publish.base")
+
+allprojects {
+  group = project.property("GROUP") as String
+  version = project.property("VERSION_NAME") as String
```
JakeWharton: Let's just inline these two and update the `RELEASING.md`?

JakeWharton (approved):

build.gradle.kts
swankjesse: Hell yes. Done.

PR #281, opened 2021-09-17
Create an API for Handle, and rewrite calls to it

swankjesse:
Strictly speaking this is a more capable API than Zipline.get().
But it's more annoying to set up so I expect it to only be used
in cases when passing Flows etc. around.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/Handle.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/Handle.kt
@@ -1,0 +1,107 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline
+
+import app.cash.zipline.internal.bridge.Endpoint
+import app.cash.zipline.internal.bridge.InboundBridge
+import app.cash.zipline.internal.bridge.InboundCallHandler
+import app.cash.zipline.internal.bridge.OutboundBridge
+import kotlinx.serialization.modules.SerializersModule
+import okio.Closeable
+
+abstract class Handle<T : Any> internal constructor() : Closeable {
+  fun get(serializersModule: SerializersModule): T {
+    error("unexpected call to Handle.get: is the Zipline plugin configured?")
+  }
+
+  @PublishedApi
+  internal abstract fun get(outboundBridge: OutboundBridge<T>): T
+}
+
+fun <T : Any> Handle(serializersModule: SerializersModule, service: T): Handle<T> {
+  error("unexpected call to Handle(): is the Zipline plugin configured?")
+}
+
+// Plugin-rewritten code calls the constructor of this class.
+@PublishedApi
+internal class InboundHandle<T : Any>(
+  private val inboundBridge: InboundBridge<T>
+): Handle<T>() {
+  private var name: String? = null
+  private var endpoint: Endpoint? = null
+
+  internal fun connect(endpoint: Endpoint, name: String): InboundCallHandler {
+    check(this.endpoint == null && this.name == null) { "already connected" }
+    this.name = name
+    this.endpoint = endpoint
+
+    val serializersModule = SerializersModule {
+      include(endpoint.builtInSerializersModule)
+      include(inboundBridge.serializersModule)
+    }
+
+    val context = InboundBridge.Context(serializersModule)
+    val result = inboundBridge.create(context)
+    endpoint.inboundHandlers[name] = result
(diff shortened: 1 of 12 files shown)


zipline/src/jniTest/kotlin/app/cash/zipline/EndpointTest.kt
```diff
@@ -163,4 +164,65 @@ internal class EndpointTest {
...
+    assertNull(responses.poll())
+    assertNull(requests.poll())
+  }
+
+  interface EchoServiceFactory {
+    fun create(greeting: String): Handle<EchoService>
+  }
+
+  @Test
+  fun transmitHandles() {
```
swankjesse, line 202: This is the big payoff. We can instantiate an interface and send handles to it without names, and without directly referencing the Zipline that they connect through.

zipline/src/jsMain/kotlin/app/cash/zipline/ziplineJs.kt
```diff
@@ -100,12 +100,12 @@ private class ZiplineJs : Zipline(), JsPlatform, CallChannel  {
...
     endpoint.set<JsPlatform>(
       name = "app.cash.zipline.jsPlatform",
-      serializersModule = ZiplineSerializersModule,
+      serializersModule = EmptySerializersModule,
       instance = this
     )
     hostPlatform = endpoint.get(
       name = "app.cash.zipline.hostPlatform",
-      serializersModule = ZiplineSerializersModule
+      serializersModule = EmptySerializersModule
```
swankjesse, line 108: one PR I have on my queue – default parameter for `serializersModule`. It’s not implemented already because it interacts with the rewriter.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/CallChannel.kt
```diff
@@ -101,3 +97,36 @@ private object ThrowableSerializer : KSerializer<Throwable> {
...
+) : KSerializer<Handle<T>> {
+  override val descriptor = PrimitiveSerialDescriptor("Handle", PrimitiveKind.STRING)
+
+  override fun serialize(encoder: Encoder, value: Handle<T>) {
+    val name = endpoint.generateName()
+    if (value is InboundHandle<T>) {
+      value.connect(endpoint, name)
+      encoder.encodeString(name)
+    } else {
+      error("serializing an outbound handle is not implemented")
```
swankjesse, line 122: There’s a fancy implementation possible here. . . if you send a Flow back to its origin we can send it back by reference and not by value.

zipline/src/commonMain/kotlin/app/cash/zipline/Handle.kt
```diff
@@ -0,0 +1,107 @@
...
+      include(outboundEndpoint.builtInSerializersModule)
+      include(outboundBridge.serializersModule)
+    }
+
+    val context = OutboundBridge.Context(name, serializersModule, outboundEndpoint)
+    return outboundBridge.create(context)
+  }
+
+  override fun close() {
+    // TODO(jwilson): support detaching outbound handles.
```
swankjesse, line 103: the only reason I haven’t done this already is it needs a bridged call to perform the removal.

zipline/src/commonMain/kotlin/app/cash/zipline/Handle.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    val context = InboundBridge.Context(serializersModule)
+    val result = inboundBridge.create(context)
+    endpoint.inboundHandlers[name] = result
+    return result
+  }
+
+  override fun get(outboundBridge: OutboundBridge<T>): T {
+    // TODO(jwilson): change generated subtypes to populate `T` on the super call.
```
swankjesse, line 63: This is an important ergonomics thing. . . if you’re writing a test you’ll have an `InboundHandle` and will want to call `get()` on it.

zipline/src/commonMain/kotlin/app/cash/zipline/Handle.kt
```diff
@@ -0,0 +1,107 @@
...
+ * limitations under the License.
+ */
+package app.cash.zipline
+
+import app.cash.zipline.internal.bridge.Endpoint
+import app.cash.zipline.internal.bridge.InboundBridge
+import app.cash.zipline.internal.bridge.InboundCallHandler
+import app.cash.zipline.internal.bridge.OutboundBridge
+import kotlinx.serialization.modules.SerializersModule
+import okio.Closeable
```
JakeWharton, line 23: Annoying

zipline/src/commonMain/kotlin/app/cash/zipline/Handle.kt
```diff
@@ -0,0 +1,107 @@
...
+    throw IllegalStateException("get() on inbound handlers not yet implemented")
+  }
+
+  override fun close() {
+    val endpoint = this.endpoint
+    val name = this.name
+    if (endpoint != null && name != null) {
+      val removed = endpoint.inboundHandlers.remove(name)
+      this.endpoint = null
+      this.name = null
```
JakeWharton, line 73: Does this mean you can somehow connect more than once?

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/Handle.kt
swankjesse: I was thinking about that. I don’t think that makes sense and I should explicitly forbid it. Lemme implement that in a follow round with lots of new tests and edge cases.

PR #283, opened 2021-09-17
Initial implementation of Flows over Zipline

swankjesse:
This is a bit rocky, especially the serialization bit. But it
works and unblocks application-layer things.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.channelFlow
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.map
+import kotlinx.serialization.Contextual
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.modules.EmptySerializersModule
+import kotlinx.serialization.modules.SerializersModule
+
+/** A flow reference is a [Flow] that can be transmitted over a Zipline call. */
+@Serializable
+class FlowReference<T> @PublishedApi internal constructor(
+  @Contextual private val referenceFlowReference: ZiplineReference<ReferenceFlow>
+) {
+  private fun getJsonFlow(): Flow<String> {
+    return channelFlow {
+      val referenceFlow = referenceFlowReference.get(EmptySerializersModule)
+      try {
+        val collector: FlowCollector<String> = object : FlowCollector<String> {
+          override suspend fun emit(value: String) {
+            this@channelFlow.send(value)
+          }
+        }
+        val collectorReference = ZiplineReference(EmptySerializersModule, collector)
+        referenceFlow.collectJson(collectorReference)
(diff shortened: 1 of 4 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
```diff
@@ -100,18 +105,20 @@ internal class OutboundCall(
...
-      val callback = RealSuspendCallback(callbackName, continuation, serializer)
-      endpoint.set<SuspendCallback>(callbackName, EmptySerializersModule, callback)
-      val encodedArguments = buffer.readByteArray()
-      endpoint.outboundChannel.invokeSuspending(
-        instanceName,
-        funName,
-        encodedArguments,
-        callbackName
-      )
+    return coroutineScope {
```
JakeWharton: This looks suspicious. What's it for?

zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
```diff
@@ -0,0 +1,84 @@
...
+    return getJsonFlow().transform {
+      emit(Json.decodeFromString(serializer, it))
+    }
+  }
+}
+
+fun <T> Flow<T>.toFlowReference(serializer: KSerializer<T>): FlowReference<T> {
+  val flowOfStrings = transform {
+    emit(Json.encodeToString(serializer, it))
+  }
```
JakeWharton: Avoiding `map`?

zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
```diff
@@ -0,0 +1,84 @@
...
+
+  fun get(serializer: KSerializer<T>): Flow<T> {
+    return getJsonFlow().transform {
+      emit(Json.decodeFromString(serializer, it))
+    }
+  }
+}
+
+fun <T> Flow<T>.toFlowReference(serializer: KSerializer<T>): FlowReference<T> {
```
JakeWharton: I'm torn on whether this is a "to" or an "as" function but leaning toward the latter.

zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
```diff
@@ -0,0 +1,84 @@
...
+import kotlinx.serialization.modules.SerializersModule
+import okio.use
+
+/** A flow reference is a [Flow] that can be transmitted over a Zipline call. */
+@Serializable
+class FlowReference<T> @PublishedApi internal constructor(
+  @Contextual private val referenceFlowReference: ZiplineReference<ReferenceFlow>
+) {
+  private fun getJsonFlow(): Flow<String> {
+    return channelFlow {
```
JakeWharton, line 36: I believe the use of this will break synchronous emission behavior of a Flow, but maybe we're unable to provide that across the bridge anyway?

A test which would catch this looks like this: https://github.com/cashapp/sqldelight/blob/10b5a6605bf2584dbef414b80ecfa601c1c76701/extensions/coroutines-extensions/src/commonTest/kotlin/com/squareup/sqldelight/runtime/coroutines/QueryAsFlowTest.kt#L37-L43 with a source of `flowOf(1)` or something.

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
swankjesse: It’s unnecessary. Removed!

zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
swankjesse: Yeah, map’s a bitter fit. Fixed 2x.

zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
swankjesse: Fixed. Yeah I think that’s right, they mutate together.

zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
swankjesse: When I attempt a non-channel `flow {}` I get this failure:
```
java.lang.Exception: java.lang.IllegalStateException: Flow invariant is violated:
		Emission from another coroutine is detected.
		Child of "coroutine#8":StandaloneCoroutine{Active}@22a13908, expected child of "coroutine#6":BlockingCoroutine{Active}@41f1f8b4.
		FlowCollector is not thread-safe and concurrent emissions are prohibited.
		To mitigate this restriction please use 'channelFlow' builder instead of 'flow'
```
There might be other mitigations but this seemed like the most natural.

PR #284, opened 2021-09-19
Don't encode parameter lists as bytes

swankjesse:
On QuickJS the performance of StringBuilder is quadratic. This impacts lots
of code in Zipline, but decoding large JSON bodies from UTF-8 is particularly
problematic.

Time to call CharArray.concatToString() running QuickJS on my MacBook Pro

| Array Size   | Time (seconds) |
| -----------: | :------------- |
|         100  | 0              |
|       1.000  | 0              |
|      10,000  | 0              |
|     100,000  | 1              |
|   1.000,000  | 24             |
|  10,000,000  | 3600+          |

We will want to solve this in QuickJS, Kotlin, and/or Okio. For now I'm
punting on the problem by not dealing with bytes where it isn't necessary.
As a bonus, the code is simpler.


JakeWharton (approved):
Can the Okio dependency be dropped as a result of this change?

swankjesse:
Yes! I’ve dropped the Okio dependency. The only thing it was being used for was so `ZiplineReference` could implement `Closeable`. Now it doesn’t, it just defines its own `close()` function. That’s slightly worse, but not worth the dependency.

PR #292, opened 2021-09-21
Propagate Zipline CoroutineScope downward into machinery

JakeWharton:
This ensures everything is launching into the same scope and will be properly canceled if the Zipline instance is closed.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
@@ -17,7 +17,6 @@ package app.cash.zipline.internal.bridge
 
 import kotlin.coroutines.Continuation
 import kotlin.coroutines.suspendCoroutine
-import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.json.Json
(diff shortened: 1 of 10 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
```diff
@@ -99,7 +98,7 @@ internal class OutboundCall(
   internal suspend fun <R> invokeSuspending(serializer: KSerializer<R>): R {
     require(callCount++ == parameterCount)
     return suspendCoroutine { continuation ->
-      CoroutineScope(context.endpoint.dispatcher).launch {
+      context.endpoint.scope.launch {
```
JakeWharton, line 101: I am skeptical that this is needed... all we're doing is bookkeeping and sending notifications.

swankjesse: I was specifically trying to handle the caller calling us on the wrong dispatcher. We should either thread-hop or detect mistakes.

swankjesse (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
JakeWharton: Hmm yeah I guess that can happen from JVM. I was thinking in the context of JS calling out.

PR #293, opened 2021-09-21
Move Zipline actual into engineMain in preparation for native

JakeWharton:


zipline/src/engineMain/kotlin/app/cash/zipline/QuickJs.kt
```diff
@@ -15,13 +15,15 @@
...
+import app.cash.zipline.internal.Closeable
+
 /**
  * An EMCAScript (Javascript) interpreter backed by the 'QuickJS' native engine.
  *
  * This class is NOT thread safe. If multiple threads access an instance concurrently it must be
  * synchronized externally.
  */
-expect class QuickJs {
+expect class QuickJs : Closeable {
```
swankjesse: I don’t think public APIs should implement internal interfaces?

swankjesse (approved):

zipline/src/engineMain/kotlin/app/cash/zipline/QuickJs.kt
JakeWharton: Yeah it's a very unfortunate situation. I also think it's pretty shitty that Kotlin forces every library has to have a public Closeable type just to typealias on the JVM. Maybe we assume people aren't using `use` and just don't implement `Closeable`?

PR #294, opened 2021-09-21
Exactly size outbound call parameter list

JakeWharton:


swankjesse (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
```diff
@@ -70,7 +70,7 @@ internal class OutboundCall(
   private val funName: String,
   private val parameterCount: Int,
 ) {
-  private val arguments = mutableListOf<String>()
+  private val arguments = ArrayList<String>(parameterCount * 2)
```
swankjesse, line 73: I was thinking we might eventually have other directives in here. When we do we can adjust the size accordingly.

swankjesse (approved):

PR #303, opened 2021-09-22
Split logging behavior between JVM and Android

JakeWharton:
JVM goes to Logger, Android goes to Log.


JakeWharton:
Gotta move a test. Going to dinner for now though.

swankjesse (approved):

PR #304, opened 2021-09-22
Split Console and EventLoop interfaces

JakeWharton:
Console is actually platform-specific whereas the EventLoop impl can be shared by all platforms.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/platform.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/platform.kt
@@ -15,21 +15,18 @@
  */
 package app.cash.zipline.internal
 
-internal const val hostPlatformName = "zipline/host"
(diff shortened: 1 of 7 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/internal/platform.kt
```diff
@@ -15,21 +15,18 @@
...
 
-/**
- * Functions that are built in to regular JavaScript environments like the browser and NodeJS, but
- * must be added to QuickJS.
- *
- * https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope
- */
-internal interface HostPlatform {
-  // See org.w3c.dom.WindowOrWorkerGlobalScope
+internal interface EventLoop {
```
swankjesse, line 22: Good name.

swankjesse (approved):

PR #307, opened 2021-09-23
Fix ZiplineReference.get() to always work

swankjesse:
Previously this crashed if the ZiplineReference was never sent
to the other runtime.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineReference.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineReference.kt
@@ -54,8 +54,7 @@ internal class InboundZiplineReference<T : Any>(
   }
 
   override fun get(outboundBridge: OutboundBridge<T>): T {
-    // TODO(jwilson): change generated subtypes to populate `T` on the super call.
-    throw IllegalStateException("get() on inbound handlers not yet implemented")
+    return inboundBridge.service
   }
 
   override fun close() {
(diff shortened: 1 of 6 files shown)


zipline/src/jniTest/kotlin/app/cash/zipline/ZiplineReferenceTest.kt
```diff
@@ -41,6 +41,13 @@ internal class ZiplineReferenceTest {
     this.endpointB = endpointB
   }
 
+  @Test
+  fun referenceCanBeUsedWithoutPassingThroughZipline() {
+    val helloService = GreetingService("hello")
+    val reference = ZiplineReference<EchoService>(helloService)
+    assertThat(reference.get()).isSameInstanceAs(helloService)
```
swankjesse, line 48: This is most useful for tests, where we don’t necessarily need Zipline to confirm the different components work as expected.

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineReference.kt
```diff
@@ -54,8 +54,7 @@ internal class InboundZiplineReference<T : Any>(
   }
 
   override fun get(outboundBridge: OutboundBridge<T>): T {
-    // TODO(jwilson): change generated subtypes to populate `T` on the super call.
-    throw IllegalStateException("get() on inbound handlers not yet implemented")
```
swankjesse, line 58: Yay

JakeWharton (approved):

PR #308, opened 2021-09-24
Don't crash publishing API docs

swankjesse:


swankjesse:
Should fix this:
```
* What went wrong:
Execution failed for task ':zipline:publishKotlinMultiplatformPublicationToMavenCentralRepository'.
> Failed to publish publication 'kotlinMultiplatform' to repository 'mavenCentral'
   > Invalid publication 'kotlinMultiplatform': multiple artifacts with the identical extension and classifier ('jar', 'null').
```

PR #310, opened 2021-09-30
Support FlowReferences without passing through Zipline

swankjesse:
This is useful for testing.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
@@ -15,6 +15,7 @@
  */
 package app.cash.zipline
 
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
 import kotlinx.coroutines.flow.channelFlow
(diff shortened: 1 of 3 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
```diff
@@ -30,9 +31,9 @@ class FlowReference<T> @PublishedApi internal constructor(
   @Contextual private val referenceFlowReference: ZiplineReference<ReferenceFlow>,
   @Contextual private val serializer: ZiplineSerializer<T>,
 ) {
-  private fun getJsonFlow(): Flow<String> {
+  @OptIn(ExperimentalCoroutinesApi::class)
```
JakeWharton, line 34: Hmm what is this for? We ideally should not have any of these in library code since it means we're subject to potential binary incompatibility in the future.

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/FlowReference.kt
swankjesse: IntelliJ is giving me warnings for `ProducerScope` which comes from using `channelFlow`.

Screen Shot 2021-09-30 at 15 49 27

JakeWharton: We'll have to get rid of `channelFlow { }` then and build it ourselves with a `flow { }` and a channel.

PR #315, opened 2021-10-07
Begin tearing down QuickJS-support for bridging interfaces

swankjesse:
This is still general-purpose in native code. I plan to drop that in
a follow-up PR.


zipline/src/jniTest/kotlin/app/cash/zipline/testUtil.kt
```diff
@@ -36,3 +37,11 @@ fun Zipline.loadTestingJs() {
     .use(BufferedReader::readText)
   quickJs.evaluate(testingJs, "testing.js")
 }
+
+operator fun <T : Any> QuickJs.set(name: String, type: KClass<T>, instance: T) {
```
swankjesse, line 41: I’ve got to triage the tests. Some can be deleted, and some should be replaced with Zipline-layer tests. Once I’m done that triage I’ll delete these two functions.

zipline/src/jniTest/kotlin/app/cash/zipline/testUtil.kt
```diff
@@ -36,3 +39,17 @@ fun Zipline.loadTestingJs() {
     .use(BufferedReader::readText)
   quickJs.evaluate(testingJs, "testing.js")
 }
+
+fun Throwable.stackTraceToString(): String {
+  val result = StringWriter()
+  printStackTrace(PrintWriter(result))
+  return result.toString()
+}
```
JakeWharton: Relation to https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/stack-trace-to-string.html?

zipline/src/jniTest/kotlin/app/cash/zipline/QuickJsCompileTest.kt
```diff
@@ -92,6 +93,7 @@ class QuickJsCompileTest {
     val value: String?
   }
 
+  @Ignore
```
JakeWharton, line 96: What is the plan for these two tests? Replace with Zipline-equivalent? Or delete?

JakeWharton (approved):

zipline/src/jniTest/kotlin/app/cash/zipline/testUtil.kt
swankjesse: I should just use that thing.

zipline/src/jniTest/kotlin/app/cash/zipline/QuickJsCompileTest.kt
swankjesse: Check if they're covering anything not already covered. Only if they're doing something original do they survive.

PR #321, opened 2021-10-13
Change JNI to make CallChannel special

swankjesse:
Previously the JNI bindings supported arbitrary Java interfaces. With this
change only CallChannel is supported.


zipline/native/InboundCallChannel.cpp
```diff
@@ -0,0 +1,111 @@
...
+  JSValue thisPointer = JS_GetPropertyStr(context->jsContext, global, name.c_str());
+  auto property = JS_NewAtom(context->jsContext, "serviceNamesArray");
+
+  JSValue jsResult = JS_Invoke(context->jsContext, thisPointer, property, 0, NULL);
+  jobject javaResult = context->toJavaObject(env, jsResult, true);
+
+  JS_FreeAtom(context->jsContext, property);
+  JS_FreeValue(context->jsContext, jsResult);
+  JS_FreeValue(context->jsContext, thisPointer);
+  JS_FreeValue(context->jsContext, global);
```
JakeWharton: Is calling `context->jsContext` 20 times free? Like is it just accessing a value in a struct on the stack or something? I just want to make sure it's not a pointer chase...

zipline/native/OutboundCallChannel.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "OutboundCallChannel.h"
+#include "Context.h"
+#include "ExceptionThrowers.h"
+
+OutboundCallChannel::OutboundCallChannel(Context* c, JNIEnv* env, const char* name, jobject object,
+                                 JSValueConst proxy)
+    : context(c),
+      name(name),
+      javaThis(env->NewGlobalRef(object)),
+      callChannelClass(static_cast<jclass>(env->NewGlobalRef(env->FindClass("app/cash/zipline/internal/bridge/CallChannel")))),
```
JakeWharton, line 26: Add this type to ProGuard rules with a full `-keep`

zipline/native/InboundCallChannel.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  return static_cast<jobjectArray>(javaResult);
+}
+
+jobjectArray InboundCallChannel::invoke(Context *context, JNIEnv* env, jstring instanceName,
+                                   jstring funName, jobjectArray encodedArguments) const {
+  JSValue global = JS_GetGlobalObject(context->jsContext);
+  JSValue thisPointer = JS_GetPropertyStr(context->jsContext, global, name.c_str());
+  auto property = JS_NewAtom(context->jsContext, "invoke");
```
JakeWharton: Are these functions all `@JsName`'d to ensure they're stable?

zipline/native/OutboundCallChannel.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+JSValue
+OutboundCallChannel::disconnect(JSContext* ctx, JSValueConst this_val, int argc, JSValueConst* argv) {
+  auto context = reinterpret_cast<const Context*>(JS_GetRuntimeOpaque(JS_GetRuntime(ctx)));
+  if (!context) {
+    return JS_ThrowReferenceError(ctx, "Null Java Proxy");
+  }
+  auto channel = reinterpret_cast<const OutboundCallChannel*>(JS_GetOpaque(this_val, context->jsClassId));
+  if (!channel) {
+    return JS_ThrowReferenceError(ctx, "Null Java Proxy");
```
JakeWharton: Is this copy/pasta? I don't expect this to ever happen, but the message seems like maybe it was from the old code.

JakeWharton (approved):

zipline/native/InboundCallChannel.cpp
swankjesse: Fixed. I extracted a local. I suspect the difference is negligible but it’s there enough that it’s more readable the way.

zipline/native/OutboundCallChannel.cpp
swankjesse: We don’t have a rules file yet so I’m punting to follow-up.
https://github.com/cashapp/zipline/issues/322

zipline/native/InboundCallChannel.cpp
swankjesse: Yes!

zipline/native/OutboundCallChannel.cpp
swankjesse: Yeah, I didn’t attempt to make behavior-changes in this PR, and that means the word ‘proxy’ still exists too much. Fixed.

zipline/native/OutboundCallChannel.cpp
JakeWharton: It's in `jniMain/resources/META-INF/proguard/`

PR #325, opened 2021-10-15
Support converting an array of values in Kotlin/Native

JakeWharton:
This is required by Zipline.

### DIFF
--- a/zipline/native/Context.cpp
+++ b/zipline/native/Context.cpp
@@ -369,7 +369,10 @@ Context::toJavaObject(JNIEnv* env, const JSValueConst& value, bool throwOnUnsupp
 
     case JS_TAG_OBJECT:
       if (JS_IsArray(jsContext, value)) {
-        const auto arrayLength = JS_VALUE_GET_INT(JS_GetPropertyStr(jsContext, value, "length"));
+        auto arrayLengthProperty = JS_GetPropertyStr(jsContext, value, "length");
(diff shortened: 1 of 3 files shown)


zipline/native/Context.cpp
```diff
@@ -393,7 +393,10 @@ Context::toJavaObject(JNIEnv* env, const JSValueConst& value, bool throwOnUnsupp
 
     case JS_TAG_OBJECT:
       if (JS_IsArray(jsContext, value)) {
-        const auto arrayLength = JS_VALUE_GET_INT(JS_GetPropertyStr(jsContext, value, "length"));
+        auto arrayLengthProperty = JS_GetPropertyStr(jsContext, value, "length");
+        const auto arrayLength = JS_VALUE_GET_INT(arrayLengthProperty);
+        JS_FreeValue(jsContext, arrayLengthProperty);
```
JakeWharton, line 374: Fixing a leak

JakeWharton: Maybe? I'm not clear on which JsValue's need freed

JakeWharton: We do it elsewhere for a call to `.length` on a JS array so I'm matching that behavior.

swankjesse: Yeah this looks right

swankjesse: If we get good at calling close() after all of our tests these leaks will jump out at us

swankjesse (approved):

PR #326, opened 2021-10-15
Enforce list item assertion order

JakeWharton:


zipline/src/jniTest/kotlin/app/cash/zipline/QuickJsInboundChannelTest.kt
```diff
@@ -161,7 +161,7 @@ class QuickJsInboundChannelTest {
     assertThat(result).asList().containsExactly(
       "disconnect",
       "service one"
-    )
+    ).inOrder()
```
swankjesse, line 164: whoa, is this different between Truth and AssertJ? I’ve been writing shitty tests

swankjesse (approved):

zipline/src/jniTest/kotlin/app/cash/zipline/QuickJsInboundChannelTest.kt
JakeWharton: Haven't used AssertJ in 6 or 7 years now so no idea. This will all change to kotlin.test shortly.

PR #329, opened 2021-10-19
Return String[] and not Object[] on CallChannel.invoke

swankjesse:
This was incorrect. The JVM didn't mind but Android crashes.


zipline/native/InboundCallChannel.cpp
```diff
@@ -33,14 +33,23 @@ jobjectArray InboundCallChannel::serviceNamesArray(Context *context, JNIEnv* env
   auto property = JS_NewAtom(jsContext, "serviceNamesArray");
 
   JSValue jsResult = JS_Invoke(jsContext, thisPointer, property, 0, NULL);
-  jobject javaResult = context->toJavaObject(env, jsResult, true);
```
swankjesse, line 36: We were using toJavaObject() because it has the nice side-effect of doing an exception check. But it didn’t know what the target type was, whereas the callsite does.

JakeWharton (approved):
I'll update the Kotlin/Native PR with this, too.

PR #330, opened 2021-10-19
Enable useArrayPolymorphism in kotlinx.serialization

swankjesse:
This isn't ideal but it fixes a crash and unblocks higher-leverage work.


JakeWharton (approved):
I thought this was the default. Is there a YouTrack for the crash?

swankjesse:
This is the closest issue I found:
https://github.com/Kotlin/kotlinx.serialization/issues/1252

PR #332, opened 2021-10-20
Add native utilities for structs in arrays

JakeWharton:
These dramatically simplify putting structs into struct arrays through operator-set and a varargs factory function.

cc @kpgalligan

### DIFF
--- a/zipline/src/nativeMain/kotlin/app/cash/zipline/nativeUtils.kt
+++ b/zipline/src/nativeMain/kotlin/app/cash/zipline/nativeUtils.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


zipline/src/nativeMain/kotlin/app/cash/zipline/nativeUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlinx.cinterop.CPointer
+import kotlinx.cinterop.CValues
+import kotlinx.cinterop.CVariable
+import kotlinx.cinterop.NativePlacement
+import kotlinx.cinterop.allocArray
+import kotlinx.cinterop.interpretCPointer
+import kotlinx.cinterop.sizeOf
+
+/** Copy the data of [item] to the [index] of [this] as if it were an array of [T] structs. */
+internal inline operator fun <reified T : CVariable> CPointer<T>.set(index: Int, item: CValues<T>) {
```
JakeWharton, line 28: Note that these types are wider than necessary. Normally you'll have a `CArrayPointer<T>` as the receiver (which is a typealias to `CPointer<T>`) and the item will be a `CValue<T>` (which inherits from `CValues<T>`). I can narrow them if desired, but nothing strictly prevents it since we're down in pointer math and memcpy here.

swankjesse (approved):

PR #339, opened 2021-10-22
Fix a bug that was crashing release builds

swankjesse:
I was doing useful work inside an assert() statement, and in release
builds that assert statement was optimized out.

Also convert QuickJS API uses to use JSAtom where possible. This is
represented as a single integer and should speed up some common operations.

### DIFF
--- a/zipline/native/Context.cpp
+++ b/zipline/native/Context.cpp
@@ -79,6 +79,11 @@ Context::Context(JNIEnv* env)
       jsRuntime(JS_NewRuntime()),
       jsContext(JS_NewContext(jsRuntime)),
       jsClassId(0),
+      lengthAtom(JS_NewAtom(jsContext, "length")),
+      serviceNamesArrayAtom(JS_NewAtom(jsContext, "serviceNamesArray")),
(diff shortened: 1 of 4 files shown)


zipline/native/Context.cpp
```diff
@@ -545,22 +555,24 @@ jstring Context::toJavaString(JNIEnv* env, const JSValueConst& value) const {
...
 
 jobjectArray Context::toJavaStringArray(JNIEnv* env, const JSValueConst& value) const {
-  assert(JS_IsArray(this->jsContext, value));
+  auto jsContext = this->jsContext;
+  assert(JS_IsArray(jsContext, value));
 
+  auto jsLength = JS_GetProperty(jsContext, value, this->lengthAtom);
   int length = 0;
-  auto jsLength = JS_GetPropertyStr(this->jsContext, value, "length");
-  assert(!JS_ToInt32(this->jsContext, &length, jsLength));
```
swankjesse, line 552: this call was not happening, and that was bad news

JakeWharton (approved):

PR #341, opened 2021-10-22
Do not use contextual serialization for ZiplineReference

JakeWharton:
This causes the serialization library to attempt to resolve a serializer for the type parameter which is almost certainly not a serializable type. This magically worked on the JVM and JS but eagerly fails on Native and thus required a different approach. Instead, we directly resolve the serializer for this type which is now held on the Endpoint and completely ignores the generic type parameter.

Also move EndpointTest and ZiplineReferenceTest into engineTest/ to get them running on native.


zipline-kotlin-plugin/hosted/src/main/kotlin/app/cash/zipline/kotlin/BridgedInterface.kt
```diff
@@ -100,32 +100,56 @@ internal class BridgedInterface(
...
-        ).apply {
-          putTypeArgument(0, type)
-          extensionReceiver = irCall(
-            callee = serializersModuleProperty.owner.getter!!,
+      if (type.classFqName == ziplineApis.ziplineReferenceFqName) {
+        // val serializer_0: KSerializer<EchoRequest> = context.endpoint.ziplineReferenceSerializer
+        //
+        // The generic type parameter of a ZiplineReference is likely not serializable. Therefore,
+        // look its serializer up directly rather than using the serialization module which would
+        // otherwise force lookup of the generic type parameter's serializer as well.
```
swankjesse, line 115: Nice

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -47,6 +43,10 @@ class Endpoint internal constructor(
   val clientNames: Set<String>
     get() = outboundChannel.serviceNamesArray().toSet()
 
+  @PublishedApi
+  internal val ziplineReferenceSerializer: KSerializer<ZiplineReference<*>> =
```
swankjesse, line 47: Nice

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -59,9 +59,11 @@ class Endpoint internal constructor(
 
   private fun computeSerializersModule(): SerializersModule {
     return SerializersModule {
+      // TODO eliminate the need to have this, possibly by special-casing FlowReference in the
+      //  compiler plugin the same way we are currently special-casing ZiplineReference.
```
swankjesse, line 63: I think I prefer having this, because it makes FlowReference not special?

swankjesse (approved):
Nice.

swankjesse:
Nice nice nice.

Is it worth the effort of making or finding a tracking bug for the difference in behavior in kotlinx.serialization here?

JakeWharton:
Yes I plan to ask about it and another problem with interfaces this week.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
JakeWharton: https://github.com/cashapp/zipline/issues/345

PR #343, opened 2021-10-22
Avoid intermediate operator for JSON encoding/decoding

JakeWharton:


swankjesse (approved):
Is this more efficient?

JakeWharton:
Fewer object allocations. Unlikely to have any real performance impact.

PR #351, opened 2021-11-01
Capture JavaScript stack traces

swankjesse:


JakeWharton:
Can you rebase on #350?

JakeWharton (approved):

swankjesse:
Done. Might need to do some more adjustment to handle polymorphic throwables. In follow-up!

PR #355, opened 2021-11-04
Rename the artifact for the native kotlin plugin

swankjesse:
Before: 'app.cash.zipline:hosted'
After:  'app.cash.zipline:zipline-kotlin-plugin-hosted'


JakeWharton:
The name is mutable. You should just be able to do `name = "whatever you want"` in the `build.gradle` of that project.

PR #356, opened 2021-11-05
Rename the artifact for the native kotlin plugin

swankjesse:
Before: 'app.cash.zipline:hosted'
After: 'app.cash.zipline:zipline-kotlin-plugin-hosted'

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -5,6 +5,10 @@ include(":zipline:testing")
 include(":zipline-gradle-plugin")
 include(":zipline-kotlin-plugin")
 include(":zipline-kotlin-plugin:hosted")
+project(":zipline-kotlin-plugin:hosted")
+  .apply {
+    name = "zipline-kotlin-plugin-hosted"
+  }
 include(":zipline-kotlin-plugin:tests")
 include(":zipline-tools")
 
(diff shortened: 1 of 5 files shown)


JakeWharton (approved):

swankjesse:
@JakeWharton please take another look. I don’t love it; I think it might be nicer to either change the name only for artifact publishing, or to move the `hosted` subdirectory to be top-level so I don’t need repetitive names when referring to it internally.

JakeWharton:
Odd that you cannot change the name without also affecting the path. Perhaps time to switch to a properties file? https://github.com/cashapp/treehouse/blob/f1e8e409508bff5bf7b65dfe6fc2ffa541e43a68/treehouse/compose/compiler-hosted/gradle.properties#L1. It's how we divorce the project name from publish name everywhere else.

swankjesse:
Yeah, that’s changing the artifact name and not the project name.

swankjesse:
Alright, 3rd time’s a charm.

swankjesse:
I dislike `gradle.properties` because it’s usually stringly-typed boilerplate. I prefer DRY gradle stuff and getting tools to yell at me if I mistype something.

PR #361, opened 2021-11-10
Modular JS builds

adrw:
Split compiled JS output into respective files and load separately into QuickJS to allow future developments such as OTA updates of much smaller JS files for changed app code but rely on Kotlin STD lib and other library code which changes less frequently that's shipped in app releases.

### DIFF
--- a/zipline/src/engineMain/kotlin/app/cash/zipline/internal/defineJs.kt
+++ b/zipline/src/engineMain/kotlin/app/cash/zipline/internal/defineJs.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.internal
+
+internal const val currentModuleId = "app_cash_zipline_currentModuleId"
+
+/**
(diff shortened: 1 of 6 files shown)


zipline/src/engineMain/kotlin/app/cash/zipline/internal/defineJs.kt
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.internal
+
+internal const val currentModuleId = "app_cash_zipline_currentModuleId"
+
+/**
+ * Implement a AMD module loader for Kotlin/JS AMD modules running on QuickJS.
+ * https://github.com/amdjs/amdjs-api/blob/master/AMD.md
```
JakeWharton, line 22: Why was AMD chosen? It was popular a long time ago, but isn't widespread anymore given that modules were standardized as part of the language. I would have expected us to implement that, and it seems like QuickJS already has some kind of support for it as well: https://bellard.org/quickjs/quickjs.html#Modules.

JakeWharton: Ah, Kotlin can't produce them yet. Still surprised we do AMD and not CommonJS given we're much closer to Node doing synchronous loading than a Browser needing to do async.

swankjesse: I don't think we want to load files directly... that won't work for QuickJS bytecode. But AMD vs. CommonJS? I have no opinions

swankjesse: (there's nothing async at the moment; this is just a very easy integration to hook up)

JakeWharton: Yeah I guess it doesn't really matter since Kotlin spits out UMD which is AMD+CommonJS(+plus vanilla) we even have the potential of changing what we implement. If Kotlin ever supports ES modules we should do those, but for now AMD 👍 .

swankjesse (approved):

adrw (commented):
LG

PR #362, opened 2021-11-10
Add iOS arm32 as a native target

AlecKazakova:


JakeWharton:
We intentionally are not shipping any 32-bit targets. What do you need it for?

JakeWharton:
This requires _massive_ code changes to support because of the difference in cinterop code gen across bit depths.

AlecKazakova:
makes sense

PR #363, opened 2021-11-15
Zipline-bytecode: read and write encoded QuickJS objects.

swankjesse:


zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/AtomSet.kt
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Maps each commonly-used string to an integer index for performance. Strings built into QuickJS
+ * are assigned an index statically; user-provided strings get an index dynamically.
+ *
+ * When encoding an object, built-in strings are not encoded.
+ */
+class AtomSet(
+  val strings: List<String>
+) {
+  private val stringToIndex = mutableMapOf<String, Int>()
```
Egorand, line 27: nit: use `buildMap`?

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectReader.kt
```diff
@@ -0,0 +1,177 @@
...
+
+  fun readJsObject(): JsObject {
+    check(!::atoms.isInitialized)
+    atoms = readAtoms()
+    return readObjectRecursive()
+  }
+
+  private fun readAtoms(): AtomSet {
+    val version = source.readByte().toInt()
+    if (version != version) {
```
Egorand: Should be `BC_VERSION`?
```suggestion
    if (version != BC_VERSION) {
```

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectWriter.kt
```diff
@@ -0,0 +1,156 @@
...
+ */
+class JsObjectWriter(
+  private val atoms: AtomSet,
+  private val sink: BufferedSink
+) : Closeable by sink {
+  private var used: Boolean = false
+
+  fun writeJsObject(value: JsObject) {
+    check(!used)
+    used = true
```
Egorand, line 36: Would it make sense to just change `writeJsObject` and `readJsObject` into top-level functions instead of class methods? That way you don't need to defend against `JsObjectWriter` being used twice.

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/lineNumbers.kt
```diff
@@ -0,0 +1,87 @@
...
+import okio.BufferedSink
+import okio.BufferedSource
+import okio.Closeable
+
+/** Decode the pc2line table. */
+class LineNumberReader(
+  functionLineNumber: Int,
+  private val source: BufferedSource,
+) : Closeable by source {
+  var pc: Int = 0
```
Egorand, line 27: `programCode` a better name?

Egorand, line 27: Also, why are these public, should be private?

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/types.kt
```diff
@@ -0,0 +1,90 @@
...
+import okio.Buffer
+import okio.ByteString
+
+sealed class JsObject
+
+object JsNull : JsObject()
+
+object JsUndefined : JsObject()
+
+data class JsBoolean(val value: Boolean) : JsObject()
```
Egorand, line 26: Can change these into `value` classes?

zipline-bytecode/src/test/kotlin/app/cash/zipline/bytecode/EncodingTest.kt
```diff
@@ -0,0 +1,107 @@
...
+ */
+package app.cash.zipline.bytecode
+
+import com.google.common.truth.Truth.assertThat
+import okio.Buffer
+import org.junit.Test
+
+class EncodingTest {
+  @Test fun intGetBit() {
+    assertThat(0b00000001.bit(0)).isFalse()
```
Egorand, line 24: A bit confusing that `bit`'s parameter is zero-based, should it be one-based, since you need "the first bit"?

Egorand (approved):
NICE!

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/AtomSet.kt
swankjesse: It’s still experimental!

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectReader.kt
swankjesse: oooh yes definitely that. Fixed.

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectWriter.kt
swankjesse: Yeah I think that’s worth a shot. Let’s pair on that?

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/lineNumbers.kt
swankjesse: I was hoping to avoid allocating a `Pair<Int, Int>` and/or a 2-element `IntArray`. If you’d like to try stuff here too, I’m game.

swankjesse: `pc` is the name from QuickJS, I think it’s a term of art for VMs.
https://en.wikipedia.org/wiki/Program_counter

I was tempted to change it to something else (instruction?) but I’m reluctant to diverge from QuickJS terms.

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/types.kt
swankjesse: Ooooh I hadn’t even considered value classes. That won’t work due to the sealed class.

zipline-bytecode/src/test/kotlin/app/cash/zipline/bytecode/EncodingTest.kt
swankjesse: No? The first element in a list is also 0-based, right?

JakeWharton: It behaves like an array. The first element is the 0th.

zipline-bytecode/src/test/kotlin/app/cash/zipline/bytecode/LineNumbersTest.kt
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package app.cash.zipline.bytecode
+
+import com.google.common.truth.Truth.assertThat
+import okio.Buffer
+import okio.ByteString.Companion.decodeHex
+import org.junit.Test
+
+class LineNumbersTest {
```
JakeWharton, line 23: It's kind of a change detector test, but I would feel more comfortable with a test which used a real `QuickJs` to produce a table, round-trip it through reader/writer, and validate it produces the same `ByteString`.

JakeWharton (approved):

zipline-bytecode/src/test/kotlin/app/cash/zipline/bytecode/LineNumbersTest.kt
swankjesse: YES! We have that in JsObjectEncodingTest, which has a function, `assertRoundTrip()`.

swankjesse: Oh, but not for line numbers. Yeah, good call. Lemme add that too.

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/AtomSet.kt
Egorand: Ah true

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectWriter.kt
Egorand: Sure!

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/lineNumbers.kt
Egorand: Good point

zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/types.kt
Egorand: Ah crap, okay

PR #364, opened 2021-11-15
Add SourceMapParser

Egorand:


zipline/src/engineMain/kotlin/app/cash/zipline/internal/SourceMapParser.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    val buffer = Buffer()
+
+    var sourceIndex = 0
+    var sourceLine = 0L
+    var sourceColumn = 0L
+    var nameIndex = 0
+
+    val groups = mutableListOf<Group>()
+    for (group in sourceMap.mappings.split(";")) {
```
Egorand: Do we care about performance here, should we optimize these splits?

JakeWharton: Not until it's a problem

zipline/src/engineMain/kotlin/app/cash/zipline/internal/SourceMapParser.kt
```diff
@@ -0,0 +1,179 @@
...
+import okio.Buffer
+import okio.BufferedSource
+import okio.IOException
+
+/**
+ * Parses the contents of a source map file to enable mapping from the elements in the transformed
+ * (e.g. minified) source to the elements in the original source. This is an implementation of the
+ * [Source Map Revision 3 Proposal](https://sourcemaps.info/spec.html).
+ */
+internal class SourceMapParser {
```
JakeWharton: Seems like we can drop the class, make all the data classes top-level and internal, and make `parse` exist directly on its `companion object`. This class is not stateful and does nothing except perform namespace scoping.

zipline/src/engineTest/kotlin/app/cash/zipline/SourceMapParserTest.kt
```diff
@@ -0,0 +1,125 @@
...
+        "version": 3,
+        "file": "kotlin-js-demo.js",
+        "sources": [
+          "../../../../../src/main/kotlin/main.kt"
+        ],
+        "sourcesContent": [
+          null
+        ],
+        "names": [],
+        "mappings": ";;;;;;;;;;;;EAAA,gB;IACE,YAAY,c;IACZ,OAAQ,KAAI,KAAJ,C;EACV,C;;;;;;"
```
JakeWharton: Does this contain a segment with an optional name? I'm missing a few golden `Segment` instances here that are asserted again to ensure the logic is fully covered.

zipline/src/engineTest/kotlin/app/cash/zipline/SourceMapParserTest.kt
```diff
@@ -0,0 +1,125 @@
...
+  @Test fun readVarints() {
+    val buffer = Buffer().writeUtf8("IAkpZI")
+    assertEquals(4L, buffer.readVarint())
+    assertEquals(0L, buffer.readVarint())
+    assertEquals(12946L, buffer.readVarint())
+    assertEquals(4L, buffer.readVarint())
+    assertTrue(buffer.exhausted())
+  }
+
+  @Test fun mapStacktrace() {
```
JakeWharton: This doesn't seem like it should be / was meant to be committed.

JakeWharton (approved):

zipline/src/engineTest/kotlin/app/cash/zipline/SourceMapParserTest.kt
Egorand: It'd be useful to have this as an end-to-end test with more complex input than `simpleSourceMap`, but looks like we'd need to copy over a bunch of map files as they're not available on all platforms. Or do you think `simpleSourceMap` with a few more cases, such as optional name, is enough?

JakeWharton: I agree a more comprehensive test is needed but this reads arbitrary data from the filesystem without appropriate task dependencies. The source files may not be there, or may be out-of-date.

Aside from that, there's a lot of logic that's going on here tangential to the API under test. There's entirely new input types with a mapping and rendering function. The API under test is hidden under all this.

I would expect an end-to-end test to be more focused on our eventual usage which is remapping bytecode.

zipline/src/engineMain/kotlin/app/cash/zipline/internal/SourceMapParser.kt
Egorand: Agree, done: b1120ab

zipline/src/engineTest/kotlin/app/cash/zipline/SourceMapParserTest.kt
Egorand: Sounds good, deleted 25d9fab

zipline/src/engineTest/kotlin/app/cash/zipline/SourceMapParserTest.kt
Egorand: Looks like kotlin-js doesn't populate `names` when generating the map, so we shouldn't be expecting to see any names in real-world code. Would probably still be a good idea to run this code against a map generated by something else, I can look into that in a followup. Meanwhile made the test a bit more descriptive and added a case for a line that doesn't map to anything in the source: 903d35a

swankjesse (approved):

PR #369, opened 2021-11-18
Tests to confirm interface serializers don't fail fast

swankjesse:
This is disappointing.


swankjesse:
https://github.com/Kotlin/kotlinx.serialization/issues/1764

JakeWharton (approved):

PR #371, opened 2021-11-19
Support exceptions in Console.log

swankjesse:
Previously we were calling toString() on them and losing stack traces.


JakeWharton (approved):
CI failure is probably the need to update Android loger

PR #378, opened 2021-11-23
Unbreak imports of extension functions

swankjesse:
Excluding these breaks builds in downstream projects.
https://stackoverflow.com/questions/61450260/extension-function-cant-be-imported-from-library

### DIFF
--- a/zipline/build.gradle.kts
+++ b/zipline/build.gradle.kts
@@ -193,9 +193,6 @@ android {
     }
 
     packagingOptions {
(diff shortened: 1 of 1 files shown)


swankjesse:
Screen Shot 2021-11-23 at 17 27 48

JakeWharton:
Interesting. I didn't know this applied to the contents of the aar. I wonder if we can somehow make it apply only to our test apk.

JakeWharton (approved):
To unblock

swankjesse:
Yeah, there’s probably a better way to get the test coverage we want for this.

PR #382, opened 2021-11-30
Clean up continuations on Zipline.close()

swankjesse:
This fixes two failures:

 * When resuming into JS code, don't crash with a 'Zipline closed' error
 * When waiting on JS code, don't wait forever


adrw (approved):
LG

PR #389, opened 2021-12-13
zipline-library reorg & Upstream KotlinJsCompile configuration to Zipline Plugin

adrw:
If this works, we won't need to manually set this on every project that uses Zipline plugin https://github.com/cashapp/zipline/blob/d5e3f7b6b2554a93e3dc47799610aae6c9dc42a4/zipline/testing/build.gradle.kts#L68-L75

Thanks to [#gradle Slack channel for help](https://cash.slack.com/archives/C6GLXUH2L/p1639430552014700)


zipline-gradle-plugin/build.gradle.kts
```diff
@@ -44,6 +45,14 @@ gradlePlugin {
   }
 }
 
+pluginManager.withPlugin("org.jetbrains.kotlin.js") {
+  tasks.withType(KotlinJsCompile::class.java).configureEach {
+    kotlinOptions {
+      freeCompilerArgs += listOf("-Xir-per-module")
```
acostama: what does that do 🤔

zipline/testing/build.gradle.kts
```diff
@@ -15,7 +15,6 @@ kotlin {
     browser()
     // TODO upstream this to ZiplinePlugin
     binaries.library()
-    binaries.executable()
```
acostama, line 18: less things todo? lgtm

acostama (approved):

zipline-gradle-plugin/build.gradle.kts
```diff
@@ -44,6 +45,14 @@ gradlePlugin {
   }
 }
 
+pluginManager.withPlugin("org.jetbrains.kotlin.js") {
```
JakeWharton: This has no effect here. Did you mean to put it in the plugin implementation?

adrw: Yeah, somebody in #gradle Slack had suggested that block though it looked like gradle.kts and not .kt like would be in the plugin implementation. Would you approach it by adding this configuring to the `override fun apply(project: Project)` so you can then get the tasks through the project?

zipline-gradle-plugin/build.gradle.kts
adrw: Turns on modular JS builds that emit a IR (Kotlin compiler intermediate representation) JS file per JS module instead of all of the JS in a single file.

zipline-gradle-plugin/build.gradle.kts
JakeWharton: Yep. Should copy/paste right in.

adrw: Works! Thanks!

adrw:
Used repo pattern from Wire to get access to the project version of Zipline Gradle Plugin in the samples to test out that KotlinJS configuration worked as expected.

adrw:
Not sure how to resolve the remaining failures, project builds and tests correctly locally

adrw:
Going to land a smaller diff first separate from the repo reorg since I think it warrants a bit more feedback on repo structure before I merge https://github.com/cashapp/zipline/pull/390

zipline-library/zipline/testing/src/commonMain/kotlin/app/cash/zipline/testing/adapters.kt
```diff
@@ -39,7 +39,7 @@ data class AdaptersResponse(
 
 internal object AdaptersRequestSerializer : KSerializer<AdaptersRequest> {
   override val descriptor = PrimitiveSerialDescriptor(
-    "app.cash.zipline.testing.AdaptersRequestSerializer",
```
swankjesse, line 42: why?

zipline-library/zipline/src/jniTest/kotlin/app/cash/zipline/QuickJsCompileJvmTest.kt
```diff
@@ -48,7 +48,7 @@ class QuickJsCompileJvmTest {
     assertEquals("JavaScript.f2(C:\\Documents\\myFile.js:8)", t.stackTrace[0].toString())
     assertEquals("JavaScript.f1(C:\\Documents\\myFile.js:4)", t.stackTrace[1].toString())
     assertEquals("JavaScript.<eval>(C:\\Documents\\myFile.js:1)", t.stackTrace[2].toString())
-    assertEquals("app.cash.zipline.QuickJs.execute(Native Method)", t.stackTrace[3].toString())
```
swankjesse, line 51: why?

swankjesse:
Wanna split this into two PRs, one that does the move, and a follow-up that makes plugin changes?

swankjesse:
Obsolete.

PR #394, opened 2021-12-29
WIP: ZiplineService

swankjesse:


zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineService.kt
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline
+
+/**
+ * Implemented by all interfaces that can safely be passed from the host platform to Kotlin/JS or
+ * vice versa.
```
oldergod, line 20: When is an interface able to safely do so?

oldergod: I see, the `ZiplineServiceAdapter` is the important part here.

Is it because it's an interface that you don't wanna do like Wire's `Message`s ? Something like

`ZipineService<A: ZiplineServiceAdapter>` ?

oldergod (approved):

swankjesse:
Obsoleted by https://github.com/cashapp/zipline/pull/395

Issue #383, opened 2021-11-30
Replace ZiplineReference annotation with ZiplineService interface


swankjesse:
Right now `ZiplineReference` exists so we can pass interface handles across the boundary and release them to terminate the link.

I’d like to replace this with a new type, `ZiplineService` with approximately this API:

```
interface ZiplineService {
  /**
   * Release any resources and remove the link to the
   * other runtime. It is an error to call any method on a
   * service after it has been closed.
   */
  fun close()
}
```

I can think of a couple Big Benefits here:

 * We will generate bigger stubs in each ZiplineService sub-interface, reducing the amount of generated code in `Zipline.get()` and `Zipline.set()`. Right now our incremental builds are broken because if you change an interface but not calls to `Zipline.get()`, those calls don’t rebuild. (This currently causes linking errors!)
 * We should be able to put `ZiplineService` instances inside other objects without the generics problems we had when doing that in Kotlin/Native with kotlinx.serialization. (restoring what was lost in https://github.com/cashapp/zipline/pull/341)

I think this might also be a bit more obvious. The biggest drawback of this approach is that it prevents Zipline from working with arbitrary interfaces like `Runnable`. I think this is a fine tradeoff.

swankjesse:
Last step: https://github.com/cashapp/zipline/pull/424

PR #395, opened 2021-12-31
ZiplineService

swankjesse:
This introduces ZiplineService (public API) and ZiplineServiceAdapter (internal)
that should soon replace other classes including InboundBridge, OutboundBridge,
and ZiplineReference.

The biggest benefit of this change is that it should make it possible to not
need the clumsy ZiplineReference box when passing interfaces across the bridge.

A side-effect of this change is that API-dependent code is now generated
alongside the interface and not the use-sites. This should fix our problems
with incremental builds. (Without this, any time an interface method is changed
but the calls to Zipline.get() or Zipline.set() don't change, a rebuild is
not triggered and things fail at runtime.)

https://github.com/cashapp/zipline/issues/383

In follow-up PRs I'd like to tear down the above features obsoleted by this
change. I'm holding off on doing that here to simplify the transition. As a
consequence, this PR introduces some duplicate code.

### DIFF
--- a/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/ir.kt
+++ b/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/ir.kt
@@ -15,12 +15,15 @@
  */
 package app.cash.zipline.kotlin
 
+import org.jetbrains.kotlin.backend.common.ScopeWithIr
 import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
 import org.jetbrains.kotlin.backend.common.ir.createDispatchReceiverParameter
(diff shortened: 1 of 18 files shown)


oldergod (approved):
:+1:

zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/ir.kt
```diff
@@ -285,3 +292,58 @@ fun irVal(
...
+    original.endOffset
+  )
+}
+
+/** This creates `companion object` if it doesn't exist already. */
+fun IrClass.getOrCreateCompanion(
+  irPluginContext: IrPluginContext,
+): IrClass {
+  val existing = declarations.firstOrNull {
+    it is IrClass && it.name.identifier == "Companion"
```
JakeWharton, line 313: What if it's a named companion?

Issue #402, opened 2022-01-12
Throw the right exceptions with useful information 


kevincianfarini:
It's hard to know what's going on in zipline when an API changes. We should throw the correct `NoSuchMethodError` if nothing can be resolved properly. 

```
NoSuchMethodError:
  invoked:
    ZiplineTreehouse.getScreen(String): ZiplineReference<ZiplineTreehouseUi>
  candidates:
    ZiplineTreehouse.getScreen(String, Navigator): ZiplineReference<ZiplineTreehouseUi>
    ZiplineTreehouse.doTheThing(String, Navigator): ZiplineReference<ZiplineTreehouseUi>
    ZiplineTreehouse.close(): ZiplineReference<ZiplineTreehouseUi>

```

swankjesse:
Currently our bridge function, `CallChannel.invoke()` returns a value (`v`), a null (`n`), or a throwable (`t`).

Let’s add a 4th option, `u` for _Unknown Method_ that includes as its argument a String containing the supported functions. Then the outbound bridge is responsible for converting that list into an appropriate exception.

swankjesse:
@veyndan all done on this?

veyndan:
Yep, not sure why https://github.com/cashapp/zipline/pull/409#issue-1120836000 didn't auto close this issue. Perhaps it's because it didn't get merged into trunk directly 🤷

PR #408, opened 2022-01-27
Report function signature on unexpected function

veyndan:
When faced with an unexpected function (e.g., via incompatible APIs) we currently just display the function name. This isn't super useful if the function name is the same but the parameter types or return type differs.

Through this change, we now display the whole function signature (except for the function location due to our use of duck typing), so instead of `echo` we'd have `fun echo(com.example.Request): com.example.Response`.

This is the first part of the solution to #402.

### DIFF
--- a/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/ir.kt
+++ b/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/ir.kt
@@ -87,28 +87,6 @@ internal val IrSimpleFunction.signature: String
     append("fun ${name.identifier}(${valueParameters.joinToString { (it.type as IrSimpleType).asString() }}): ${(returnType as IrSimpleType).asString()}")
   }
 
-/** Inspired by [org.jetbrains.kotlin.ir.backend.js.utils.asString]. */
-fun IrSimpleType.asString(): String =
-  classifier.asString() +
-    (arguments.ifNotEmpty {
-      joinToString(separator = ",", prefix = "<", postfix = ">") { it.asString() }
-    } ?: "") +
-    (if (hasQuestionMark) "?" else "")
-
-/** Copied from [org.jetbrains.kotlin.ir.backend.js.utils.asString]. */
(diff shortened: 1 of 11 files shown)


zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/ir.kt
```diff
@@ -56,19 +56,58 @@ import org.jetbrains.kotlin.ir.expressions.impl.IrDelegatingConstructorCallImpl
...
+    if (isSuspend) {
+      append("suspend ")
+    }
+    append("fun ${name.identifier}(${valueParameters.joinToString { (it.type as IrSimpleType).asString() }}): ${(returnType as IrSimpleType).asString()}")
+  }
+
+/** Inspired by [org.jetbrains.kotlin.ir.backend.js.utils.asString]. */
+fun IrSimpleType.asString(): String =
+  classifier.asString() +
+    (if (hasQuestionMark) "?" else "") +
```
swankjesse: `?` goes after `>`

zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/ir.kt
```diff
@@ -56,19 +56,58 @@ import org.jetbrains.kotlin.ir.expressions.impl.IrDelegatingConstructorCallImpl
...
+
+internal val IrSimpleFunction.signature: String
+  get() = buildString {
+    if (isSuspend) {
+      append("suspend ")
+    }
+    append("fun ${name.identifier}(${valueParameters.joinToString { (it.type as IrSimpleType).asString() }}): ${(returnType as IrSimpleType).asString()}")
+  }
+
+/** Inspired by [org.jetbrains.kotlin.ir.backend.js.utils.asString]. */
```
swankjesse: Can you take these two functions that derive code from Jetbrains and put them into a new file with a Jetbrains copyright up top? Maybe say, `typeToString.kt` ?

swankjesse (approved):

PR #411, opened 2022-02-02
Rename get/set to take/bind

veyndan:
You shouldn't be calling get or set multiple times.

### DIFF
--- a/samples/emoji-search/presenters/src/jsMain/kotlin/app/cash/zipline/samples/emojisearch/RealEmojiSearchPresenter.kt
+++ b/samples/emoji-search/presenters/src/jsMain/kotlin/app/cash/zipline/samples/emojisearch/RealEmojiSearchPresenter.kt
@@ -32,7 +32,7 @@ class RealEmojiSearchPresenter(
         loadImageIndex()
         send(produceModel())
 
-        val events = eventsReference.get()
+        val events = eventsReference.take()
         events.collectLatest { event ->
           when (event) {
             is EmojiSearchEvent.SearchTermEvent -> {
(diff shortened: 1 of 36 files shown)


oldergod:
Can you call `put` 3 times and then do 3 `take` ?

veyndan:
I'm assuming you mean with the same service name, so something like this would pass:

```kotlin
@Test fun multiPutAndTake(): Unit = runBlocking(dispatcher) {
  zipline.put<EchoService>("helloService", JvmEchoService("hi"))
  zipline.put<EchoService>("helloService", JvmEchoService("yo"))
  zipline.take<EchoService>("helloService")
  zipline.take<EchoService>("helloService")
}
```

But the reason for the name change is that we don't want to encourage this. You should be calling `close` on the service when it's no longer in use to prevent a resource leak. Having multiple `take`s encourages bad practice, so it's better to just store a reference to the service in a variable somewhere and then call `close` at some point.

JakeWharton:
I told Jesse yesterday that put wasn't very strong here. What about register/consume?

veyndan:
Yeah that sounds good. If we go with `register` though, presumably the better counterpart would be `unregister`. Thoughts?

JakeWharton:
I'm okay with the terms being asymmetric because you never really unregister. Once you register we retain the claim on the name forever, even if you consume.

veyndan:
@JakeWharton Updated the naming of things.

JakeWharton:
@swankjesse Are you okay with this naming?

oldergod (approved):

JakeWharton (approved):

swankjesse (approved):
Thank you!

PR #412, opened 2022-02-07
Use the ZiplineLoader in the emoji sample

swankjesse:


JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompiler.kt
```diff
@@ -83,3 +83,13 @@ class ZiplineCompiler {
...
 }
+
+/**
+ * Compile .js files to .zipline files. This is intended for internal use only; callers should
+ * instead use the Gradle plugin.
+ */
+fun main(vararg args: String) {
+  val outputDir = File(args[1])
+  outputDir.mkdirs()
+  ZiplineCompiler().compile(inputDir = File(args[0]), outputDir = outputDir)
```
JakeWharton, line 94: Makes me wonder why this requires an instance

PR #419, opened 2022-02-10
Release on x86_64 and aarch64 architectures

swankjesse:


swankjesse:
I think I’ve done this wrong. We should build both architectures, but only build on x86_64.

JakeWharton (approved):

PR #422, opened 2022-02-14
Support serializing ZiplineServices

swankjesse:
Currently this only works if the service is a parameter or return
type of another ZiplineService.

I tried quite hard to make it work everywhere using `@Serializable`
but that required some pretty ugly compromises. In particular we'd
need to either register services explicitly, or lean on implementation
details of kotlinx.serialization and convert ZiplineService from an
interface to an abstract class.

### DIFF
--- a/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
+++ b/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
@@ -93,7 +93,7 @@ internal class AdapterGenerator(
   private fun getOrCreateAdapterClass(
     companion: IrClass
   ): IrClass {
-    // object Adapter : ZiplineServiceAdapter<SampleService>() {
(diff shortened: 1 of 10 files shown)


adrw (approved):
LG

adrw:
> convert ZiplineService from an interface to an abstract class

I'm not offended if we were to do an abstract class, would that have ended up being simpler in other aspects? What would be the downsides of an abstract class?

JakeWharton (approved):

swankjesse:
I did all the work to convert it to an abstract class and it doesn’t work out as well as I’d hoped.

In particular, not just ZiplineService needs to be an abstract class but all the user-defined services, like `EchoService` and such. This is slightly worse, but okay, because they get uglier syntactically:

```
abstract class EchoService : ZiplineService() {
  abstract fun echo(request: EchoRequest): EchoResponse
}
```

But the dealbreaker is we now create a way for our users to put state in those classes, and that breaks the principal of least surprise. In particular, you could do this:

```
abstract class EchoService : ZiplineService() {
  var prefix = "Hello, "
  
  abstract fun echo(request: EchoRequest): EchoResponse
}
```

That state definitely won’t be available in the stub, and I think that’s gonna be very surprising.

PR #427, opened 2022-02-17
Build Linux binaries in Docker

swankjesse:


zipline/src/jvmMain/CMakeLists.txt
```diff
@@ -5,7 +5,6 @@ set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_COMPILER "clang++")
```
swankjesse, line 8: Not using clang on Linux because I couldn’t figure out how to build with it on Linux without dragging in newer libstdc++.

I believe we should still get clang on macOS via some defaults? If not we could add it to `release.yml`.

swankjesse: Confirmed.
```
-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
```

.github/workflows/release.yaml
```diff
@@ -50,7 +54,9 @@ jobs:
           cmake --build build/jni/${{ matrix.arch }}/ --verbose
           cp -v build/jni/${{ matrix.arch }}/libquickjs.* zipline/src/jvmMain/resources/jni/${{ matrix.arch }}/
 
-      - run: ./gradlew build
+      - run: |
+          mkdir -p zipline/build/kotlin/sessions
```
swankjesse: Attempting to workaround https://youtrack.jetbrains.com/issue/KT-51374

.github/workflows/build.yaml
```diff
@@ -38,12 +38,20 @@ jobs:
...
-      # like "file not found" for `#include <string>`. Explicitly (re?)installing fixes things.
-      - run: sudo apt-get install clang
+      - name: Build native library (Linux)
         if: matrix.os == 'ubuntu-latest'
-
-      - name: Build native library
+        run: |
+          cp -a . ../zipline-dockerbuild
+          docker run \
+            --mount src=`pwd`/../zipline-dockerbuild,target=/zipline,type=bind \
```
JakeWharton: This syntax is awful. I'll use the old volume syntax until they tear it from my cold, dead hands.

.github/workflows/build.yaml
```diff
@@ -38,12 +38,20 @@ jobs:
...
+            amd64/centos:7 \
+            /zipline/.github/workflows/build-linux-amd64.sh
+          mkdir -p zipline/src/jvmMain/resources/jni/amd64
+          cp -v ../zipline-dockerbuild/build/jni/amd64/libquickjs.* zipline/src/jvmMain/resources/jni/amd64/
+          docker system prune -a -f
```
JakeWharton: Is this needed on CI? I would think a `--rm` on the `run` command (or really, doing nothing) would be fine.

JakeWharton (approved):

.github/workflows/build.yaml
swankjesse: Fixed.

.github/workflows/build.yaml
swankjesse: I was thrashing trying to figure out what was causing the Kotlin daemon to get stuck. I think it’s cause I’d had the project directory in a Docker mount when Kotlin was running, and that might have broken special files?

.github/workflows/build.yaml
JakeWharton: I mean you didn't have to change it. This is what they recommend now. It's just so yucky.

PR #429, opened 2022-02-22
Use ZiplineCache in ZiplineLoader

adrw:
This also checks in the kotlin-js yarn.lock file which seems to be automatically generated after the Kotlin 1.6 bump


zipline-loader/src/nativeMain/kotlin/app/cash/zipline/loader/cacheDatabaseNative.kt
```diff
@@ -0,0 +1,16 @@
...
+actual class DriverFactory {
+  actual fun createDriver(): SqlDriver {
+    return NativeSqliteDriver(
+      schema = Database.Schema,
+      name = "zipline-loader.db",
+    )
+  }
+}
+
+// TODO find the native exception class
```
JakeWharton, line 30: File an issue so we don't forget? Or are you planning on doing this before merge.

JakeWharton (approved):

zipline-loader/src/nativeMain/kotlin/app/cash/zipline/loader/cacheDatabaseNative.kt
adrw: Good call: https://github.com/cashapp/zipline/issues/433

PR #435, opened 2022-02-24
Use prebuilt Manifest in resources when offline

adrw:


JakeWharton (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -130,5 +139,9 @@ class ZiplineLoader(
   suspend fun localFileHashes(): List<ByteString> {
     TODO()
   }
+
+  companion object {
+    internal const val PREBUILT_MANIFEST_FILE_NAME = "prebuilt.manifest.json"
```
swankjesse: Rename to `manifest.zipline.json` ?

swankjesse (approved):

PR #437, opened 2022-03-03
Make the FlowReference wrapper unnecessary

swankjesse:
This drops support for Flows as fields in objects passed
through Zipline services. But flows can be fields in any
parameter or return type. This is the same limitation of
ZiplineService instances.

All flows passed to or returned from a ZiplineService is
exactly one-shot - not consuming or double-consuming are
both unsupported.


JakeWharton:
Nice. I got like 2/3rds of the way doing this a few months ago. Glad you took it across the line.

Will review tomorrow.

zipline/testing/src/jsMain/kotlin/app/cash/zipline/testing/sealedClassSerializersJs.kt
```diff
@@ -31,14 +29,12 @@ class JsSealedClassMessageService : SealedClassMessageService {
...
-  override fun colorSwapFlow(request: FlowReference<SealedMessage>): FlowReference<SealedMessage> {
-    val requestFlow = request.take()
-    val flow: Flow<SealedMessage> = flow {
-      requestFlow.collect { message ->
+  override fun colorSwapFlow(flow: Flow<SealedMessage>): Flow<SealedMessage> {
+    return flow {
+      flow.collect { message ->
         emit(colorSwap(message))
       }
     }
```
JakeWharton: This is just a `flow.map(::colorSwap)` now, if you want.

JakeWharton (approved):
Such a big ergonomic win

zipline/testing/src/jsMain/kotlin/app/cash/zipline/testing/sealedClassSerializersJs.kt
swankjesse: Ooooh yes please.

PR #443, opened 2022-03-09
Add to Gradle plugin Download task to populate app resources

adrw:
Zipline Loader has a parameter to identify a local resources directory in the app package where Zipline files are shipped with the app release. This prevents latency on first app open after install since it can rely on the shipped Zipline files and not have to do any network requests. To improve developer velocity, this new Gradle task makes it easy to populate a resources directory with the latest compiled Zipline files downloaded from the network.


zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineDownloadTask.kt
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package app.cash.zipline.gradle
+
+import java.io.File
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+
+open class ZiplineDownloadTask : DefaultTask() {
```
JakeWharton:
```suggestion
abstract class ZiplineDownloadTask : DefaultTask() {
```

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineDownloadTask.kt
```diff
@@ -0,0 +1,46 @@
...
+  var manifestUrl: String? = null
```
JakeWharton:
```suggestion
  @get:Input
  abstract val manifestUrl: Property<String>
```

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineDownloadTask.kt
```diff
@@ -0,0 +1,46 @@
...
+
+  @OutputDirectory
+  var downloadDir: File? = null
```
JakeWharton:
```suggestion
  @get:OutputDirectory
  abstract val downloadDir: DirectoryProperty
```

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineDownloadTask.kt
```diff
@@ -0,0 +1,46 @@
...
+      logger.info("manifestUrl is null")
+      return
+    }
+
+    if (downloadDir == null) {
+      logger.info("downloadDir file null")
+      return
+    }
+
+    ziplineDownloader.download(manifestUrl!!, downloadDir!!)
```
JakeWharton:
```suggestion
    ziplineDownloader.download(manifestUrl.get(), downloadDir.get().asFile)
```

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineDownloaderTest.kt
```diff
@@ -0,0 +1,34 @@
...
+package app.cash.zipline.gradle
+
+import java.io.File
+import org.junit.Test
+
+class ZiplineDownloaderTest {
+  private val ziplineDownloader = ZiplineDownloader()
+
+  @Test
+  fun `integration test to load from real url`() {
```
JakeWharton: Is this doing external HTTP calls? I'm not comfortable with such a test being enabled by default. My suite shouldn't fail just because I'm testing on a plane.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -105,43 +149,56 @@ class ZiplineLoader(
...
+    val upstreams = mutableListOf<Deferred<*>>()
+
+    /**
+     * Attempt to load from, in prioritized order: resources, cache, network;
+     * Then run postDownloadHook with the downloaded ZiplineFile
+     */
+    suspend fun load() {
+      val ziplineFileBytes = getZiplineFileBytesHook(id, module.sha256, module.url)()
       val ziplineFile = ZiplineFile.read(Buffer().write(ziplineFileBytes))
       upstreams.awaitAll()
```
JakeWharton: Unrelated: If you don't care about results you could call `launch` and store the resulting `Job` into the list and then call `joinAll()`. I'm not sure if there's any performance or memory implications, but my intuition is that a `Job` is either equal in memory or ever-so slightly lighter weight than a `Deferred`.

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineDownloaderTest.kt
swankjesse: Maybe MockWebServer??

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -58,21 +59,42 @@ class ZiplineLoader(
...
-    fileSystem = fileSystem,
-    directory = cacheDirectory,
-    maxSizeInBytes = cacheMaxSizeInBytes.toLong(),
-    nowMs = nowMs
-  )
-
-  suspend fun load(zipline: Zipline, url: String) {
-    // TODO fallback to manifest shipped in resources for offline support
+  // TODO add schema version checker and automigration for non-Android platforms
+  private val cache by lazy {
```
swankjesse: Rather than lazy lets use an In-memory driver on JVM?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -81,15 +103,37 @@ class ZiplineLoader(
...
+
+  suspend fun load(
+    zipline: Zipline,
+    manifest: ZiplineManifest,
+  ) {
+    loadImpl(manifest, getFromResourcesCacheThenNetworkHook(), loadIntoZiplineHook(zipline))
+  }
 
-    load(zipline, manifest)
+  private fun downloadToDirectoryHook(downloadDirectory: Path) = { ziplineFile: ZiplineFile, ziplineFileBytes: ByteString, id: String, sha256: ByteString ->
```
swankjesse: Maybe make an `ExecutionMode` interface with two implementations, `Buildtime` and `Runtime`. Then all the different bits are features of the interface.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
JakeWharton: https://github.com/cashapp/zipline/pull/447

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/strategy/LoadStrategy.kt
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.loader.strategy
+
+import app.cash.zipline.loader.ZiplineFile
+import okio.ByteString
+
+/**
+ * The ZiplineLoader will use a given [LoadStrategy] to process the dependency sorted modules
```
JakeWharton:
```suggestion
 * The ZiplineLoader will use a given [LoadStrategy] to process the dependency sorted modules.
```

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/strategy/LoadStrategy.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+interface LoadStrategy {
+  /**
+   * Get the module's [ZiplineFile]
+   *
+   * Implementations could get from network, resources, or cache as desired
```
JakeWharton:
```suggestion
   * Get the module's [ZiplineFile].
   *
   * Implementations could get from network, resources, or cache as desired.
```

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/strategy/LoadStrategy.kt
```diff
@@ -0,0 +1,45 @@
...
+   */
+  suspend fun getZiplineFile(
+    id: String,
+    sha256: ByteString,
+    url: String
+  ): ZiplineFile
+
+  /**
+   * Called once the [ZiplineFile] is retrieved to save to disk, seed cache, or load into a
+   * Zipline instance
```
JakeWharton:
```suggestion
   * Called once the [ZiplineFile] is retrieved to save to disk, seed cache, or load into a
   * Zipline instance.
```

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineDownloader.kt
```diff
@@ -0,0 +1,59 @@
...
+class ZiplineDownloader {
+  private val executorService = Executors.newSingleThreadExecutor { Thread(it, "Zipline") }
+  private val dispatcher = executorService.asCoroutineDispatcher()
+  private val client = OkHttpClient()
+
+  // Dummy base url
+  private val baseUrl = "http://10.0.2.2:8080/".toHttpUrl()
+
+  // Run SQLite in memory
+  private val driver = DriverFactory().createDriver()
```
JakeWharton: The need to do this makes me feel like maybe the layering here is slightly off or perhaps we're giving too much responsibility to a single type. I don't know the implementation well enough to give a concrete suggestions.

Right now this type knows about the local cache, the embedded cache, and the HTTP source. Am I right in thinking that we can layer these into an interceptor-like pattern? Where we have something like an `HttpModuleLoader`, an `FsCacheModuleLoader`, and an `EmbeddedModuleLoader` and they're a hierarchy which delegates to the next on a miss?

That would remove knowledge of the HTTP client, the filesystem, and the database from ZiplineLoader and have it take a single `ModuleLoader`.

It probably also eliminates the need for the strategy classes because the strategy just becomes how you layer the various loaders while also de-duplicating them since both strategies currently have the same logic for downloading.

adrw: Good call, I like how the interceptor pattern would better fit here and prevent having to plumb all possible dependencies into the ZiplineLoader constructor which as you've seen has made testing more difficult.

JakeWharton: Can be for a follow-up PR rather than keeping this open

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineDownloaderTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import okio.ByteString.Companion.encodeUtf8
+import okio.FileSystem
+import okio.Path.Companion.toOkioPath
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+
+class ZiplineDownloaderTest {
+  private lateinit var quickJs: QuickJs
+  private val webServer = MockWebServer()
```
swankjesse: nice

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineDownloaderTest.kt
```diff
@@ -0,0 +1,103 @@
...
+        .setBody(manifestJsonString)
+        .addHeader("Server", "treehouse-server")
+    )
+
+    // Enqueue the zipline file
+    webServer.enqueue(
+      MockResponse()
+        .setResponseCode(200)
+        .setBody(Buffer().write(alphaByteString))
+        .addHeader("Server", "treehouse-server")
```
swankjesse: Is this header useful?

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineDownloaderTest.kt
```diff
@@ -0,0 +1,103 @@
...
+    val manifestUrl = webServer.url("/latest/app/manifest.zipline.json").toString()
+
+    ziplineDownloader.download(manifestUrl, downloadDir)
+
+    // Confirm files successfully downloaded
+    val fileSystem = FileSystem.SYSTEM
+    val downloadDirPath = downloadDir.toOkioPath()
+    assertTrue(fileSystem.exists(downloadDirPath / ZiplineLoader.PREBUILT_MANIFEST_FILE_NAME))
+    assertEquals(manifestJsonString.encodeUtf8(), fileSystem.read(downloadDirPath / ZiplineLoader.PREBUILT_MANIFEST_FILE_NAME) { readByteString() })
+    assertTrue(fileSystem.exists(downloadDirPath / alphaBytecode(quickJs).sha256().hex()))
```
swankjesse: extract a variable for this sha256() ?

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/testFixturesJvm.kt
```diff
@@ -0,0 +1,53 @@
+package app.cash.zipline.gradle
```
swankjesse: nit: Copyright 2022

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/testFixturesJvm.kt
```diff
@@ -0,0 +1,53 @@
...
+import app.cash.zipline.QuickJs
+import app.cash.zipline.loader.CURRENT_ZIPLINE_VERSION
+import app.cash.zipline.loader.ZiplineFile
+import app.cash.zipline.loader.ZiplineManifest
+import app.cash.zipline.loader.ZiplineModule
+import okio.Buffer
+import okio.ByteString
+import okio.ByteString.Companion.toByteString
+
+val alphaJs = """
```
swankjesse: Would it be worthwhile for us to have a class `TextFixtures` that takes a `QuickJs` in its constructor?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineFile.kt
```diff
@@ -34,6 +35,21 @@ class ZiplineFile(
     sink.write(quickjsBytecode)
   }
 
+  fun toByteString(): ByteString {
+    val buffer = Buffer()
+    writeTo(buffer)
+    return buffer.readByteString()
+  }
+
+  override fun equals(other: Any?): Boolean = when (other) {
```
swankjesse: Make this a data class and get this for free?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineModuleLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+  ) {
+    val upstreams = mutableListOf<Job>()
+
+    /**
+     * Follow given strategy to get ZiplineFile and then process
+     */
+    suspend fun load() {
+      var ziplineFile: ZiplineFile? = null
+      for (interceptor in getterInterceptors) {
+        ziplineFile = interceptor.get(id, module.sha256, module.url)
```
swankjesse: Is Interceptor the right name for this? Maybe this is a Reader? Or a Loader?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineModuleLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+        if (ziplineFile != null) break
+      }
+      if (ziplineFile == null) throw IllegalStateException("Unable to get ZiplineFile for [module=$module]")
```
swankjesse: `check(...)` ?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineModuleLoader.kt
swankjesse: Maybe a Reader and a Writer ?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/interceptors/getter/FsCacheGetterInterceptor.kt
```diff
@@ -0,0 +1,13 @@
+package app.cash.zipline.loader.interceptors.getter
+
+import app.cash.zipline.loader.ZiplineCache
+import app.cash.zipline.loader.ZiplineFile
+import app.cash.zipline.loader.ZiplineFile.Companion.toZiplineFile
+import okio.ByteString
+
+class FsCacheGetterInterceptor(
+  private val cache: ZiplineCache,
+  ): GetterInterceptor {
```
swankjesse: I don’t think this is an interceptor because it doesn’t have a chaining feature . . . Maybe this is a reader?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/interceptors/getter/GetterInterceptor.kt
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * A list of [GetterInterceptor] delegate in order responsibility of getting the desired [ZiplineFile].
+ *
+ * If an interceptor does not get a file, it returns null and the next interceptor is called.
+ */
+interface GetterInterceptor {
+  /**
+   * Get the module's [ZiplineFile] or null if not found.
+   */
+  suspend fun get(
```
swankjesse: I think this interface might be simpler if it returns a ByteString ? Then the caller can convert?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/interceptors/handler/HandlerInterceptor.kt
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * After getting a [ZiplineFile], handler interceptors are iteratively called in order to handle.
+ */
+interface HandlerInterceptor {
+  /**
+   * Handle a [ZiplineFile] which can include loading it into a [Zipline] runtime
+   *  or saving to fileSystem.
+   */
+  suspend fun handle(
+    ziplineFile: ZiplineFile,
```
swankjesse: I think this interface is simpler if this is a `ByteString` instead of `ZiplineFile`

zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/interceptors/DownloadOnlyInterceptorTest.kt
```diff
@@ -0,0 +1,89 @@
...
+    cacheDbDriver.close()
+  }
+
+  @Test
+  fun getFileFromNetwork(): Unit = runBlocking {
+    httpClient.filePathToByteString = mapOf(
+      alphaFilePath to alphaBytecode(quickJs),
+    )
+
+    val ziplineFile = moduleLoader.load(
```
swankjesse: Formatting here is surprising

zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/testFixturesJvm.kt
```diff
@@ -0,0 +1,49 @@
+package app.cash.zipline.loader
+
+import app.cash.zipline.QuickJs
+import okio.Buffer
+import okio.ByteString
+import okio.ByteString.Companion.toByteString
+
+val alphaJs = """
```
swankjesse: Is this repeated with testFixturesCommon?

swankjesse (commented):
I think rename the interceptors to say, FileSource and FileTarget ?

swankjesse:
Great PR.

swankjesse (approved):

zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/testFixturesJvm.kt
adrw: There's some copy-pasting, I'll clean up in a follow up and potentially spin out a testing subproject for Loader to be used in Loader and Gradle Plugin tests.

JakeWharton (commented):
Nice!

PR #444, opened 2022-03-10
Directly delegate to the contextual serializer

swankjesse:
Calling encodeSerializableValue was causing us problems in JS
where values were being encoded with polymorphism unnecessarily.


JakeWharton (approved):

JakeWharton:
Any way to test?

swankjesse:
I have since learned that this is not the cause of the problem. I’m gonna merge this anyway as I think it’s a good simplification.

PR #447, opened 2022-03-11
Track a Job for loading work with no results

JakeWharton:


swankjesse (approved):
Any behavior consequence?

JakeWharton:
Nope. The previous `async` call was always returning `Unit` and the `awaitAll()` call produced a `List<Unit>` that was ignored.

PR #450, opened 2022-03-12
Introduce Zipline.json

### LINKED ISSUE #449: Create a property, Zipline.json, for out-of-band JSON encoding
We already expose `serializersModule`, but this is clumsy because kotlinx.serialization still needs the `SerializersModule` on the encoder to handle contextual serializers.

swankjesse:
This is the best way to encode and decode variables with
Zipline because it supports contextual serialization. Using
serializers alone is not sufficient.

This also simplifies endpoint initialization on Kotlin/JS.
The old code was mutable to work around initialization cycles.
With this we just forward calls to CallChannel to an instance
that might not be ready when the forwarder is created.

Closes: https://github.com/cashapp/zipline/issues/449

### DIFF
--- a/zipline/src/jsMain/kotlin/app/cash/zipline/ziplineJs.kt
+++ b/zipline/src/jsMain/kotlin/app/cash/zipline/ziplineJs.kt
@@ -21,7 +21,6 @@ import app.cash.zipline.internal.JsPlatform
 import app.cash.zipline.internal.bridge.CallChannel
 import app.cash.zipline.internal.bridge.Endpoint
 import app.cash.zipline.internal.bridge.ZiplineServiceAdapter
-import app.cash.zipline.internal.bridge.inboundChannelName
 import app.cash.zipline.internal.bridge.outboundChannelName
 import app.cash.zipline.internal.consoleName
 import app.cash.zipline.internal.eventLoopName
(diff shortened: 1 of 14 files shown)


zipline/src/jsMain/kotlin/app/cash/zipline/ziplineJs.kt
```diff
@@ -75,33 +75,16 @@ actual class Zipline internal constructor() {
...
-    // Connect platforms using our newly-bootstrapped channels.
-    val eventLoop = endpoint.take<EventLoop>(name = eventLoopName)
-    val console = endpoint.take<Console>(name = consoleName)
-    endpoint.bind<JsPlatform>(
-      name = jsPlatformName,
-      instance = RealJsPlatform(eventLoop, console),
-    )
-  }
+  internal val eventLoop: EventLoop = endpoint.take(eventLoopName)
+  internal val console: Console = endpoint.take(consoleName)
```
swankjesse, line 87: Fun fact: this line gets rewritten with Kotlin/IR!

zipline/src/jsMain/kotlin/app/cash/zipline/ziplineJs.kt
```diff
@@ -127,94 +110,28 @@ actual class Zipline internal constructor() {
...
-    val argumentsList = mutableListOf<Any?>()
-    for (argument in arguments) {
-      if (throwable == null && argument is Throwable) {
-        throwable = argument
-      } else {
-        argumentsList += argument
-      }
+      return Zipline(serializersModule)
+        .apply {
+          bind<JsPlatform>(jsPlatformName, GlobalBridge)
```
swankjesse, line 123: Another line that gets rewritten thanks to IR!

JakeWharton (approved):

PR #453, opened 2022-03-13
Use Gradle best practices for task structure

JakeWharton:


adrw (approved):
Thanks for making this idiomatic Gradle

PR #454, opened 2022-03-18
Embedded resources should use hex representation for file name

JakeWharton:
Previously the SHA256 was being used in byte form. If the hash started with 0x5C this was interpreted as a backslash which made the relative path resolution fail as that is the representation for a Windows-style root.

~There are no tests for this type... but clearly we need some.~ Ah, they're JVM-only!


veyndan (approved):

adrw:
Thanks for catching this

PR #456, opened 2022-03-22
Detect leaked services

oldergod:
Services are considered leaked if they have been GC'd before they were closed.

### DIFF
--- a/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
+++ b/zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
@@ -40,8 +40,10 @@ import org.jetbrains.kotlin.ir.builders.irGet
 import org.jetbrains.kotlin.ir.builders.irGetObject
 import org.jetbrains.kotlin.ir.builders.irInt
 import org.jetbrains.kotlin.ir.builders.irReturn
+import org.jetbrains.kotlin.ir.builders.irSet
 import org.jetbrains.kotlin.ir.builders.irString
 import org.jetbrains.kotlin.ir.builders.irTemporary
+import org.jetbrains.kotlin.ir.builders.irTrue
 import org.jetbrains.kotlin.ir.builders.irWhen
 import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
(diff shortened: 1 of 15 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
```diff
@@ -33,6 +33,7 @@ internal interface OutboundBridge {
     @PublishedApi internal val endpoint: Endpoint,
   ) {
     val serializersModule = json.serializersModule
+    var closed = false
```
oldergod, line 36: We don't define that one in IR, that's cool?

zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
```diff
@@ -763,16 +765,32 @@ internal class AdapterGenerator(
...
+        origin = IrDeclarationOrigin.DEFINED
+      }
+
+      val newCall = irCall(ziplineApis.outboundBridgeContextNewCall).apply {
+        dispatchReceiver = irGet(contextLocal)
         putValueArgument(0, irString(bridgedFunction.signature))
         putValueArgument(1, irInt(bridgedFunction.valueParameters.size))
       }
 
+      // If this is the close() function, tell the OutboundContext that this instance is closed.
```
swankjesse, line 786: Can you add a comment that matches what the source code would do?
```
// context.closed = true
```
I find it really helpful in IR stuff.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
swankjesse: I don’t follow?

swankjesse (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/outbound.kt
oldergod: We added code in IR to do `context.close = true`, don't we need IR code that add the line `var closed = false` ?

zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
oldergod: done

PR #460, opened 2022-03-29
Update the README and add a basic example

swankjesse:


README.md
```diff
@@ -1,213 +1,158 @@
 # Zipline
 
 This library streamlines using Kotlin/JS libraries from Kotlin/JVM and Kotlin/Native programs. It
-makes it possible to do continuous deployment of mobile apps, just like we do for servers and web
+makes it possible to do continuous deployment of mobile code, just like we do for servers and web
 applications.
```
JakeWharton: I would feel better if the second sentence was phrased as one possible use case rather than the only use case. Continuous deployment is only one use case among things like user scripts, plugin systems, etc. and mobile is only one domain where you could easily run this in desktop app, servers, games, etc. Don't want to pigeon hole the project.

README.md
```diff
@@ -1,213 +1,158 @@
...
 
 It'd be simpler to do continuous deploys via the App Store & Play Store! But that process is too
 slow. Even if we could release via these channels every day, we can’t guarantee that user’s devices
-will take the updates immediately. For Cash App we support apps that haven’t been updated for a full
-year.
+will take the updates immediately.
```
JakeWharton: This whole paragraph belongs with the second sentence above because it's part of the same single example usage.

README.md
```diff
@@ -1,213 +1,158 @@
...
+
+Zipline works by embedding the [QuickJS JavaScript engine][qjs] in your Kotlin/JVM or Kotlin/Native
+program. It's a small and fast JavaScript engine that's well-suited to embedding in mobile
```
JakeWharton:
```suggestion
program. It's a small and fast JavaScript engine that's well-suited to embedding in
```

README.md
```diff
@@ -1,213 +1,158 @@
...
+You can see the served application manifest at
+[localhost:8080/manifest.zipline.json](http://localhost:8080/manifest.zipline.json). It references
+all the code modules for the application.
 
-Imagine a world where we don't have [Okio's `ByteString`][okio]. Fortunately, there's a [Duktape
-builtin][dukdec] that allows us to convert hex-encoded strings back into UTF-8! We can easily set up a
-proxy that allows us to use it directly from our Java code. First, define a Java interface
-that declares the JavaScript methods you would like to call:
+In `jvmMain` we need write a program that downloads our Kotlin/JS code and calls it. We use
+`ZiplineLoader` which handles code downloading, caching, loading, and caching. We create a
```
JakeWharton: caching listed twice

README.md
```diff
@@ -1,213 +1,158 @@
...
+    dispatcher = dispatcher,
+    httpClient = OkHttpZiplineHttpClient(baseUrl, OkHttpClient()),
+    embeddedDir = "/".toPath(),
+    embeddedFileSystem = FileSystem.RESOURCES,
+    cacheDbDriver = DriverFactory().createDriver(),
+    cacheDir = "zipline-cache".toPath(),
+    cacheFileSystem = FileSystem.SYSTEM,
+    cacheMaxSizeInBytes = 10 * 1024 * 1024,
+    nowMs = System::currentTimeMillis,
+  )
```
JakeWharton, line 84: I don't think this is the API anymore. Didn't the interceptor-y redesign land?

README.md
```diff
@@ -1,213 +1,158 @@
...
 
+After using a bridged interface it must be closed so the peer object can be garbage collected. This
+is difficult to get right, so Zipline borrows ideas from [LeakCanary] and aggressively detects
+when a `close()` call is missed.
 
-## License
+### License
 
-    Copyright 2015 Square, Inc.
+    Copyright 2022 Block, Inc.
```
JakeWharton: We've never updated these in the past

samples/trivia/trivia-host/src/main/kotlin/app/cash/zipline/samples/trivia/launchZiplineJvm.kt
```diff
@@ -0,0 +1,53 @@
...
+    dispatcher = dispatcher,
+    httpClient = OkHttpZiplineHttpClient(baseUrl, OkHttpClient()),
+    embeddedDir = "/".toPath(),
+    embeddedFileSystem = FileSystem.RESOURCES,
+    cacheDbDriver = DriverFactory().createDriver(),
+    cacheDir = "zipline-cache".toPath(),
+    cacheFileSystem = FileSystem.SYSTEM,
+    cacheMaxSizeInBytes = 10 * 1024 * 1024,
+    nowMs = System::currentTimeMillis,
+  )
```
JakeWharton, line 45: ... I guess it didn't land. WOOF this API.

samples/trivia/trivia-shared/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+
+plugins {
+  kotlin("multiplatform")
+  kotlin("plugin.serialization")
+}
+
+kotlin {
+  jvm {
+    withJava()
+  }
```
JakeWharton, line 11:
```suggestion
  jvm()
```

samples/trivia/trivia-shared/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+
+  js {
+    browser()
+  }
+
+  sourceSets {
+    val commonMain by getting {
```
JakeWharton:
```suggestion
    commonMain {
```

samples/trivia/trivia-js/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+kotlin {
+  js {
+    browser()
+    binaries.executable()
+  }
+
+  sourceSets {
+    val commonMain by getting {
```
JakeWharton:
```suggestion
    commonMain {
```

JakeWharton (approved):

samples/trivia/trivia-shared/src/commonMain/kotlin/app/cash/zipline/samples/trivia/trivia.kt
```diff
@@ -0,0 +1,43 @@
...
+class Question(
+  val id: Int,
+  val text: String,
+)
+
+@Serializable
+class AnswerResult(
+  val correct: Boolean,
+  val message: String,
+)
```
veyndan, line 43: Do you think making this a `sealed class` (or `sealed interface` if they work?) would better indicate that it's flexible enough to deal cases other than a standard `class`? I'm just guessing that skimming this may give the incorrect assumption that we can only pass a standard `class` over the bridge.

```kotlin
sealed class Answer {

  abstract val message: String

  data class Correct(override val message: String): Answer()

  data class Wrong(override val message: String): Answer()
}
```

README.md
```diff
@@ -1,213 +1,158 @@
...
-## QuickJS
 
-This library uses the [QuickJS embeddable JavaScript engine][qjs] to execute Kotlin/JS. It's
-included in this library and supports both Android and the JVM.
+### Code Example
 
-### QuickJS Usage
+Let's make a trivia game that has fresh questions every day, even if our users don't update their
+apps. We define our interface in `commonMain` so that we can call it from Kotlin/JVM and implement
+it in Kotlin/JS.
```
veyndan: Could we have a link to the `samples/trivia` directory in the `README`? The only indication that the sample actually exists is within the Gradle commands below, which even then assumes that the user checked out zipline and is using it as the root project.

veyndan (approved):

README.md
swankjesse: Done.

-----

This library streamlines using Kotlin/JS libraries from Kotlin/JVM and Kotlin/Native programs. It
makes it fetching code as easy as fetching data:

 * For continuous deployment within mobile apps, just like we do for servers and web apps. It'd be
   simpler to do continuous deploys via the app stores! But that process is too slow and we can't
   guarantee that user’s devices will update immediately.
 * For user-customizable behavior and plugin systems
 * For updating business rules, like pricing or payments
 * For fresh content like games

README.md
swankjesse: YES! Fixed.

README.md
swankjesse: Ugh, one of the two hard problems.

README.md
swankjesse: This is still what the user-facing frontend of this API looks like.

I think this API is too big, with too many parameters and stuff that should be optional. I wanna work with @adrw to fix that. The examples here are real, though I think we could do better syntactically.

README.md
swankjesse: Yeah, that’s an artifact of me starting from scratch, gathering stuff from a few projects. Fixed.

samples/trivia/trivia-host/src/main/kotlin/app/cash/zipline/samples/trivia/launchZiplineJvm.kt
swankjesse: WOOF

samples/trivia/trivia-shared/build.gradle.kts
swankjesse: Fixed!

samples/trivia/trivia-js/build.gradle.kts
swankjesse: Fixed.

samples/trivia/trivia-shared/src/commonMain/kotlin/app/cash/zipline/samples/trivia/trivia.kt
swankjesse: I wanna keep this particular example as small as possible, and then maybe do recipes to talk through all the different features. It’s tempting to build out flows and other features into this example!

README.md
swankjesse: Great idea.

PR #461, opened 2022-03-29
Add Zipline-Cli to allow use by non-Gradle projects

### LINKED ISSUE #455: Add GraalVM executable as alternative to Gradle download task
For use in iOS tool chain or non-Gradle projects, a quick starting GraalVM executable can reuse the ZiplineDownloader to make seeding shipped-with-apps Zipline files easy for any project.

adrw:
[CLOSES #455] The CLI allows usage of Zipline code from non-Gradle projects.


swankjesse (approved):
This is brilliant

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -31,6 +33,8 @@ object Dependencies {
   val okHttpMockWebServer = "com.squareup.okhttp3:mockwebserver:4.9.1"
   val okio = "com.squareup.okio:okio:3.0.0"
   val okioFakeFileSystem = "com.squareup.okio:okio-fakefilesystem:3.0.0"
+  val picocli = "info.picocli:picocli:4.6.3"
+  val picocliCompiler = "info.picocli:picocli-codegen:4.6.3"
```
JakeWharton, line 37: Would strongly prefer clikt

zipline-cli/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+  manifest {
+    attributes("Automatic-Module-Name" to "app.cash.zipline.cli")
+    attributes("Main-Class" to "app.cash.zipline.cli.Main")
+  }
+}
+
+// resources-templates.
+sourceSets {
+  main {
+    resources.srcDirs("$buildDir/generated/resources-templates")
```
JakeWharton, line 25: Instead of sharing the path in two places, you can add the task provider directly to the source set and its outputs will automatically be included and a task dependency will automatically be added.

zipline-cli/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+  id("org.jetbrains.dokka")
+  id("com.vanniktech.maven.publish.base")
+  id("com.palantir.graal")
+  id("com.github.johnrengelman.shadow")
+}
+
+tasks.jar {
+  manifest {
+    attributes("Automatic-Module-Name" to "app.cash.zipline.cli")
+    attributes("Main-Class" to "app.cash.zipline.cli.Main")
```
JakeWharton, line 18: Why is this needed if it is specified below?

zipline-cli/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+tasks.register<Copy>("copyResourcesTemplates") {
+  from("src/main/resources-templates")
+  into("$buildDir/generated/resources-templates")
+  expand("projectVersion" to "${project.version}")
+  filteringCharset = Charsets.UTF_8.toString()
+}.let {
+  tasks.processResources.dependsOn(it)
+  tasks["javaSourcesJar"].dependsOn(it)
```
JakeWharton, line 72: These are the things you would be able to remove with the above source set / task dependency suggestion. In a good Gradle script you should never need to write a `dependsOn`.

zipline-cli/src/main/kotlin/app/cash/zipline/cli/Download.kt
```diff
@@ -0,0 +1,69 @@
...
+      downloadFileSystem = FileSystem.SYSTEM,
+    )
+    runBlocking {
+      ziplineDownloader.download(manifestUrl)
+    }
+  }
+
+  override fun run() {
+    download(manifestUrl = manifestUrl, downloadDir = downloadDir)
+  }
```
JakeWharton, line 64: This code is exactly whats in ZiplineDownloader or whatever it's called. Can we share?

zipline-cli/src/main/kotlin/app/cash/zipline/cli/Main.kt
```diff
@@ -0,0 +1,81 @@
...
+import picocli.CommandLine
+import picocli.CommandLine.Command
+import picocli.CommandLine.HelpCommand
+import picocli.CommandLine.IVersionProvider
+import picocli.CommandLine.Model.CommandSpec
+import picocli.CommandLine.Option
+import picocli.CommandLine.ParameterException
+import picocli.CommandLine.Spec
+
+@Command(name = NAME, description = ["Use Zipline without Gradle."],
```
JakeWharton, line 31: Need a better description. Zipline already works fine without Gradle. Something like "Standalone Zipline Loader file download"?

zipline-cli/src/main/resources-templates/zipline-cli-version.properties
```diff
@@ -0,0 +1 @@
+version=${projectVersion}
```
JakeWharton, line 1: If this is just for the version you can codegen a constant instead and save yourself a lot of build configuration and runtime code.

zipline-cli/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+
+  testImplementation(Dependencies.junit)
+  testImplementation(Dependencies.kotlinTest)
+}
+
+tasks.shadowJar {
+  mergeServiceFiles()
+}
+
+graal {
```
JakeWharton, line 44: We would need to build this on all three platforms on CI, right? Otherwise you have to build it locally yourself first to get the correct version for your host OS. Should we just use the application plugin which produces a zip containing a version that runs on all three platforms and starter scripts? I don't think there's any advantage to graal except a lot of extra build configuration.

PR #469, opened 2022-04-05
Get the JS engine to do JSON decoding

swankjesse:
This saved 35% of overall execution time in one benchmark.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/jsonCommon.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/jsonCommon.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2022 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/internal/jsonCommon.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.json.Json
+
+/*
+ * In QuickJS (version 2021-03-27) it's faster to ask the JavaScript engine to parse JSON to a
+ * dynamic value (ie. JSON.parse()) and bind that to a Kotlin Objects, than to parse and bind all
+ * in Kotlin/JS. In one sample this optimization reduced overall execution time by 35%.
+ */
+
+expect fun <T> Json.decodeFromStringFast(deserializer: KSerializer<T>, string: String): T
```
adrw, line 27: nice function name

adrw (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/jsonCommon.kt
JakeWharton, line 27: I do not like this name. Nothing should ever have "fast" in it. I would prefer something that either communicates behavior or has "internal" or something appended.

Also why is this public?

PR #470, opened 2022-04-06
Stop assuming the HTTP client has a base URL

swankjesse:
Instead do URL resolution inside the manifest that we've just
downloaded. That way manifests written to disk always have
absolute URLs in them.

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineDownloader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineDownloader.kt
@@ -66,7 +66,7 @@ class ZiplineDownloader(
     val manifestByteString = concurrentDownloadsSemaphore.withPermit {
       httpClient.download(manifestUrl)
     }
-    val manifest = Json.decodeFromString<ZiplineManifest>(manifestByteString.utf8())
+    val manifest = Json.decodeFromString(ZiplineManifest.serializer(), manifestByteString.utf8())
     return httpClient.resolveUrls(manifest, manifestUrl)
   }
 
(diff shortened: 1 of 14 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineDownloader.kt
```diff
@@ -66,7 +66,8 @@ class ZiplineDownloader(
     val manifestByteString = concurrentDownloadsSemaphore.withPermit {
       httpClient.download(manifestUrl)
     }
-    return Json.decodeFromString(manifestByteString.utf8())
+    val manifest = Json.decodeFromString<ZiplineManifest>(manifestByteString.utf8())
```
JakeWharton:
```suggestion
    val manifest = Json.decodeFromString(ZiplineManifest.serializer(), manifestByteString.utf8())
```
Your version still goes through a form of reflection.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -93,6 +93,7 @@ class ZiplineLoader(
       }
     }
 
-    return Json.decodeFromString(manifestByteString.utf8())
+    val manifest = Json.decodeFromString<ZiplineManifest>(manifestByteString.utf8())
```
JakeWharton:
```suggestion
    val manifest = Json.decodeFromString(ZiplineManifest.serializer(), manifestByteString.utf8())
```

JakeWharton (approved):

JakeWharton:
Much better

adrw (approved):
LG

PR #472, opened 2022-04-06
Refactor ZiplineLoader to have pluggable caching and embedded fs

adrw:


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -28,38 +31,72 @@ import okio.IOException
...
+    fsEmbeddedFetcher = null
+  )
+
+  fun withCache(
+    cacheDbDriver: SqlDriver, // SqlDriver is already initialized to the platform and SQLite DB on disk
+    cacheDir: Path,
+    cacheFileSystem: FileSystem,
+    cacheMaxSizeInBytes: Int = 100 * 1024 * 1024,
+    nowMs: () -> Long, // 100 MiB
+  ): ZiplineLoader = ZiplineLoader(
```
JakeWharton: I would prefer these with-ers move to be constructors on their respective fetchers. The loader ideally should not know anything about any subtypes of fetcher and it should only accept a single fetcher. Outside of the loader, you can compose fetchers together in any order and from any source that you want. Then that composite fetcher is the sole one supplied when creating a loader.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -28,38 +31,72 @@ import okio.IOException
...
-  private val embeddedFileSystem: FileSystem,
-  cacheDbDriver: SqlDriver, // SqlDriver is already initialized to the platform and SQLite DB on disk
-  cacheDir: Path,
-  val cacheFileSystem: FileSystem,
-  cacheMaxSizeInBytes: Int = 100 * 1024 * 1024,
-  nowMs: () -> Long, // 100 MiB
+  private val fsCachingFetcher: FsCachingFetcher? = null,
+  private val fsEmbeddedFetcher: FsEmbeddedFetcher? = null,
 ) {
+  constructor(dispatcher: CoroutineDispatcher, httpClient: ZiplineHttpClient) : this(
```
JakeWharton: Shouldn't the HTTP client also be a fetcher?

JakeWharton (commented):
I'm still struggling with this design that seems very tailored to our setup rather than being general purpose.

I'm imagining that you set it up something like this:
```kotlin
val fetcher = EmbeddedFetcher("/cache".toPath(), AssetFileSystem(context))
    + FsCachingFetcher(context.cacheDir.toPath(), FileSystem.SYSTEM)
    + HttpFetcher(client)
val loader = ZiplineLoader(dispatcher, fetcher)
```

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
adrw: I was going back and forth on that one, I'll set it up as a fetcher too

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
swankjesse: I thought the overall API was simpler with fetchers being implementation details . . . . do callers want to assemble their own thing or get something more opinionated? I prefer opinionated here

swankjesse (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
JakeWharton: I'm not sure I'd call an API where you have to supply the embedded filesystem and path; the cache filesystem, path, database, size, and clock; and the http client as simple. You're creating rigidity in the implementation while also exposing the implementation in the API.

The implementation keeps changing and many of the tests have to change with it because so much of the implementation detail is plumbed out through the API.

I don't think what I wrote is meaningfully distant from the current with-ers API but it's definitely more flexible and definitely more testable. Each fetcher can be tested in isolation and they don't have to live in this library. A composite fetcher can be tested in isolation to ensure it correctly falls back. And the loader integration to the fetcher can be tested without concern for what the fetch does.

adrw:
After incorporating interface ideas from @JakeWharton, it seemed like the difference between ZiplineLoader, ZiplineModuleLoader, and ZiplineDownloader were minimal enough to make the separation of abstractions not useful anymore so I folded them back in which seems to have been the right thing to do. Open to feedback, especially on which constructor signatures to keep.

adrw:
@JakeWharton @swankjesse mind doing another pass since a lot changed since the last review?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -16,34 +16,107 @@
...
+        cacheDir = cacheDir,
+        cacheFileSystem = cacheFileSystem,
+        cacheMaxSizeInBytes = cacheMaxSizeInBytes,
+        nowMs = nowMs,
+        httpClient = httpClient
+      ),
+    )
+  )
+
+  /** Default production constructor */
```
swankjesse: I don’t believe this comment and the one on line 54 can both be true.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -52,48 +125,145 @@ class ZiplineLoader(
...
+  fun withCachedNetwork(
+    cacheDbDriver: SqlDriver, // SqlDriver is already initialized to the platform and SQLite DB on disk
+    cacheDir: Path,
+    cacheFileSystem: FileSystem,
+    cacheMaxSizeInBytes: Int = 100 * 1024 * 1024,
+    nowMs: () -> Long, // 100 MiB
+  ): ZiplineLoader = ZiplineLoader(
+    dispatcher = dispatcher,
+    httpClient = httpClient,
+    fetchers = fetchers + listOf(
```
swankjesse: This is where I don’t like the current API. There’s an implied precedence of the calls now.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -52,48 +125,145 @@ class ZiplineLoader(
...
+    manifestUrl: String,
+  ): ZiplineManifest {
+    val manifestByteString = fetchers.fetch(
+      concurrentDownloadsSemaphore = concurrentDownloadsSemaphore,
+      id = PREBUILT_MANIFEST_FILE_NAME,
+      // Store manifest as the url SHA.
+      sha256 = manifestUrl.encodeUtf8().sha256(),
+      url = manifestUrl,
+      // Override fileName to support a static manifest location in embedded fs.
+      fileNameOverride = PREBUILT_MANIFEST_FILE_NAME,
```
swankjesse, line 136: I think this needs to always come from the user. We want to support multiple embedded applications.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/fetcher/Fetcher.kt
```diff
@@ -14,6 +16,29 @@ interface Fetcher {
...
+  url: String,
+  fileNameOverride: String? = null,
+): ByteString = concurrentDownloadsSemaphore
+  .withPermit {
+    var byteString: ByteString? = null
+    for (fetcher in this) {
+      byteString = fetcher.fetch(id, sha256, url, fileNameOverride)
+      if (byteString != null) break
+    }
+    if (byteString == null) throw IllegalStateException(
```
swankjesse: `check(byteString != null)` ?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/fetcher/FsCachingFetcher.kt
```diff
@@ -10,8 +15,54 @@ class FsCachingFetcher(
...
+    cache = createZiplineCache(
+      driver = cacheDbDriver,
+      fileSystem = cacheFileSystem,
+      directory = cacheDir,
+      maxSizeInBytes = cacheMaxSizeInBytes.toLong(),
+      nowMs = nowMs
+    ), delegate = delegate
+  )
+
+  constructor(
```
swankjesse: FWIW I dislike all these cascading constructors

swankjesse (approved):
Agreed that there’s a ton of different inputs for the 3 different data sources. That complexity is with us no matter what API we chose.

Unlike say, arbitrary orders of OkHttp Interceptors, there’s a single correct way to wire up the various fetchers and receivers.

Exposing users to the Fetchers abstraction and asking users to order their Fetchers correctly is more work than what they should need to do. In particular, bad things happen if you run the embedded fetcher after the caching fetcher because then the embedded fetcher never gets used.

I’d like for our API to have users provide each fetcher - cache/network/embedded, and each receiver - runtime/download in separate calls just to reduce accidents mixing up the embedded directory from the cache directory, or whatever. And also to make caching and embedding optional so you don’t have to pay for these features until you need them.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
JakeWharton: All these with-ers need to be removed. There's no reason this type should have any knowledge of fetcher subtypes. Fetcher precedence is entirely something that should be explicit

JakeWharton:
There's a single way that _we_ want to order things. There's no reason deployments of zipline need to load from HTTP or have a persistent cache. Even if they want to load from the network and have a cache we're assuming they want to use OkHttp to access individual files and store on the filesystem with SQLDelight.

An opinionated API will look great in our internal multiplatform codebase with this generic mechanism as an implementation detail.

swankjesse:
Works for me.

adrw:
Sounds good, I'll simplify and leave only the generic APIs and move the guardrails helper constructor method into our internal multiplatform codebase.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/fetcher/Fetcher.kt
adrw: `checkNotNull` :feelsgoodman:

PR #473, opened 2022-04-07
Merge EventListener's `callFailed` with `callEnd`

veyndan:
Removes the potential issue of overriding `callEnd` while forgetting to override `callFailed`, resulting in call failures without any handling (which could be surprising).


JakeWharton (approved):
Is nothing invoking these yet?

veyndan:
Just working on it at the moment. Got https://github.com/cashapp/zipline/pull/474 open which is a partial implementation of what we want (just does outbound non-suspending calls).

PR #474, opened 2022-04-07
Invoke callbacks for outbound calls

veyndan:


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -135,6 +135,7 @@ class Endpoint internal constructor(
 
     val outboundContext: OutboundBridge.Context = newOutboundContext(name)
```
JakeWharton: Is this the only invocation of `newOutboundContext`? It's the only one which was updated to set the `service`. I would expect since it's `@PublishedApi internal` that we need more places changed (maybe in the IR codegen) or that we can make the function `private` since it's only used from here.

zipline/src/jniTest/kotlin/app/cash/zipline/LoggingEventListener.kt
```diff
@@ -0,0 +1,22 @@
+package app.cash.zipline
```
swankjesse, line 16: supernit: Block copyright

swankjesse (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
veyndan: Tbh I forgot to check, though I'm looking at alternatives as I'm not a huge fan of the `lateinit var service`, and it seems to be hindering the garbage collection of the `Service` instance as initializing the variable is breaking `LeakedServicesTest#jvmLeaksService()`.

I'm thinking that maybe a marginally better (and less fragile) thing to do would be to just pass the `service` instance as an argument to `OutboundCall#invoke`. This'd solve the GC problem and forgetting to update all usages of `newOutboundContext`.

Also, I'm not too sure what the real world implementations of `EventListener` will look like, so idk how useful being passed the instance of the service will be, as we already are being passed the `name` (which I'm assuming is enough for debugging purposes?)

veyndan: Just going with what I said in the second paragraph.

JakeWharton: Yeah I'm not sure how useful it is to pass the service either. You're not allowed to interact with it in the callback, right? So the best you could do would be to read some metadata from it (such as its type).

veyndan: I haven't tried interacting with the service from the callback, but it does seem wrong, and I feel that such an API has the potential to be misunderstood (if you're not allowed to interact with it) or abused (if you are allowed to interact with it — like someone has a service where a function is only called once, so instead of closing it correctly, they just put an if-else on the service and function name in the `callEnd` callback).

Passing the type sounds good, though presumably they know this from the `name`? Passing the type as well might give the impression that the `name` and `type` are two disparate things. Idk how valid that concern is though.

swankjesse: I expect name will often be useless because it'll be something Zipline generated for pass-by-value.

Having the service is nice cause you can do a type check:

```
if (service is ZiplineTreehouseUi) {
  // find associated view somehow
}
```

I think calling methods on the services themselves, either in here or as side-effects is also plausible. Maybe ask a service to print its state before and after a call when debugging?

PR #475, opened 2022-04-07
Use `ArrayDeque` instead of blocking `Channel` calls

veyndan:
`events.receive()` would wait forever if nothing was sent to the `events` channel.

This might seem overly pedantic, but I'm extracting the `EventListener` implementation in this test into a separate class, and it's easier to do a bit of TDD when tests fail instead of hang.


oldergod (approved):
I forgot if it's racy or not, if it isn't; all good.
If it is, should we just set a 2s timeout or something like that? The error would be the same either way I think

swankjesse (approved):

PR #477, opened 2022-04-11
Add a 'service/' prefix when passing by reference

veyndan:


JakeWharton (approved):
Motivation?

veyndan:
Mainly to make the generated name a bit more useful than it currently is now. @swankjesse and I were discussing how a name like `"zipline/17"` isn't very useful while debugging, so by specifying the way that the service is being handled by a prefix slightly aids with that.

swankjesse (approved):

PR #482, opened 2022-04-11
Invoke  callbacks for inbound calls

veyndan:


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/inbound.kt
```diff
@@ -59,29 +62,33 @@ internal class InboundCall(
...
+          // Ignore unknown argument.
+          i += 2
+          continue
         }
       }
+      arguments += result
+      i += 2
+      if (i == encodedArguments.size) {
+        callStartResult = context.endpoint.eventListener.callStart(context.name, context.service, funName, arguments)
+      }
```
JakeWharton, line 84: Not sure about this as a hook to trigger a side-effect. I would expect the function to which this instance is passed (once all parameters are added) to invoke the side-effect.

veyndan: To clarify through an example, do you mean editing the generated code at:

https://github.com/cashapp/zipline/blob/69f10b170316aab5ebb3c1217f9ad205edac1027/zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt#L78-L85

to something like:

```kotlin
"fun ping(app.cash.zipline.SampleRequest): app.cash.zipline.SampleResponse" -> {
  val request = inboundCall.parameter(requestSerializer)
  // If there were more arguments it'd go here.
  context.endpoint.eventListener.callStart(
    context.name,
    service,
    "fun ping(app.cash.zipline.SampleRequest): app.cash.zipline.SampleResponse",
    listOf(
      request
    )
  )
  inboundCall.result(
    responseSerializer,
    service.ping(
      request
    )
  )
}
```

swankjesse (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/inbound.kt
swankjesse: Maybe this:

```
"fun ping(app.cash.zipline.SampleRequest): app.cash.zipline.SampleResponse" -> {
  val request = inboundCall.parameter(requestSerializer)
  inboundCall.callStart()
  val response = service.ping(request)
  inboundCall.callEnd(
    responseSerializer,
    response
  )
}
```

(I’m a bit reluctant to change generated code just ’cause everytime I do it takes a whole day.)

swankjesse:
I recommend merging as-is, we can fix the clumsy side-effecting thing in follow-up.

PR #484, opened 2022-04-12
Move Android app into a subfolder

JakeWharton:
Make room for iOS


veyndan (approved):

JakeWharton:
Hmm we should disable the publish jobs on PRs.

JakeWharton:
I guess it uploads jars and aars that you can play with. Whatever.

Issue #485, opened 2022-04-12
Migrate to version catalogs and enable dependabot for them


JakeWharton:
Death to `buildSrc/`

veyndan:
Dependabot doesn't support version catalogs yet (https://github.com/dependabot/dependabot-core/issues/3121)

PR #489, opened 2022-04-19
Automatically install the Zipline compiler in the Zipline plugin

### LINKED ISSUE #426: Zipline plugin on a Kotlin/JS project should automatically compile JS
This would be convenient!

Perhaps something like this?

```
val compileDevelopmentZipline by tasks.creating(ZiplineCompileTask::class) {
  dependsOn("compileDevelopmentExecutableKotlinJs")
  inputDir = file("$buildDir/compileSync/main/developmentExecutable/kotlin")
  outputDir = file("$buildDir/zipline")
}
val compileReleaseZipline by tasks.creating(ZiplineCompileTask::class) {
  dependsOn("compileReleaseExecutableKotlinJs")
  inputDir = file("$buildDir/compileSync/main/releaseExecutable/kotlin")
  outputDir = file("$buildDir/zipline")
}
```

### LINKED ISSUE #452: Automatically enhance the jsBrowserDevelopmentRun task
We put this boilerplate in all our Zipline Kotlin/JS projects:

```
val compileZipline by tasks.creating(ZiplineCompileTask::class) {
  dependsOn("compileDevelopmentExecutableKotlinJs")
  inputDir = file("$buildDir/compileSync/main/developmentExecutable/kotlin")
  outputDir = file("$buildDir/zipline")
}

val jsBrowserDevelopmentRun by tasks.getting {
  dependsOn(compileZipline)
}
```

We also always add `webpack.config.d/devserver.js` to each project to ensure the dev server serves `.zipline`-compiled resources:

```
if (config.devServer) {
  // Accept calls from any device. Typically this is a phone on the same WiFi network.
  config.devServer.host = "0.0.0.0";
  config.devServer.allowedHosts = 'all';

  // Don't inject served JavaScript with live reloading features.
  config.devServer.liveReload = false;
  config.devServer.hot = false;
  config.devServer.client = false;

  // Offer update notifications via a websocket.
  config.devServer.webSocketServer = {
    type: 'ws',
    options: {
      path: '/socket'
    }
  };

  // Don't format JS for Chrome/Firefox Developer Tools.
  config.devtool = false

  // Don't open Chrome.
  delete config.devServer.open;

  // Serve .zipline files.
  config.devServer.static.push('../../../../treehouse-apps-activity/build/zipline');
}
```

Adding this configuration automatically would also be nice. If serving compiled bytecode it’s not necessary to change the devServer’s liveReload, hot, or client. It’s also not necessary to set `devtool` to false.

swankjesse:
Closes: https://github.com/cashapp/zipline/issues/452
Closes: https://github.com/cashapp/zipline/issues/426


swankjesse:
# Not ready for review

Next steps:

 * [ ] Test a project that doesn’t build any JS
 * [ ] Move webpack.config.d boilerplate into plugin
 * [ ] Make `jsBrowserProductionRun` depend on `compileProductionExecutableKotlinJsZipline`
 * [ ] End-to-end test with plugin (serve code, download it with zipline-loader, call a function, assert output)
 * [ ] Either eliminate the `NodeJsRootPlugin` hacks or promote them into the plugin

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +40,39 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
     version = BuildConfig.KOTLIN_PLUGIN_VERSION,
   )
 
+  override fun apply(target: Project) {
+    super.apply(target)
+
+    // If the build has compileDevelopmentExecutableKotlinJs, create developmentExecutableZipline.
+    // Symmetrically for production.
+    target.tasks.whenTaskAdded { task ->
```
JakeWharton: This forces all tasks to be configured. I'm going to recommend a change of approach below, but if we were keeping this strategy the lazy API to use is `configureEach`.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +40,39 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+      if (task !is KotlinJsCompile) {
+        return@whenTaskAdded
+      }
+      val compileType = when (task.name) {
+        "compileDevelopmentExecutableKotlinJs" -> "developmentExecutable"
+        "compileProductionExecutableKotlinJs" -> "productionExecutable"
+        else -> return@whenTaskAdded
+      }
```
JakeWharton: This does not support a single JS target with a custom name or multiple JS targets. We will need to iterate over `kotlin.targets`, filter out those which are JS, and create a task associated with each target using its name.

Will need tests which do
```groovy
kotlin {
  js("customName")
}
```
and
```groovy
kotlin {
  js("foo")
  js("bar")
}
```
to confirm behavior.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +40,39 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+      target.tasks.create("${compileType}Zipline", ZiplineCompileTask::class.java) { createdTask ->
```
JakeWharton: `create` forces eager configuration of the task. Use `register` so that it's lazy. Syntax should mostly be identical otherwise.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +40,39 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+        createdTask.inputDir.set(
+          target.file("${target.buildDir}/compileSync/main/$compileType/kotlin")
```
JakeWharton: If this is a task output we ideally would get this as a property from that task so that it's both dynamic and implicitly creates a task dependency. Kotlin source are impossible to navigate on GitHub so I couldn't quickly find where this comes from on a target (presumbly it's an output of the main compilation) but we can track it down in the IDE.

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplinePluginTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.Test
+
+class ZiplinePluginTest {
+  private var quickJs: QuickJs? = null
+
+  @After
+  fun tearDown() {
+    quickJs?.close()
+  }
+
```
JakeWharton:
```suggestion
```

zipline-gradle-plugin/src/test/projects/basic/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+    maven {
+      url = file("$rootDir/../../../../../build/testMaven").toURI()
+    }
+    mavenCentral()
+    google()
+  }
+  dependencies {
+    println("ZIPLINE VERSION = " + project.property("ziplineVersion"))
+    classpath("app.cash.zipline:zipline-gradle-plugin:${project.property("ziplineVersion")}")
+    classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.10")
```
JakeWharton, line 11: You can import the root project version catalog to keep dependencies in these fixture projects in sync with the enclosing project.

https://github.com/cashapp/redwood/blob/c8454d716119b50e80ae44da6fc09a7529727463/redwood/redwood-gradle-plugin/src/test/fixture/compose-ui/settings.gradle#L7-L13

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: Oooooh that’s what I was looking for. Good call.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: I spend 30 minutes trying to convert `KotlinJsCompile` outputs to be my Zipline compile inputs, both instances of `DirectoryProperty`. Eventually I gave up.

zipline-gradle-plugin/src/test/projects/basic/build.gradle.kts
swankjesse: Oooooh yes that’s what I need.

JakeWharton: Just realized we don't have version catalogs here yet though! #485.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
JakeWharton: Hmm should just be `yourDP.set(otherDP)`, I think. `DirectoryProperty` is both a `Provider<Directory>` and a `Property<Directory>` the latter of which [has `set(Provider<? extends T>)`](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html#set-org.gradle.api.provider.Provider-). You should only have to do a `.map { }` or something if you need to change the value.

Example here (switch to Kotlin tab): https://docs.gradle.org/current/userguide/lazy_configuration.html#connecting_properties_together

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: New technique is this:

```
    val extension = project.extensions.findByType(KotlinMultiplatformExtension::class.java)
      ?: return

    extension.targets.withType(KotlinJsIrTarget::class.java).all { kotlinTarget ->
      ...
    }
```

I _think_ I need `.all {}` so I can create my task early enough in the lifecycle.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: Fixed.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: OMG I had no idea. I end up not needing that cause I’m starting with configuration and not tasks, though I might be able to make that work for more robustness.

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplinePluginTest.kt
swankjesse: Fixed.

zipline-gradle-plugin/src/test/projects/basic/build.gradle.kts
swankjesse: Yep.

swankjesse:
@JakeWharton this one’s ready for your attention!

JakeWharton:
Yeah I saw it late last night. I'm out until Monday though.

🏁 🚙💨

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
```diff
@@ -16,25 +16,75 @@
...
   // https://docs.gradle.org/current/userguide/custom_tasks.html#incremental_tasks
   // https://docs.gradle.org/current/userguide/lazy_configuration.html#working_with_files_in_lazy_properties
   // @get:Incremental
   @get:InputDirectory
-  abstract val inputDir: DirectoryProperty
+  lateinit var inputDir: File
 
   @get:OutputDirectory
-  abstract val outputDir: DirectoryProperty
+  lateinit var outputDir: File
```
JakeWharton, line 44: If this task type is not public and this field is not a `Property` subtype how do I do things such as zip these up and bundle them into my application without hardcoding task names and directory folders?

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +45,53 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+    val extension = project.extensions.findByType(KotlinMultiplatformExtension::class.java)
+      ?: return
+
+    extension.targets.withType(KotlinJsIrTarget::class.java).all { kotlinTarget ->
+      kotlinTarget.binaries.withType(JsIrBinary::class.java).all { kotlinBinary ->
+        registerCompileZiplineTask(project, kotlinBinary)
+      }
+    }
+
+    project.tasks.withType(Delete::class.java).getByName("clean") { clean ->
```
```suggestion
    project.tasks.named("clean", Delete::class.java).configure { clean ->
```
Existing version eagerly configures the task

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +45,53 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+    project.tasks.register(compileZiplineTaskName, ZiplineCompileTask::class.java) { createdTask ->
+      createdTask.description = "Compile .js to .zipline"
+      createdTask.dependsOn(kotlinBinary.linkTaskName)
+      val linkTask = kotlinBinary.linkTask.get()
+      val linkOutputDir = project.file(linkTask.kotlinOptions.outputFile!!).parentFile
+      createdTask.inputDir = linkOutputDir
+      createdTask.outputDir = linkOutputDir.parentFile.resolve("${linkOutputDir.name}Zipline")
+    }
+
+    project.tasks.withType(KotlinWebpack::class.java).all { kotlinWebpack ->
```
```suggestion
    project.tasks.withType(KotlinWebpack::class.java).configureEach { kotlinWebpack ->
```
`all` is eager

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
```diff
@@ -16,25 +16,75 @@
...
+  val webpackConfigFile: File by lazy {
+    project.projectDir.resolve(configFilePath)
+  }
 
   @TaskAction
   fun task() {
-    ZiplineCompiler.compile(inputDir.get().asFile, outputDir.get().asFile)
+    ZiplineCompiler.compile(inputDir, outputDir)
+
+    val webpackHome = project.rootDir.resolve("build/js/packages/placeholder-name")
```
JakeWharton: The build directory is not guaranteed to be a subdirectory of the project directory. You need to use `project.buildDir` to reference it.

This also relies on the directory structure inside of the build directory from a task we don't control which is likely to be brittle moving forward.

zipline-gradle-plugin/src/test/projects/jvmOnly/lib/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
+plugins {
+  kotlin("multiplatform")
+  id("app.cash.zipline")
+}
+
+kotlin {
+  jvm {
+    withJava()
+  }
```
JakeWharton:
```suggestion
  jvm()
```
No Java here

zipline-gradle-plugin/src/test/projects/basic/lib/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+
+plugins {
+  kotlin("multiplatform")
+  id("app.cash.zipline")
+}
+
+kotlin {
+  jvm {
+    withJava()
+  }
```
JakeWharton:
```suggestion
  jvm()
```
No Java here

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +45,53 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
     version = BuildConfig.KOTLIN_PLUGIN_VERSION,
   )
 
+  override fun apply(project: Project) {
+    super.apply(project)
+
+    val extension = project.extensions.findByType(KotlinMultiplatformExtension::class.java)
+      ?: return
```
JakeWharton, line 52: Eventually we'll need support for the plain JS plugin, too.

JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: Agreed.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -39,14 +45,53 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+      createdTask.description = "Compile .js to .zipline"
+      createdTask.dependsOn(kotlinBinary.linkTaskName)
+      val linkTask = kotlinBinary.linkTask.get()
+      val linkOutputDir = project.file(linkTask.kotlinOptions.outputFile!!).parentFile
+      createdTask.inputDir = linkOutputDir
+      createdTask.outputDir = linkOutputDir.parentFile.resolve("${linkOutputDir.name}Zipline")
+    }
+
+    project.tasks.withType(KotlinWebpack::class.java).configureEach { kotlinWebpack ->
+      kotlinWebpack.dependsOn(compileZiplineTaskName)
```
swankjesse, line 83: ... I assume there’s a mechanism to specify `dependsOn` without even configuring the task?

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
swankjesse: Oooh good call. Any advice here? Make the task and field public?

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
swankjesse: Fixed to use `rootProject.buildDir`. Agreed on the brittleness.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: ...I couldn’t find it.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
swankjesse: Made public for now. I think later we might want to expose additional state from the compilation, like production vs. development.

swankjesse (commented):
Thanks for the review. Nice having somebody who knows Gradle much better than I do reviewing this!

PR #500, opened 2022-04-27
Use DirectoryProperty instead of File

swankjesse:


zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
```diff
@@ -32,16 +35,18 @@ import org.gradle.api.tasks.TaskAction
...
+abstract class ZiplineCompileTask @Inject constructor(
+  objectFactory: ObjectFactory
+) : DefaultTask() {
   // TODO handle incremental and skip the quickjs compile when incremental
   // https://docs.gradle.org/current/userguide/custom_tasks.html#incremental_tasks
   // https://docs.gradle.org/current/userguide/lazy_configuration.html#working_with_files_in_lazy_properties
   // @get:Incremental
   @get:InputDirectory
-  lateinit var inputDir: File
+  val inputDir: DirectoryProperty = objectFactory.directoryProperty()
```
JakeWharton, line 46: Another way to do this (that's no more or less correct than what you have) is to mark these as `abstract` and Gradle will automatically do the factory stuff. Exact same result as what you have so no change required.

JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
swankjesse: Good to know though

PR #506, opened 2022-04-29
Migrate graal to version catalog

veyndan:


JakeWharton (approved):
I am still of the opinion that using Graal is a mistake but no one seemed to respond to my PR comments on the initial change.

veyndan:
Just linking for the curious: https://github.com/cashapp/zipline/pull/461#discussion_r838050108

Do you think it's worth to just create an issue(s) out of your comments on the PR? Otherwise, I'm guessing they're gonna be buried for a long time…

PR #511, opened 2022-04-29
Test for canceling the current scope

swankjesse:


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -83,25 +84,22 @@ class Endpoint internal constructor(
...
-          val inboundCall = InboundCall(handler.context, funName, encodedArguments)
-          val result = try {
-            handler.callSuspending(inboundCall)
-          } catch (e: Exception) {
-            inboundCall.resultException(e)
-          }
-          scope.ensureActive() // Don't resume a continuation if the Zipline has since been closed.
-          suspendCallback.call(result)
-        } finally {
+        cancelCallback.initScope(this@launch)
```
JakeWharton: `scope.launch` returns a `Job` which is usually the handle you want to hold on to for cancellation. I'm a bit unclear about the implications of the scope inside a launch with regard to cancellation. I _think_ it's unique, but I'm not sure. I mean it definitely should have the same `Job` in its context.

At the very least, it would feel more usual if we did `val job = scope.launch { .. }; cancelCallback.initJob(job)`.

Another thing that _might_ work is doing
```kotlin
val job = Job()
val cancelCallback = RealCancelCallback(job)
scope.launch(job) { .. }
```
I'm pretty sure this will attach the `Job` to the hierarchy correctly while giving us the cancellation signal without any kind of lateinit/nullable race. Not 100% on that either.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -83,25 +84,22 @@ class Endpoint internal constructor(
...
+        coroutineContext.job.invokeOnCompletion {
           remove(cancelCallbackName)
         }
```
JakeWharton: The above suggestions (especially the second one) would also simplify this allowing it to exist outside the launched block. This feels like a weird pattern in generally since I think a try/finally would work equally as well.

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
swankjesse: Ooooh just doing everything on the returned job works very nicely. Changed to do that and avoid the implied race in the other thing.

JakeWharton: Yeah wow your thing is even better! No type needed.

PR #515, opened 2022-04-30
Add developer ergonomics to the loader API

swankjesse:
Make it foolishly easy to configure ZiplineLoader on JVM and Android
platforms.


JakeWharton (approved):

JakeWharton:
👍 In my aging branch with an iOS emoji sample I also removed the embedded and caching fetchers.

PR #522, opened 2022-05-02
Inline abiFilters

veyndan:
As the goal is to remove `buildSrc`, I'm just inlining `ndkAbiFilters` as I haven't really thought about where we'd put it (I don't think having it in `libs.versions.toml` would make sense or is even possible) and I don't currently see the value in figuring out where to put it atm.


JakeWharton (approved):
If it's not shared, there's no need to centralize it. We only have a single Android build so this seems fine.

PR #530, opened 2022-05-05
Make CallChannel do more with encoded parameters

swankjesse:
This is a big refactoring to the internals of CallChannel:

 * Promote serviceName, funName, and callbackName as encoded data. The encoding is a bit clumsy but it means we need a lot less wiring.
 * Use one function for both regular and suspending invocations. The presence of a non-empty callback distinguishes the two cases.
 * Rename `invoke()` to `call()`.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/inbound.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/inbound.kt
@@ -61,9 +61,9 @@ internal class InboundCall(
   val encodedArguments: Array<String>,
 ) {
   internal lateinit var context: InboundBridge.Context
-  internal val serviceName: String
-  internal val funName: String
-  internal val callbackName: String
+  val serviceName: String
+  val funName: String
+  val callbackName: String
 
   private val arguments = ArrayList<Any?>(encodedArguments.size / 2)
   private var callStartResult: Any? = null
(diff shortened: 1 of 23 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/inbound.kt
```diff
@@ -58,14 +58,39 @@ internal interface InboundCallHandler {
...
+  internal lateinit var context: InboundBridge.Context
+  internal val serviceName: String
+  internal val funName: String
+  internal val callbackName: String
+
   private val arguments = ArrayList<Any?>(encodedArguments.size / 2)
   private var callStartResult: Any? = null
   private var i = 0
 
+  init {
```
swankjesse, line 72: This is where all the other deleted code went! Now encodedArguments does more stuff.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/inbound.kt
```diff
@@ -58,14 +58,39 @@ internal interface InboundCallHandler {
...
+    var serviceName: String? = null
+    var funName: String? = null
+    var callbackName: String? = null
+    var skippedArguments = false
+    while (i < encodedArguments.size) {
+      when (encodedArguments[i]) {
+        LABEL_SERVICE_NAME -> serviceName = encodedArguments[i + 1]
+        LABEL_FUN_NAME -> funName = encodedArguments[i + 1]
+        LABEL_CALLBACK_NAME -> callbackName = encodedArguments[i + 1]
+        else -> skippedArguments = true
```
JakeWharton, line 82: Is there a reason to support this case where they interleave?

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/inbound.kt
swankjesse: I’m trying to leave the door open to adding additional call data without breaking backward compatibility. I don’t feel strongly that it’s worth the indirection.

PR #537, opened 2022-05-26
Add LoaderEventsTest and get it passing

swankjesse:
This required some re-work of internals. In particular I needed
to pass exceptions differently for expected vs. unexpected failures.
Also it is now permitted to load a manifest without a URL. This
happens after loading a manifest by URL fails.


adrw (approved):
LG

zipline-loader-testing/src/engineMain/kotlin/app/cash/zipline/loader/testing/LoaderTestFixtures.kt
```diff
@@ -142,15 +142,15 @@ class LoaderTestFixtures(
       serializersModule = serializersModule,
       httpClient = httpClient,
       fetchers = listOf(
-        FsCachingFetcher(
```
adrw: Why do we need to flip them again?

swankjesse: For modules I wanna prefer fastest. Checksums ensure we have the right bytes so we want the most-local copy of the data we can find.

For manifests I wanna prefer freshest. We want server first, then cache, then embedded.

swankjesse: (I think this makes the current API a hazard because there’s huge performance consequences of getting it wrong, but it’ll appear to work.)

PR #540, opened 2022-05-26
Rename ZiplineLoader.load to loadOrFail

swankjesse:
Also change APIs to always return a Zipline for consistency

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -77,15 +77,22 @@ class ZiplineLoader(
     initializer: (Zipline) -> Unit = {}
   ): Zipline {
     return try {
-      createZiplineAndLoad(applicationName, manifestUrl, initializer)
+      createZiplineAndLoad(applicationName, manifestUrl, null, initializer)
     } catch (e: Exception) {
-      createZiplineAndLoad(applicationName, null, initializer)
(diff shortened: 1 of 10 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -145,63 +143,41 @@ class ZiplineLoader(
...
-        applicationName,
-      )
-      eventListener.applicationLoadEnd(applicationName, manifestUrl)
-    } catch (e: Exception) {
-      eventListener.applicationLoadFailed(applicationName, manifestUrl, e)
-      throw e
-    }
+    initializer: (Zipline) -> Unit = {},
+  ): Zipline {
+    return createZiplineAndLoad(applicationName, manifestUrl, null, initializer)
```
adrw, line 151: Nice to unify all paths into this same function

adrw (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -145,63 +143,41 @@ class ZiplineLoader(
     }
 
   /** Load application into Zipline without fallback on failure functionality. */
-  suspend fun load(
-    zipline: Zipline,
```
adrw, line 149: Is it true that we generally don't want to be able to pass a Zipline instance in to be loaded and instead want to always use the Loader to both construct a Zipline and load the code?

PR #544, opened 2022-05-30
Change the calling convention from string arrays to JSON

swankjesse:
This means we’re using kotlinx.serialization everywhere in Zipline.
That simplifies things and gives us opportunities to do more things
automatically, such as tracking the names of the pass-by-reference
services in calls.

This also introduces some new types:
 * `ZiplineFunction` – internal only (for now), stub to a single function
   of a single service instance
 * `ZiplineCall` – public API replacing 4 arguments on `EventListener`.
   The concrete implementation is encoded and decoded as the
   full calling convention


zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt
```diff
@@ -52,74 +51,55 @@ interface SampleService : ZiplineService {
...
-        private val service: SampleService,
-        override val context: InboundBridge.Context,
-      ) : InboundCallHandler {
-        private val requestSerializer = context.serializersModule.serializer<SampleRequest>()
-        private val responseSerializer = context.serializersModule.serializer<SampleResponse>()
-
-        private val supportedFunctionNames = listOf(
-          "fun close(): kotlin.Unit",
-          "fun ping(app.cash.zipline.SampleRequest): app.cash.zipline.SampleResponse",
+      override fun ziplineFunctions(
```
swankjesse, line 54: # ⚠️

SampleService.kt is a great place to start . . . see what the before/after generated code looks like, and everything else mostly falls out from there.

zipline/src/commonMain/kotlin/app/cash/zipline/EventListener.kt
```diff
@@ -38,12 +38,7 @@ abstract class EventListener {
    * @return any object. This value will be passed back to [callEnd] or [callFailed] when the call
    *     is completed. The base function always returns null.
    */
-  open fun callStart(
-    name: String,
-    service: ZiplineService,
-    functionName: String,
-    args: List<Any?>,
-  ): Any? {
+  open fun callStart(call: ZiplineCall): Any? {
```
swankjesse, line 41: Public API change!

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineCall.kt
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline
+
+/**
+ * A single invocation of a function declared on a [ZiplineService].
+ */
+interface ZiplineCall {
```
swankjesse, line 21: I’m 50/50 on renaming this to `Call`.

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineCall.kt
```diff
@@ -0,0 +1,39 @@
...
+  val serviceName: String
+
+  /**
+   * The service instance. Do not downcast this to a concrete type; it may be a generated
+   * implementation of the interface.
+   */
+  val service: ZiplineService
+
+  /** The signature of the function that was invoked. */
+  val functionName: String
```
swankjesse, line 32: @veyndan this becomes public API. Should we rename it to something like `functionSignature` maybe?

zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt
```diff
@@ -52,74 +51,55 @@ interface SampleService : ZiplineService {
...
+      class ZiplineFunction0(
+        argSerializers: List<KSerializer<*>>,
+        resultSerializer: KSerializer<*>,
+      ) : ZiplineFunction<SampleService>(
+        "fun ping(app.cash.zipline.SampleRequest): app.cash.zipline.SampleResponse",
+        argSerializers,
+        resultSerializer,
+      ) {
+        override fun call(service: SampleService, args: List<*>): Any? =
+          service.ping(args[0] as SampleRequest)
```
swankjesse, line 75: neat

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineCall.kt
veyndan: Yep that sounds good. It'd also closer align with the kdocs which is always good

zipline/src/commonMain/kotlin/app/cash/zipline/EventListener.kt
veyndan: Nice, it's a lot easier now to see the similarities between `callStart` and `callEnd`

veyndan (approved):
This looks awesome! It greatly simplifies things which is good 😄

PR #549, opened 2022-06-01
Use more generics internally in generated code

swankjesse:
This doesn't have a functional difference (today) but it gets us
much closer to what hand-written code does and that helps me sleep.

### DIFF
--- a/zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt
+++ b/zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt
@@ -39,13 +39,15 @@ data class SampleResponse(
  * Note that this interface is unused. It only serves as a sample for what the code generators
  * should consume and produce.
  */
-interface SampleService : ZiplineService {
+interface SampleService<T> : ZiplineService {
   fun ping(request: SampleRequest): SampleResponse
+
+  suspend fun reduce(request: List<T>): T
 
   companion object {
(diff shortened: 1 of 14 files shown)


zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt
```diff
@@ -59,60 +66,85 @@ interface SampleService : ZiplineService {
...
+
+      override fun ziplineFunctions(
+        serializersModule: SerializersModule,
+      ): List<ZiplineFunction<SampleService<TX>>> {
+        val types = types
+        val sampleRequestSerializer = serializersModule.serializer(types[0])
+        val sampleResponseSerializer = serializersModule.serializer(types[1])
+        val listOfTSerializer = serializersModule.serializer(types[2])
+        val unitSerializer = serializersModule.serializer(types[3])
+        val serializers = serializers
```
veyndan, line 118: Is `val types = types` and `val serializers = serializers` necessary?

veyndan (approved):

zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt
swankjesse: Not at all, but it’s what the generate code does. I want this to match that so I can more easily diff things as I make changes. (The generated code does it to reduce how much code we generate.)

PR #560, opened 2022-06-23
DNM WIP Add applicationId and prepareFunction to ZiplineManifest

adrw:


zipline-cli/src/test/kotlin/app/cash/zipline/cli/DownloadTest.kt
```diff
@@ -89,6 +89,8 @@ class DownloadTest {
     // Zipline files
     val applicationName = "app1"
     val manifest = ZiplineManifest.create(
+      applicationId = "",
```
swankjesse: should we make these nullable? Empty is kind of the worst of both worlds; it’s not usable but it looks usable

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompiler.kt
```diff
@@ -77,10 +78,32 @@ object ZiplineCompiler {
...
     val manifestFile = File(outputDir.path, "manifest.zipline.json")
     manifestFile.writeText(Json.encodeToString(manifest))
   }
+
+  internal fun getPrepareFunctionName(tsTypeDefinitionFile: File): String {
+    val contents = tsTypeDefinitionFile.readText()
+    return getPrepareFunctionName(contents)
+  }
+
+  internal fun getPrepareFunctionName(contents: String): String {
```
swankjesse, line 100: I don’t think we can do this generally. . . For example, this is the activity.d.ts file:

```
type Nullable<T> = T | null | undefined
declare const __doNotImplementIt: unique symbol
type __doNotImplementIt = typeof __doNotImplementIt
export namespace com.squareup.cash.js {
    class WebScriptsConfig {
        constructor(cashAppBaseUrl: string);
        get statementsBaseUrl(): string;
        get refundableWindowDays(): number;
        get paymentUrlBase(): string;
        get rightToRefundLinks(): any/* kotlin.js.Json */;
        get squareAddressUrl(): string;
        get helpCenterLinks(): any/* kotlin.js.Json */;
        get documentsBaseUrl(): string;
        get installmentPlanUrl(): string;
        get cryptoOrderReceiptUrl(): string;
    }
}
export namespace com.squareup.cash.js {
    class PaymentHistoryApi {
        constructor();
        getOfflinePaymentHistoryDataJson(serializedInitiatePaymentRequest: string, serializedAppVersion: string, serializedClientData: Nullable<string>): string;
        getPaymentHistoryDataJson(serializedPayment: string, serializedSender: string, serializedRecipient: string, serializedAppVersion: string, serializedClientData: Nullable<string>, serializedReceipt: Nullable<string>, serializedCheckDepositReceipt: Nullable<string>): string;
        getSerializedOfflinePaymentHistoryData(): any;
        getSerializedPaymentHistoryData(): any;
    }
}
export namespace com.squareup.cash.js.lib {
    const Logger: {
        log(message: string): void;
        warn(message: string): void;
        error(message: string): void;
    };
}
export namespace com.squareup.cash.treehouse.activity {
    function prepareTreehouse(): void;
}
export as namespace payment_history_treehouse;
```

swankjesse: What if we required a naming convention for these? Like, for example, the function name needed to start with `prepareTreehouse` ? We’d have to document it but it’d be automatic-enough

swankjesse: Another option that’s a bit more work . . . We could create our own annotation, `@PrepareTreehouse`

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineCompilerTest.kt
```diff
@@ -85,6 +86,23 @@ class ZiplineCompilerTest {
     }
   }
 
+  @Test
+  fun `extract qualified prepare function`() {
```
swankjesse, line 90: Love this test

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompiler.kt
```diff
@@ -77,10 +78,32 @@ object ZiplineCompiler {
...
+    val jsPackages = "(?<=(export namespace\\s)).*(?=\\s\\{)".toRegex().findAll(contents).toList().map { it.value }.first()
+    val functionName = "(?<=(function\\s)).*(?=\\(\\))".toRegex().findAll(contents).toList().map { it.value }.first()
```
swankjesse: Rather than `.first()` I’d prefer `.single()` since it explodes better if it’s ambiguous

swankjesse (approved):
This is awesome

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompiler.kt
adrw: Perfect, I was looking for that API but forgot what it was called

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompiler.kt
adrw: Agreed, I was suspicious that there'd be a more complex Treehouse repo that broke my brittle RegEX parsing. 

Some initial looking at annotations seems a bit tricky since the current Zipline compiler is only accepting JS files, not the Kotlin src, so we don't get any reflection. I can look into adding src code reflection if that would be best.

Alternatively, I could add a parameter to the ZiplineCompile task where the developer needs to specify the main function explicitly and have it match their code.

adrw:
After some brainstorming in meeting we landed on the following.

User wants to write this:

```kotlin
package com.squareup.activity

@ZiplineMain
fun prepareTreehouse(zipline: Zipline) {
}
```

Our tools eventually generate this! But for now you can hand-write it. Note that it’s always in the default package.

```kotlin
@JsExport
fun ziplineMain() {
  com.squareup.activity.prepareTreehouse()
}
```

The `ZiplineLoader` always runs this, where `myExportedProject` is the last element in the topological sort, which we assume is the entry point.

```javascript
val zipline = require("zipline-zipline").com.squareup.zipline.Zipline.INSTANCE
require("myExportedProject").ziplineMain(zipline)
```

zipline-cli/src/test/kotlin/app/cash/zipline/cli/DownloadTest.kt
adrw: In honesty, I was lazy in the first pass of filling testing values and was planning to come back to it in a follow up commit. As an API, it probably should be non-null so there isn't ambiguity on whether to include it in the manifest or not.

adrw:
Closing as this has been merged in https://github.com/cashapp/zipline/pull/567 and https://github.com/cashapp/zipline/pull/571

Issue #534, opened 2022-05-24
ZiplineCall


swankjesse:
Lets make a type from the arguments to `EventListener.callStart`:

```kotlin
interface Call {
  val name: String
  val service: ZiplineService
  val functionName: String
  val args: List<Any?>
}
```

This lets us change the signature of `EventListener`:

```diff
  open fun callStart(
-     name: String,
-     service: ZiplineService,
-     functionName: String,
-     args: List<Any?>,
+     call: Call,
  ): Any? {
    return null
  }
```

Next let’s add a property to `Call` with the services that were passed-by-reference to this call.

```kotlin
  /**
   * Services passed-by-reference in the arguments to this call.
   * Services may be either the arguments themselves, or they may
   * be the properties of the objects passed in the call.
   *
   * It is the receiver’s responsibility to close all services passed
   * in to a call.
   *
   * The keys in this map are the names that the services were bound
   * under. This is an opaque system-generated identifier.
   */
  val services: Map<String, ZiplineService>
```

Because they also have services, we should do likewise for the result.

```kotlin
interface CallResult {
  val result: Result<Any?>
  val services: Map<String, ZiplineService>
}
```

```diff
  open fun callEnd(
-     name: String,
-     service: ZiplineService,
-     functionName: String,
-     args: List<Any?>,
+     call: Call,
-     result: Result<Any?>,
+     result: CallResult,
    callStartResult: Any?
  ) {
  }
```

The motivation behind this is to support creation of a `DebugEventListener` that can attribute where a given leaked service came from.

swankjesse:
This introduces the initial type.
https://github.com/cashapp/zipline/pull/544

Some next steps:
 * [x] Change `CallChannel.call()` to take a single string, not an array
 * [x] Change `SuspendCallback` to take a `Result<T>`. This would prevent double-JSON encoding. (Maybe easier with generics?)
 * [x] Capture pass-by-reference services and expose them in the `Call` to event listeners

swankjesse:
Another idea I’m considering: split our call types in two:
 * `RealCall` is internal-only and doesn’t implement any interfaces. Maybe rename to `InternalCall`.
 * `Call` is a value object that describes a `RealCall`. In addition to its existing data we add more metadata around pass-by-reference services and the encoded request + response object payload sizes.

swankjesse:
Next steps for this issue:

 * [x] Change the pass-by-reference prefix from `service/` to `zipline/`
[...]

PR #564, opened 2022-07-04
Promote ZiplineFunction to the public API

swankjesse:
This splits ZiplineFunction into three types:

 * ZiplineFunction - the limited interface in the public API
 * ReturningZiplineFunction - implementation for non-suspending code
 * SuspendingZiplineFunction - implementation for suspending code

Generated code now uses the appropriate type when generating
implementations.

https://github.com/cashapp/zipline/issues/534


zipline-kotlin-plugin-hosted/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
```diff
@@ -533,7 +538,10 @@ internal class AdapterGenerator(
     return functionClass
   }
 
-  /** Override either `ZiplineFunction.call()` or `ZiplineFunction.callSuspending()`. */
+  /**
+   * Override either `ReturningZiplineFunction.call()` or
+   * `SuspendingZiplineFunction.callSuspending()`.
```
adrw, line 543: Good call to have both prefixed and not have an implicit return default

adrw (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
```diff
@@ -218,3 +224,41 @@ internal class ResultSerializer<T>(
...
   }
 }
+
+/** Returns a function that always throws [ZiplineApiMismatchException] when called. */
+private fun <T : ZiplineService> unknownFunction(
+  functionName: String,
+  inboundService: InboundService<*>?,
+  suspendCallback: SuspendCallback<Any?>?,
+): ZiplineFunction<T> {
+  val message = buildString {
```
JakeWharton, line 234: How about a local function to produce this rather than doing so eagerly?

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
swankjesse: I think it’s probably not worth the extra code, if only ’cause this message gets used 100% of the time when this method is called.

JakeWharton: Does creation of a `ZiplineFunction` object immediately translate to an invocation of that function? I thought these objects were perhaps created and not necessarily invoked.

PR #575, opened 2022-07-14
Import EdDSA code from Google Tink, v1.6.1 

swankjesse:


swankjesse:
This is definitely worth reviewing as two separate commits.

swankjesse:
Next steps as I see 'em:

 * [ ] Mechanically convert test code from `.java` to `.kt`.
 * [ ] Mechanically convert production code from `.java` to `.kt`. This will be annoying to perform, but the upside is we’ll have a single implementation that works on all of our target platforms.
 * [ ] Change packages from `com.google.crypto.tink.subtle` to `app.cash.zipline.loader.internal.tink.subtle`. 
 * [ ] Merge the `zipline-eddsa` submodule into the existing `zipline-loader` submodule.

At that point we’ll be able to start using EdDSA inside ZiplineLoader. I intend to follow the plan in our internal design doc.

JakeWharton (approved):

adrw:
I'm curious if it will be long term easier to maintain/upstream any changes if we keep the separate `zipline-eddsa` submodule vs merging it into Loader. The primary benefit to merging it as I understand it would be faster builds or are there other benefits?

swankjesse:
I don’t want to expose any public API for this. The fact that we ship our own EdDSA code rather than getting it from the platform is an implementation detail!

Issue #566, opened 2022-07-05
ZiplineLoader should produce a LoadResult


swankjesse:
We don’t capture much metadata when a load completes. This would be useful, especially when building systems that update running code if it isn’t fresh-enough.

```
class LoadResult {
    val zipline: Zipline

    val manifest: ZiplineManifest

    /** The date the loaded code was known to be fresh. */
    val freshAt: Long

    val source: CodeSource
}
```

```
enum class CodeSource {
  Embedded,
  Cache,
  Network,
}
```

### Example

The `freshAt` value would be the most-recent time we knew the URL of the manifest was up-to-date. For example:

#### 2001-01-02 at 02:00

 * Backend sets current URL to https://example.com/releases/v3

#### 2001-01-03 at 03:00

 * Client requests current URL, gets https://example.com/releases/v3
 * Client proceeds to download code

#### 2001-01-03 at 03:02

 * Client finishes downloading code. freshAt is `2001-01-01 at 03:00` because that’s when the URL was known fresh

#### 2001-01-04 at 04:00

 * Client requests current URL, gets https://example.com/releases/v3 (unchanged)
 * Client serves cached code. freshAt is now `2001-01-04 at 04:00` because the URL is revalidated

#### 2001-01-05 at 05:00

 * Client requests current URL, fails due to network error
 * Client serves cached code. freshAt is still `2001-01-04 at 04:00` because that’s the most recent URL freshAt time

adrw:
Maybe we include the URL and time we got it from the network.
LoadResult probably just is LoadedManifest.

adrw:
LoadedManifest probably should be stored in the Cache so we can get the timestamp when the code was downloaded and initially put in the cache.

swankjesse:
We designed a complete solution for the underlying problem and landed somewhere different. No action to take on this.

https://github.com/cashapp/zipline/issues/617

adrw:
After some more thought, we'll still want to wrap the returned loaded zipline with some metadata, specifically freshAtEpochMs

PR #583, opened 2022-07-15
APIs to sign and verify manifests

swankjesse:


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/fetcher/Fetcher.kt
```diff
@@ -76,6 +76,15 @@ internal interface Fetcher {
   )
 }
 
+/**
+ * A manifest plus the original bytes we loaded for it. We need the original bytes for signature
+ * verification.
+ */
+class FetchedManifest(
```
swankjesse: POSSIBLY this grows into the public API type we're planning,
https://github.com/cashapp/zipline/issues/566

JakeWharton: "Fetched" is not my favorite prefix here.

JakeWharton (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/fetcher/Fetcher.kt
swankjesse: Let's do LoadedManifest.

PR #586, opened 2022-07-25
Avoid unnecessary encode/decode steps on manifests

swankjesse:
When we roundtrip a JSON-encoded manifest through the ZiplineManifest
model class, we lose all unknown fields. This is potentially hazardous
when combining disk caching with forwards-compatibility.

Avoid this by avoiding roundtripping through the model class completely.
The only difficulty this introduces is relative-to-absolute URL mapping
needs to happen in the JSON object model and not the data class. This is
clumsier but tolerable.

This also introduces tests to confirm that unknown fields are retained.

Note that this drops EventListener callbacks when the manifest encounters
a parse failure when sourced from the cache or embedded file system. This
was untested previously and is not a critical behavior.

### DIFF
--- a/zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/HttpFetcherTest.kt
+++ b/zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/HttpFetcherTest.kt
@@ -1,0 +1,107 @@
+/*
+ * Copyright (C) 2022 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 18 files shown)


zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/HttpFetcherTest.kt
```diff
@@ -0,0 +1,107 @@
...
+      manifest = json.parseToJsonElement(manifestWithRelativeUrls),
+      baseUrl = "https://example.com/path/",
+    )
+
+    assertEquals(
+      """
+      |{
+      |    "modules": {
+      |        "./hello.js": {
+      |            "url": "https://example.com/path/hello.zipline",
```
adrw, line 58: nice use of snapshot tests here to clarify the changes to the entire manifest at a glance

adrw (approved):
LG

PR #587, opened 2022-07-25
Use target mode as task dependency discriminator

yissachar:
Checks the Kotlin executable target mode against the webpack target mode, and only adds the Zipline compile task as a task dependancy if they both match. This prevents the issue where running development webpack causes both compile Zipline production and compile Kotlin production to run. 

This approach is slightly fragile as it depends on both modes having the same name, but that's pretty unlikely to change.


swankjesse (approved):

swankjesse:
NICE

adrw (approved):

PR #590, opened 2022-07-26
Remove the hosted module

JakeWharton:
Shadow the sources directly. Remove specifying the native-specific artifact in the Gradle plugin as well since it would only be used with the legacy package names and our version would actually break in that case.

Didn't test this at all locally. Relying on CI.


zipline-kotlin-plugin/build.gradle.kts
```diff
@@ -5,20 +5,28 @@ import com.vanniktech.maven.publish.MavenPublishBaseExtension
...
+  kotlin("kapt")
+  id("com.github.gmazzo.buildconfig")
   id("org.jetbrains.dokka")
   id("com.vanniktech.maven.publish.base")
+  id("com.github.johnrengelman.shadow")
 }
 
 dependencies {
-  compileOnly(projects.ziplineKotlinPluginHosted)
+  compileOnly(kotlin("compiler"))
```
swankjesse, line 16: When I tried it, I depended on compiler-embedded and dropped the shadowJar

swankjesse: (this also works!)

swankjesse (approved):

zipline-kotlin-plugin/build.gradle.kts
JakeWharton: Oh fun. Let's do that. I'll PR.

JakeWharton:
Oh I just found your other PR. Whoops. I thought you may have just missed my comment on your PR since it was after the first review :facepalm:

PR #594, opened 2022-07-27
Change internal service names to be more useful

swankjesse:
We need anonymous names for services passed-by-reference and callbacks.
Previously the prefix was like service/1, without separate namespaces
for JS-inbound services and host-inbound services.

The new prefixes are zipline/host-1 for host-inbound services and
zipline/js-1 for JS-inbound services.


adrw (approved):
LG

PR #596, opened 2022-07-27
Tidy up visibilities in zipline-loader

swankjesse:
This adds the 'internal' modifier where possible.
It moves internal APIs to the internal package.


adrw (approved):
LG

PR #599, opened 2022-07-27
Promote some loader tests to commonTest

swankjesse:
Unfortunately there's a bug where Okio's FakeFileSystem doesn't
work on Kotlin 1.7.0 on native due to a transitive dependency on
kotlinx-datetime.

https://youtrack.jetbrains.com/issue/KT-52554

I'm going to need to update Okio and ship a release. In the interim
just use the regular file system for these tests.

### DIFF
--- a/zipline-loader/src/androidTest/kotlin/app/cash/zipline/loader/loaderTestsAndroid.kt
+++ b/zipline-loader/src/androidTest/kotlin/app/cash/zipline/loader/loaderTestsAndroid.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2022 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


adrw (approved):

JakeWharton (approved):

swankjesse:
FYI I’ve added a commit to disable Android tests for zipline-loader.

Previously we had JVM tests only. Now that I’ve promoted tests to `commonMain` we’re seeing the Android tests crash because QuickJS and SQLite are not available.

zipline-loader/src/androidTest/kotlin/app/cash/zipline/loader/loaderTestsAndroid.kt
```diff
@@ -0,0 +1,34 @@
...
+// them on the JVM.
+
+actual val systemFileSystem = FileSystem.SYSTEM
+
+actual fun testZiplineLoader(
+  dispatcher: CoroutineDispatcher,
+  httpClient: ZiplineHttpClient,
+  eventListener: EventListener,
+  manifestVerifier: ManifestVerifier?,
+): ZiplineLoader = error("testZiplineLoader not available for Android")
```
adrw, line 34: What's the motivation of moving all these loader tests to common if the multiplatform targets aren't runnable?

swankjesse: I’m growing it from JVM-only to JVM + Native. Doing Android turns out to be more annoying because the test infra assumes some Android unit tests can run on the JVM, which they can’t.

PR #600, opened 2022-07-27
[CLOSES #597] Extract Mac build script

### LINKED ISSUE #597: Extract bash script for Mac builds
- Github Actions currently has the steps for how to do builds on Macs (Intel and ARM). 
- Pulling that out of YAML into a Bash script will let us add docs for new devs to easily run locally to start contributing to Zipline
- Stretch: don't even require running the script manually locally, setup a Gradle task that runs whenever native code changes or the native artifacts don't exist

adrw:


JakeWharton (approved):
Would suggest running shellcheck over it if you get the chance but it's not really urgent.

adrw:
@JakeWharton already done ✅

PR #603, opened 2022-07-27
Application plugin

JakeWharton:
Migration to the application plugin for building the CLI.


swankjesse (approved):

zipline-cli/README.md
```diff
@@ -1,6 +1,6 @@
 zipline-cli
```
adrw: With the `zipline-cli` bash file removed, could the docs be updated for how to use the CLI after these changes?

adrw (approved):

zipline-cli/README.md
JakeWharton: Yeah I was being lazy. It's basically the same, though, but the script is inside the zip now.

JakeWharton: Or if you're building locally it's exactly the same it's just in a new location.

Issue #573, opened 2022-07-13
Throw Different Exceptions for User- vs. API crashes 


swankjesse:
Currently the exception types we propagate are not useful. Instead we should do something like NoSuchMethodError / InvocationTargetException.

PR #609, opened 2022-07-31
Change how throwables are serialized

swankjesse:
This PR is backwards-incompatible and applications built with
this will not work with runtimes built on an earlier version,
or vice versa.

This splits serialized exceptions to express type information
separately. It will allow us to customize which exception types
are propagated. Initially this is used to make handing a
ZiplineApiMismatchException easy and safe.


swankjesse:
Closes: https://github.com/cashapp/zipline/issues/573

JakeWharton (approved):

PR #613, opened 2022-08-03
Use main function as identifier, not call

adrw:
Main function should just be an identifier, not the entire call to allow for more flexible use.


swankjesse (approved):
TY!

dnagler (approved):
Thanks!

PR #620, opened 2022-08-04
Add fresh at timestamp to cache DB schema

adrw:
New filename for cache DB will prevent collision with old versions during pre 1.0.0 development.


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCache.kt
```diff
@@ -251,7 +251,9 @@ internal class ZiplineCache internal constructor(
       manifest_for_application_name = manifestForApplicationName,
       file_state = FileState.DIRTY,
       size_bytes = 0L,
-      last_used_at_epoch_ms = nowMs()
+      last_used_at_epoch_ms = nowMs(),
+      // TODO pass this in from http client
```
swankjesse, line 255: 👍🏻

swankjesse (approved):

PR #626, opened 2022-08-04
Pipe freshAt timestamp through from HttpFetcher, add LoadedZipline

### LINKED ISSUE #566: ZiplineLoader should produce a LoadResult

adrw:
[CLOSES #566]

### DIFF
--- a/zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineGradleDownloaderTest.kt
+++ b/zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineGradleDownloaderTest.kt
@@ -19,6 +19,7 @@ package app.cash.zipline.gradle
 import app.cash.zipline.loader.ZiplineManifest
 import app.cash.zipline.loader.internal.getApplicationManifestFileName
 import app.cash.zipline.loader.testing.LoaderTestFixtures
+import app.cash.zipline.loader.testing.LoaderTestFixtures.Companion.assertDownloadedToEmbeddedManifest
 import java.io.File
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
@@ -27,7 +28,6 @@ import kotlinx.serialization.json.Json
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import okio.Buffer
-import okio.ByteString.Companion.encodeUtf8
 import okio.FileSystem
 import okio.Path.Companion.toOkioPath
 import org.junit.After
@@ -89,9 +89,11 @@ class ZiplineGradleDownloaderTest {
     val fileSystem = FileSystem.SYSTEM
     val downloadDirPath = downloadDir.toOkioPath()
     assertTrue(fileSystem.exists(downloadDirPath / getApplicationManifestFileName(applicationName)))
-    assertEquals(
-      manifestJsonString.encodeUtf8(),
-      fileSystem.read(downloadDirPath / getApplicationManifestFileName(applicationName)) { readByteString() })
+
+    val actualManifestByteString = fileSystem.read(
+      downloadDirPath / getApplicationManifestFileName(applicationName)
+    ) { readByteString() }
+    assertDownloadedToEmbeddedManifest(manifest, actualManifestByteString)
     assertTrue(fileSystem.exists(downloadDirPath / testFixtures.alphaSha256Hex))
     assertEquals(
       testFixtures.alphaByteString,
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -169,7 +169,7 @@ class ZiplineLoader internal constructor(
     manifestUrlFlow: Flow<String>,
     oldestCodeToLoad: Duration = Duration.INFINITE,
     initializer: (Zipline) -> Unit,
-  ): Flow<Zipline> {
+  ): Flow<LoadedZipline> {
     return flow {
       var isFirstLoad = true
       var previousManifest: ZiplineManifest? = null
@@ -189,7 +189,7 @@ class ZiplineLoader internal constructor(
           try {
             val networkZipline = loadFromManifest(applicationName, networkManifest, initializer)
             cachingFetcher?.pin(applicationName, networkManifest) // Pin after successful loads.
-            emit(networkZipline)
+            emit(LoadedZipline(networkZipline, networkManifest.freshAtEpochMs!!))
             previousManifest = networkManifest.manifest
           } catch (e: Exception) {
             cachingFetcher?.unpin(applicationName, networkManifest) // Unpin after failed loads.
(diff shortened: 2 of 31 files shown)


zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineGradleDownloaderTest.kt
```diff
@@ -89,9 +91,14 @@ class ZiplineGradleDownloaderTest {
...
     assertTrue(fileSystem.exists(downloadDirPath / getApplicationManifestFileName(applicationName)))
-    assertEquals(
-      manifestJsonString.encodeUtf8(),
-      fileSystem.read(downloadDirPath / getApplicationManifestFileName(applicationName)) { readByteString() })
+
+    val actualManifestByteString = fileSystem.read(
+      downloadDirPath / getApplicationManifestFileName(applicationName)
+    ) { readByteString() }
+    val actualManifest = Json.decodeFromString<ZiplineManifest>(actualManifestByteString.utf8())
+    assertNotNull(actualManifest.builtAtEpochMs) // builtAtEpochMs has been filled out on download
```
swankjesse: Nice

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -180,12 +180,16 @@ class ZiplineLoader internal constructor(
         //  - pin the application if the load succeeded; unpin if it failed.
         withLifecycleEvents(applicationName, manifestUrl) {
           val networkManifest = fetchManifestFromNetwork(applicationName, manifestUrl)
-          if (networkManifest.manifest == previousManifest) return@withLifecycleEvents // Unchanged.
+          if (networkManifest.manifest == previousManifest) {
+            // Unchanged. Update freshness timestamp in cache DB.
```
swankjesse, line 184: Nice

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCache.kt
```diff
@@ -173,13 +175,19 @@ internal class ZiplineCache internal constructor(
...
   }
 
   /** Pins manifest and unpins all other files and manifests */
   fun pinManifest(applicationName: String, loadedManifest: LoadedManifest) {
     val manifestBytes = loadedManifest.manifestBytes
-    val manifestMetadata = getOrPutManifest(applicationName, manifestBytes) ?: return
+    val manifestMetadata = getOrPutManifest(
+      applicationName = applicationName,
+      content = manifestBytes,
+      putFreshAtMs = loadedManifest.freshAtEpochMs ?: throw IllegalArgumentException(
```
swankjesse: nit: is this a likely exception? Might be simpler as just `!!`

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCache.kt
```diff
@@ -238,22 +246,28 @@ internal class ZiplineCache internal constructor(
...
     isManifest: Boolean,
+    manifestFreshAtMs: Long? = null
   ): Files? = try {
     val manifestForApplicationName = if (isManifest) {
       applicationName
     } else {
       null
     }
 
+    val freshAtEpochMs = if (isManifest) {
```
swankjesse, line 255: Perfect

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCache.kt
```diff
@@ -367,4 +381,19 @@ internal class ZiplineCache internal constructor(
   private fun path(metadata: Files): Path {
     return directory / "entry-${metadata.id}.bin"
   }
+
+  /**
+   * Update file record freshAt timestamp to reflect that the manifest is still seen as fresh
+   */
+  fun updateManifestFreshAt(applicationName: String, loadedManifest: LoadedManifest) {
+    val freshAtMs = loadedManifest.freshAtEpochMs ?: throw IllegalArgumentException(
```
swankjesse: Could this be a `!!` ? It’s all internal callers

Maybe a better way to describe this... can we trigger this being null in calling code?
 - if we can, use a nice exception as you have here, and write a test to confirm it
 - if we can’t, use `!!` and save some lines of code

As-is, it reads like this is a thing that can actually happen and I don’t think it can

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCache.kt
```diff
@@ -367,4 +381,19 @@ internal class ZiplineCache internal constructor(
...
+      "Loaded manifest is missing freshtAtMs timestamp."
+    )
+    val manifestMetadata =
+      getOrPutManifest(applicationName, loadedManifest.manifestBytes, freshAtMs) ?: return
```
swankjesse: nit: add a comment after `?: return` like say `?: return // Pruned.`

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/fetcher/HttpFetcher.kt
```diff
@@ -42,6 +43,7 @@ internal class HttpFetcher(
   ) = fetchByteString(applicationName, url)
 
   suspend fun fetchManifest(applicationName: String, url: String): LoadedManifest {
+    val fetchStartMs = nowMs()
```
swankjesse, line 46: nice

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheTest.kt
```diff
@@ -333,6 +334,24 @@ class ZiplineCacheTest {
     }
   }
 
+  @Test
+  fun manifestFreshAtMs(): Unit = runBlocking {
```
swankjesse, line 338: Nice

swankjesse (approved):
LOVE IT

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCache.kt
adrw: Good point, I missed updating that. Ended up that this case isn't possible so no exception or `!!` required.

PR #633, opened 2022-08-05
Move some manifest properties to 'unsigned'

swankjesse:
This is a new property for all data not covered by signatures.


swankjesse:
@JakeWharton is `unsigned` a terrible name for data that isn't covered by signatures? In particular is it super confusing that I’m using the word 'unsigned' to mean 'not signed' instead of 'neither positive nor negative' ?

adrw (approved):

adrw:
I agree `unsigned` might not be the right term, not sure if `_private`, `ignoreWhenSigning`, or `doNotIncludeInSignature` is any better

JakeWharton:
Maybe insecure? Unsigned definitely is accurate but immediately evocative of the wrong thing.

PR #634, opened 2022-08-05
Replace webpack with dedicated static file server

yissachar:
Replaces the webpack serve with [http4k](https://www.http4k.org/) to serve the Zipline build dir. Webpack was trying to also serve the JS build dir, resulting in warnings that we don't care about, as well as being slower overall: incremental builds are now ~700ms faster with this PR.

Note that this adds new `serveDevelopmentZipline`/`serveProductionZipline` tasks that should be run instead of `jsBrowserDevelopmentRun`/`jsBrowserProductionRun` tasks.

Unfortunately I'm not quite sure how add tests for this; the GradleRunner test kit doesn't seem to have file watching support, even if I add --watch-fs. Open to any ideas!


.gitignore
```diff
@@ -1,3 +1,6 @@
+# macOS
+.DS_Store
```
JakeWharton: This should go in your global gitignore rather than in each project. 

Not the guide I followed, but this was the first google result for a guide: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

yissachar: Want me to remove? A bunch of our other projects have this per-project but I'm fine either way.

swankjesse:
I think this is awesome but there’s a huge problem – I don’t know if http4k is safe to embed in a Gradle plugin. In particular Gradle’s dependencies and http4k’s dependencies need to fit well together.

swankjesse:
Ooooh this is promising:

> The core library has no dependencies and it’s all you need to create fully-functional applications.

swankjesse:
One really nice thing this allows us to do is delete our hacky webpack config for web serving. Hooray!

.gitignore
JakeWharton: Please! We should be removing it anywhere it remains in favor of platform-global excludes.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -73,10 +70,11 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
       createdTask.outputDir.set(linkOutputDir.parentFile.resolve("${linkOutputDir.name}Zipline"))
     }
 
-    project.tasks.withType(KotlinWebpack::class.java).configureEach { kotlinWebpack ->
-      if (kotlinBinary.mode.toString().equals(kotlinWebpack.mode.toString(), ignoreCase = true)) {
-        kotlinWebpack.dependsOn(compileZiplineTaskName)
-      }
+    val serveTaskName = "serve${kotlinBinary.mode.toString().lowercase().capitalized()}Zipline"
```
adrw: `.lowercase().capitalized()` isn't the lowercase redundant?

adrw (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -73,10 +70,11 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+    project.tasks.register(serveTaskName, ZiplineServeTask::class.java) { createdTask ->
```
adrw, line 80: Can you add update docs if any of the commands for continuous development of single build have changed?

adrw (commented):
Seems promising, I'll try it out locally next week

zipline-gradle-plugin/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ dependencies {
   implementation(libs.kotlinx.serialization.json)
   implementation(libs.okHttp.core)
   implementation(libs.okio.core)
+  implementation("org.http4k:http4k-core:4.27.4.0")
```
JakeWharton: Please move to `gradle/libs.versions.toml`

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -50,21 +50,18 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
-      clean.delete.add(target.projectDir.resolve(ZiplineCompileTask.configFilePath))
-    }
   }
 
   private fun registerCompileZiplineTask(project: Project, kotlinBinary: JsIrBinary) {
     // Like 'compileDevelopmentExecutableKotlinJsZipline'.
     val linkTaskName = kotlinBinary.linkTaskName
     val compileZiplineTaskName = "${linkTaskName}Zipline"
 
+    fun getZiplineOutputDir() {}
```
```suggestion
```

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
JakeWharton: Both `lowercase()` and `capitalized()` should be given `Locale.US` so that we do not create different task names based on the user's locale.

Also I think you want `.name` not `.toString()`, assuming the `mode` is an enum. `toString()` defaults to `name`, but `toString()` also can be customized whereas `name` cannot be (other than renaming the constant, of course).

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -73,10 +70,11 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
 
-    project.tasks.withType(KotlinWebpack::class.java).configureEach { kotlinWebpack ->
-      if (kotlinBinary.mode.toString().equals(kotlinWebpack.mode.toString(), ignoreCase = true)) {
-        kotlinWebpack.dependsOn(compileZiplineTaskName)
-      }
+    val serveTaskName = "serve${kotlinBinary.mode.toString().lowercase().capitalized()}Zipline"
+    project.tasks.register(serveTaskName, ZiplineServeTask::class.java) { createdTask ->
+      createdTask.description = "Serves Zipline files"
+      createdTask.dependsOn(compileZiplineTaskName)
+      createdTask.inputDir.set(ziplineCompileTask.get().outputDir)
```
JakeWharton: You should never need to write `dependsOn` with property-based tasks.
```suggestion
      createdTask.inputDir.set(ziplineCompileTask.map { it.outputDir })
```
(This diff is just a guess, but it should be something like this)

JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
yissachar: Unfortunately not. The enum returns an uppercased string and captialized only ensures the first character is uppercase and doesn't care whether the rest of the string is upper or lower case.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
yissachar: Thanks - looking for uses of the existing webpack run points out that I've broken the samples with this PR and need to fix them.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
yissachar: Thanks! Apparently, I was using the Gradle version of capitalized which doesn't take a Locale, and the Kotlin one is deprecated, so I'll just use replaceFirstChar here.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
yissachar: Amazing - I was wondering why I was getting two builds every time I changed a file. Ended up coming up with a hack to pass the serve task a string instead of a Directory, but this is much better. Speed gains are even more now!

yissachar:
@swankjesse Let me know if you have any objections to me merging this!

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
JakeWharton: Yeah the Kotlin replacement is a bit pedantic. I hate how verbose it is.

swankjesse (approved):

PR #637, opened 2022-08-09
Use Spotless

### LINKED ISSUE #516: Use Spotless
I've noticed that some PRs are implementing one thing but then tidying up the formatting/adding license headers to seeming unrelated files (which [is awesome](https://wiki.c2.com/?BoyScoutRule)) but I think we can slightly reduce the cognitive load of the reviewers by enforcing that these things are fixed (as much as possible) at the time of implementation. We can use [Spotless](https://github.com/diffplug/spotless/tree/main/plugin-gradle) to fix the formatting via [ktlint](https://github.com/diffplug/spotless/tree/main/plugin-gradle#ktlint) and [enforce license headers](https://github.com/diffplug/spotless/tree/main/plugin-gradle#license-header).

veyndan:
Closes #516.


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Run './gradlew spotlessApply'
+2341806496ba2f23532a42775c4ef53b965e07c0
```
veyndan: **Important**: Don't squash or rebase when merging as that'll invalidate this rev.

build.gradle.kts
```diff
@@ -26,10 +27,24 @@ buildscript {
...
 apply(plugin = "com.vanniktech.maven.publish.base")
 
+configure<SpotlessExtension> {
+  kotlin {
+    target("**/*.kt")
+    ktlint()
+      .editorConfigOverride(
+        mapOf(
+          "indent_size" to "2",
+          "disabled_rules" to "filename, no-unused-imports, indent, comment-spacing",
```
veyndan: I just disabled these rules as they were causing some issues. They can be enabled in some follow up PRs.

samples/emoji-search/presenters/src/jvmMain/kotlin/app/cash/zipline/samples/emojisearch/EmojiSearchZipline.kt
```diff
@@ -27,6 +25,8 @@ import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
 import okhttp3.OkHttpClient
+import java.util.concurrent.Executors
+import kotlin.coroutines.EmptyCoroutineContext
```
JakeWharton: This import order is incorrect.

Please add `'ij_kotlin_imports_layout': '*',` to your override map to ensure there is no special grouping.

build.gradle.kts
```diff
@@ -26,10 +27,24 @@ buildscript {
...
+  alias(libs.plugins.spotless)
 }
 
 apply(plugin = "com.vanniktech.maven.publish.base")
 
+configure<SpotlessExtension> {
+  kotlin {
+    target("**/*.kt")
+    ktlint()
+      .editorConfigOverride(
```
JakeWharton, line 49: The fact the Spotless doesn't read editorconfig is ridiculous.

I have started switching projects to kotlinter which ensures you can centralize all configuration in editorconfig and it will be shared by the IDE and autoformatter without special configuration. It's worth considering here as well. Here's an example: https://github.com/cashapp/molecule/commit/aee2271c8f287658c818bbd5e0f7d5520a5c620d

JakeWharton (approved):

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Run './gradlew spotlessApply'
```
swankjesse, line 1: NEAT

swankjesse (approved):

build.gradle.kts
swankjesse: Yeah, we should just paste molecule’s `.editorconfig` here.

veyndan: Copy-pasted `.editorconfig` from molecule and converted the whole thing over to kotlinter. It's a bit sad to give up on Spotless as it's a more generic formatting tool over kotlinter, but a centralized `.editorconfig` is a _huge_ plus over (the possibility of) generic formatting using a single tool.

samples/emoji-search/presenters/src/jvmMain/kotlin/app/cash/zipline/samples/emojisearch/EmojiSearchZipline.kt
veyndan: Done!

JakeWharton (approved):

JakeWharton:
@veyndan you want me to help push this over the line?

veyndan:
That'd be super helpful. I've been meaning to get to this but I realistically won't have time in the next week or so. Judging by the error it looks like the `build` folder is being linted as well, so just need to somehow ignore the folder.

veyndan:
@JakeWharton What are your opinions on getting this merged? Going through https://pinterest.github.io/ktlint/install/integrations/#gradle-integration, it looks like they all have some sort of problem:

* autostyle doesn't support `.editorconfig` (https://github.com/autostyle/autostyle/issues/28)
* kotlinter can't trivially ignore build folders (https://github.com/jeremymailen/kotlinter-gradle/issues/208)
* Spotless doesn't support `.editorconfig` (https://github.com/diffplug/spotless/issues/142)

I'm thinking we just revert back to Spotless (as we use it elsewhere) and then just migrate to another tool once any of the above issues are fixed.

build.gradle.kts
veyndan: This concern is obsolete with [Spotless Gradle Plugin 6.13.0](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#added). Reverting to use Spotless over Kotlinter.

PR #641, opened 2022-08-10
Tools and documentation to generate EdDSA keys

swankjesse:
It's also possible to use the Tink API directly. Using the Tink
CLI isn't great; those keys are protos inside JSON which makes
them awkward to deal with.

### DIFF
--- a/zipline-cli/src/main/kotlin/app/cash/zipline/cli/GenerateKeyPair.kt
+++ b/zipline-cli/src/main/kotlin/app/cash/zipline/cli/GenerateKeyPair.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2022 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package app.cash.zipline.cli
+
+import app.cash.zipline.cli.GenerateKeyPair.Companion.NAME
+import app.cash.zipline.loader.internal.generateKeyPair
+import java.io.PrintStream
+import picocli.CommandLine.Command
+
+@Command(
+  name = NAME,
+  description = [
+    "Generate an Ed25519 key pair for ManifestSigner and ManifestVerifier"
+  ],
+  mixinStandardHelpOptions = true,
+  versionProvider = Main.VersionProvider::class,
+)
+class GenerateKeyPair : Runnable {
+  override fun run() {
+    run(System.out)
+  }
+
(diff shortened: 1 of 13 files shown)


adrw (approved):
LG

zipline-cli/src/main/kotlin/app/cash/zipline/cli/GenerateKeyPair.kt
```diff
@@ -0,0 +1,46 @@
...
+  name = NAME,
+  description = [
+    "Generate an Ed25519 key pair for ManifestSigner and ManifestVerifier"
+  ],
+  mixinStandardHelpOptions = true,
+  versionProvider = Main.VersionProvider::class,
+)
+class GenerateKeyPair : Runnable {
+  override fun run() {
+    run(System.out)
```
JakeWharton: Clikt lets you do constructor DI. Just sayyyiiinnn.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -95,4 +99,21 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
       listOf() // No options.
     }
   }
+
+  private fun createGenerateKeyPairTask(project: Project) {
+    val name = "generateZiplineManifestKeyPair"
+    val rootProjectTasks = project.rootProject.tasks
+    if (rootProjectTasks.findByName(name) == null) {
+      rootProjectTasks.create(name) { task ->
```
JakeWharton: Plugins can only be applied once. There's no need to do this conditional create.

Also `create` should be `register` in order to defer the task configuration when not invoked.

JakeWharton (approved):

zipline-cli/src/main/kotlin/app/cash/zipline/cli/GenerateKeyPair.kt
swankjesse: Oooh nice. Clikt is so good.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: I’m cheating here and adding a task to the root project even if the plugin is installed on a subproject. Does that change this advice?

JakeWharton: Ah. Yes and no!

So yes because then doing it lazy is required. But no, because project isolation (a forthcoming Gradle feature) will actively prevent you from doing this.

If you run `gw taskName` and the task only exists in `:foo:bar:baz:taskName` it will still run correctly. So it should be safe to add the task only to the project to which the plugin is applied.

swankjesse: Ahhh, I’ll define it in the right place.

zipline-cli/src/main/kotlin/app/cash/zipline/cli/GenerateKeyPair.kt
swankjesse: I defined a constructor parameter with a default value and it worked nicely.

PR #642, opened 2022-08-10
Move samples to their own project

yissachar:
Makes samples dir its own Gradle project. Removes Octane sample since it's not actually being used. 

Note that the samples are actually broken in two ways right now but this is independent of this PR:
- jsBrowserProductionRun doesn't actually work, only jsBrowserDevelopmentRun does. This is caused by [my build mode optimization](https://github.com/cashapp/zipline/blob/trunk/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt#L77) since the Webpack and Kotlin modes are not actually lining up. This will be fixed with my follow up http4k PR which removes our reliance on Webpack entirely.
- Running the emoji sample app has the following stacktrace: `Caused by: java.lang.ClassNotFoundException: Didn't find class "app.cash.zipline.loader.LoaderJvmKt" on path` - I don't know what's going on and again, this is present on trunk as well so I don't want to explode the scope of this PR but if someone knows what's happening we should fix this in a separate PR.


JakeWharton (approved):
Awesome!

samples/emoji-search/README.md
```diff
@@ -14,7 +14,7 @@ Serving presenters.js
 Run this:
 
 ```
-./gradlew samples:emoji-search:presenters:jsBrowserProductionRun --info
+../gradlew emoji-search:presenters:jsBrowserProductionRun --info
```
adrw: I'd lean towards these commands being always from root directory or explicitly mention in the README these commands are relative to the samples directory.

Issue #644, opened 2022-08-11
Nicer syntax for Gradle plugin configuration


swankjesse:
Today:

```
tasks.withType(ZiplineCompileTask::class) {
  mainFunction.set("com.squareup.cash.treehouse.playground.prepareTreehouse")
}
```

Possible alternative:

```
zipline {
  mainFunction.set("com.squareup.cash.treehouse.playground.prepareTreehouse")
}
```

adrw:
Probably lean on examples like Wire https://github.com/square/wire/blob/master/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt

JakeWharton:
https://docs.gradle.org/current/userguide/custom_plugins.html#sec:getting_input_from_the_build has a bit of a more modern example. Basically just an abstract type and `Property`-based properties.

PR #665, opened 2022-08-18
Move task scalar configuration to extension

JakeWharton:
Propagating the signing keys is a bit more complicated and will be done separately.

Refs #644


samples/emoji-search/presenters/build.gradle.kts
```diff
@@ -58,6 +57,6 @@ android {
   namespace = "app.cash.zipline.samples.emojisearch.presenters"
 }
 
-tasks.withType(ZiplineCompileTask::class) {
+zipline {
```
swankjesse, line 60: YAY

swankjesse (approved):

adrw:
So nice! 🎉

PR #669, opened 2022-08-18
Don't return non-200 responses

swankjesse:


zipline-loader/build.gradle.kts
```diff
@@ -48,6 +48,7 @@ kotlin {
       dependsOn(engineMain)
       dependencies {
         implementation(libs.okHttp.core)
+        implementation(libs.okHttp.mockWebServer)
```
JakeWharton: This should be in jniTest

zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/OkHttpZiplineHttpClientTest.kt
```diff
@@ -0,0 +1,70 @@
+/*
```
JakeWharton: Could move to jniTest to run on both JVM and Android

JakeWharton (approved):

zipline-loader/build.gradle.kts
swankjesse: ooooh good catch. Definitely that was my intent

zipline-loader/src/jvmTest/kotlin/app/cash/zipline/loader/OkHttpZiplineHttpClientTest.kt
swankjesse: Fixed.

PR #671, opened 2022-08-18
Fetch modules in parallel

yissachar:
This is all @swankjesse, it just happened to occur on my machine.

Fetches are now actually occurring in parallel, though I don't see any performance difference.

### DIFF
--- a/zipline-loader/src/androidMain/kotlin/app/cash/zipline/loader/internal/internalAndroid.kt
+++ b/zipline-loader/src/androidMain/kotlin/app/cash/zipline/loader/internal/internalAndroid.kt
@@ -16,6 +16,11 @@
 package app.cash.zipline.loader.internal
 
 import app.cash.zipline.Zipline
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.Dispatchers
 
-internal actual fun Zipline.multiplatformLoadJsModule(bytecode: ByteArray, id: String) =
+internal actual fun Zipline.multiplatformLoadJsModule(bytecode: ByteArray, id: String) {
   loadJsModule(bytecode, id)
+}
+
+actual val ioDispatcher: CoroutineContext = Dispatchers.IO
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -40,6 +40,7 @@ import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.joinAll
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.sync.Semaphore
 import kotlinx.coroutines.sync.withPermit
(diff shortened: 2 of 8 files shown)


zipline-loader/src/nativeMain/kotlin/app/cash/zipline/loader/internal/internalNative.kt
```diff
@@ -16,6 +16,10 @@
 package app.cash.zipline.loader.internal
 
 import app.cash.zipline.Zipline
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
 
 internal actual fun Zipline.multiplatformLoadJsModule(bytecode: ByteArray, id: String) =
   loadJsModule(bytecode, id)
+
+actual val ioDispatcher: CoroutineContext = EmptyCoroutineContext
```
JakeWharton, line 25:
```suggestion
actual val ioDispatcher: CoroutineContext = Dispatchers.Default
```
Native should do work off the main thread too.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -109,7 +110,10 @@ class ZiplineLoader internal constructor(
       maxSizeInBytes = maxSizeInBytes,
       nowEpochMs = nowEpochMs,
     )
-    cache.initialize()
+    // TODO: make ZiplineLoader.withCache() defer initialization or launch it
+    runBlocking {
+      cache.initialize()
+    }
```
JakeWharton, line 116: Yeah this will prevent the strict mode crash, but it does so by tricking strict mode by blocking on the same work moved but moved to a background thread.

I would launch this in an `async` and then save the `Deferred` such that whenever you need to access the cache you call `await()` which will either return synchronously if the initialization finished or it will suspend waiting for it to finish.

zipline-loader/src/androidMain/kotlin/app/cash/zipline/loader/internal/internalAndroid.kt
```diff
@@ -16,6 +16,11 @@
...
 import app.cash.zipline.Zipline
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.Dispatchers
 
-internal actual fun Zipline.multiplatformLoadJsModule(bytecode: ByteArray, id: String) =
+internal actual fun Zipline.multiplatformLoadJsModule(bytecode: ByteArray, id: String) {
   loadJsModule(bytecode, id)
+}
+
+actual val ioDispatcher: CoroutineContext = Dispatchers.IO
```
JakeWharton, line 26: I am _extremely_ apprehensive about having static dispatchers encoded rather than using inversion of control.

JakeWharton (approved):
There is no actual parallelism here. That would require some calls to `launch` or `async`. Everything still runs linearly but now adds thread hops so as to not do file system work on the caller thread. It should technically be every-so-slightly slower than before, although it's so small that it should be lost in measurement noise.

swankjesse:
Lemme take over this PR to add the parallelism & IOC!

swankjesse:
We went another way with this.

PR #674, opened 2022-08-19
Remove explicit opt-in of RequiresOptIn

veyndan:
https://kotlinlang.org/docs/whatsnew17.html#stable-opt-in-requirements


JakeWharton (approved):
omg yes. so many projects to update.

PR #675, opened 2022-08-19
Add support for ECDSA keys

swankjesse:
This is not implemented on native yet, that's for follow-up.


JakeWharton (approved):

zipline-cli/src/main/kotlin/app/cash/zipline/cli/GenerateKeyPair.kt
```diff
@@ -32,8 +34,15 @@ import picocli.CommandLine.Command
...
+    description = ["Signing algorithm to use."],
+  )
+  var algorithm: SignatureAlgorithmId = SignatureAlgorithmId.Ed25519
+
   override fun run() {
-    val keyPair = generateKeyPair()
+    val keyPair = algorithm.generateKeyPair()
+    out.println("  ALGORITHM: $algorithm")
     out.println(" PUBLIC KEY: ${keyPair.publicKey.hex()}")
     out.println("PRIVATE KEY: ${keyPair.privateKey.hex()}")
```
nmahendru, line 47: We should probably add an option to write this to a file instead.
```
names = [ "-p",  "--private-key"]
```

Printing out the private key that someone might end up using is not the best default option!
Or is it clearly communicated otherwise to wipe out shell history or reboot the machine or something like that ?

PR #683, opened 2022-08-20
Require callers to explicitly opt out of signature checks

swankjesse:
Previously the ManifestVerifier was nullable and had a default
value of null. Code that had signature checks disabled didn't
look insecure.

With this change callers must explicitly opt-out of signature
checks by passing ManifestVerifier.NO_SIGNATURE_CHECKS.


JakeWharton (approved):

nmahendru:
looks good! 🚀

Issue #690, opened 2022-08-21
Zipline 0.9.1 doesn’t work on Apple Silicon


swankjesse:
Looks like we’re shipping an x86_64 artifact where the aarch64 goes.

```
$ unzip zipline-jvm-0.9.1.jar

$ file jni/amd64/libquickjs.so
jni/amd64/libquickjs.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=5e8eaeba18cd1c4d4fe46a1b9a27796bd9bc16b3, not stripped

 $ file jni/x86_64/libquickjs.dylib
jni/x86_64/libquickjs.dylib: Mach-O 64-bit dynamically linked shared library x86_64

$ file jni/aarch64/libquickjs.dylib
jni/aarch64/libquickjs.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```

adrw:
Maybe a bug here: https://github.com/cashapp/zipline/blob/trunk/.github/workflows/build-mac.sh

Do we use that script for releases or something else?

swankjesse:
Fixed.

PR #691, opened 2022-08-21
Build for Apple Silicon in Github Actions

swankjesse:


.github/workflows/release.yaml
```diff
@@ -48,14 +48,7 @@ jobs:
...
         if: matrix.os == 'macOS-latest'
-        run: |
-          mkdir -p build/jni/${{ matrix.arch }}/
-          mkdir -p zipline/src/jvmMain/resources/jni/${{ matrix.arch }}/
-          cmake -S zipline/src/jvmMain/ -B build/jni/${{ matrix.arch }}/ \
-            -DQUICKJS_VERSION="$(cat zipline/native/quickjs/VERSION)" \
-            -DCMAKE_SYSTEM_PROCESSOR=${{ matrix.cmake-arch }}
-          cmake --build build/jni/${{ matrix.arch }}/ --verbose
-          cp -v build/jni/${{ matrix.arch }}/libquickjs.* zipline/src/jvmMain/resources/jni/${{ matrix.arch }}/
+        run: ./.github/workflows/build-mac.sh -a ${{ matrix.arch }} -c ${{ matrix.cmake-arch }}
```
swankjesse, line 51: This is consistent with `build.yaml`.

.github/workflows/build-mac.sh
```diff
@@ -56,7 +56,7 @@ function build() {
   echoRun mkdir -p zipline/src/jvmMain/resources/jni/$MACOS_ARCH/
   echoRun cmake -S zipline/src/jvmMain/ -B build/jni/$MACOS_ARCH/ \
     -DQUICKJS_VERSION="$(cat zipline/native/quickjs/VERSION)" \
-    -DCMAKE_SYSTEM_PROCESSOR=$CMAKE_ARCH
+    -DCMAKE_OSX_ARCHITECTURES=$CMAKE_ARCH
```
swankjesse, line 59: I determined experimentally that this is the variable I want to change.
https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html
https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html

.github/workflows/release.yaml
adrw: Whoops, I missed updating the release to match build

adrw (approved):

adrw:
Does this close #690?

swankjesse:
YES.

PR #700, opened 2022-08-23
Rename LoadedZipline to LoadAttempt and emit on failure

### LINKED ISSUE #693: Change LoadedZipline to LoadResult
LoadResult will be like `LoadedZipline`, but it’ll add a new field, `Exception` if the load failed.

(Maybe it should use the `Result` type?)

adrw:
[CLOSES #693]

- [x] Update to LoadResult name instead
- [x] Use Sealed classes and pass back exception

### DIFF
--- a/samples/world-clock/presenters/src/hostMain/kotlin/app/cash/zipline/samples/worldclock/host.kt
+++ b/samples/world-clock/presenters/src/hostMain/kotlin/app/cash/zipline/samples/worldclock/host.kt
@@ -16,7 +16,7 @@
 package app.cash.zipline.samples.worldclock
 
 import app.cash.zipline.Zipline
-import app.cash.zipline.loader.LoadedZipline
+import app.cash.zipline.loader.LoadResult
 import app.cash.zipline.loader.ZiplineLoader
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
@@ -39,7 +39,7 @@ fun startWorldClockZipline(
   models: MutableStateFlow<WorldClockModel>,
 ) {
   scope.launch(ziplineDispatcher + SupervisorJob()) {
-    val ziplineFlow: Flow<LoadedZipline> = ziplineLoader.load(
+    val loadResultFlow: Flow<LoadResult> = ziplineLoader.load(
       applicationName = "world-clock",
       manifestUrlFlow = repeatFlow(manifestUrl, 500L),
       initializer = { zipline: Zipline ->
@@ -49,23 +49,25 @@ fun startWorldClockZipline(
 
     var previousJob: Job? = null
 
-    ziplineFlow.collect { loadedZipline ->
+    loadResultFlow.collect { result ->
       previousJob?.cancel()
 
-      val zipline = loadedZipline.zipline
-      val presenter = zipline.take<WorldClockPresenter>("WorldClockPresenter")
+      if (result is LoadResult.Success) {
+        val zipline = result.zipline
+        val presenter = zipline.take<WorldClockPresenter>("WorldClockPresenter")
 
-      val job = launch {
-        models.emitAll(presenter.models(events))
+        val job = launch {
+          models.emitAll(presenter.models(events))
+        }
+
+        job.invokeOnCompletion {
+          presenter.close()
+          // TODO(jwilson): make this safe.
+          // zipline.close()
+        }
+
+        previousJob = job
       }
-
-      job.invokeOnCompletion {
-        presenter.close()
-        // TODO(jwilson): make this safe.
-        // zipline.close()
-      }
-
-      previousJob = job
     }
   }
 }
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/LoadResult.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/LoadResult.kt
@@ -18,9 +18,21 @@ package app.cash.zipline.loader
 import app.cash.zipline.Zipline
 
 /**
- * Zipline engine loaded with code and related metadata on the load attempt.
+ * Final result state from a ZiplineLoader attempt.
  */
-data class LoadedZipline(
-  val zipline: Zipline,
-  val freshAtEpochMs: Long,
-)
+sealed class LoadResult {
+  /**
+   * [Zipline] engine successfully loaded with code and corresponding metadata.
+   */
+  data class Success(
+    val zipline: Zipline,
+    val freshAtEpochMs: Long,
+  ): LoadResult()
+
+  /**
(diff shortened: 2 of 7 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -186,9 +186,10 @@ class ZiplineLoader internal constructor(
           try {
             val networkZipline = loadFromManifest(applicationName, networkManifest, initializer)
             cachingFetcher?.pin(applicationName, networkManifest) // Pin after successful loads.
-            emit(LoadedZipline(networkZipline, networkManifest.freshAtEpochMs))
+            emit(LoadAttempt(LoadStatus.SUCCESS, networkZipline, networkManifest.freshAtEpochMs))
             previousManifest = networkManifest.manifest
           } catch (e: Exception) {
+            emit(LoadAttempt(LoadStatus.FAILURE, null, networkManifest.freshAtEpochMs))
             cachingFetcher?.unpin(applicationName, networkManifest) // Unpin after failed loads.
             throw e
```
adrw, line 192: @swankjesse should we still be throwing here if we're now emitting on failure?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/LoadResult.kt
```diff
@@ -18,9 +18,22 @@ package app.cash.zipline.loader
...
-  val zipline: Zipline,
-  val freshAtEpochMs: Long,
-)
+sealed class LoadResult {
+  /**
+   * [Zipline] engine successfully loaded with code and corresponding metadata.
+   */
+  data class Success(
+    val zipline: Zipline,
+    val freshAtEpochMs: Long,
```
JakeWharton, line 29: Not sure what this is used for, but could be an `abstract val` on the `LoadResult` type if we think there's code that would unconditionally want to access it.

JakeWharton (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/LoadResult.kt
swankjesse: Maybe even simpler - omit the `freshAtEpochMs` for failure, since that’s not particularly useful? The main utility of this thing is to know when you’re reading stale code from the cache, but it’s not relevant that a failure is fresh or stale.

swankjesse (approved):

samples/world-clock/presenters/src/hostMain/kotlin/app/cash/zipline/samples/worldclock/host.kt
```diff
@@ -49,23 +49,25 @@ fun startWorldClockZipline(
 
     var previousJob: Job? = null
 
-    ziplineFlow.collect { loadedZipline ->
+    loadResultFlow.collect { result ->
       previousJob?.cancel()
 
-      val zipline = loadedZipline.zipline
-      val presenter = zipline.take<WorldClockPresenter>("WorldClockPresenter")
+      if (result is LoadResult.Success) {
```
adrw, line 55: @swankjesse @JakeWharton is this the correct migration for this callsite?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
adrw: Ended up discovering we still need to throw so the event listener wrapping method can catch the exception and emit the corresponding failure event.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -152,9 +155,9 @@ class ZiplineLoader internal constructor(
     applicationName: String,
     manifestUrlFlow: Flow<String>,
     serializersModule: SerializersModule = EmptySerializersModule(),
-    initializer: (Zipline) -> Unit,
-  ): Flow<LoadedZipline> {
-    return flow {
+    initializer: (Zipline) -> Unit = {},
+  ): Flow<LoadResult> {
+    return channelFlow {
```
adrw, line 160: To support throwing on failure for the event listener to catch, I needed to change this to a `channelFlow`. Are there any downstream consequences that would make this ill advised?

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/ZiplineLoaderSigningTest.kt
```diff
@@ -98,9 +99,8 @@ class ZiplineLoaderSigningTest {
...
       bravoUrl to testFixtures.alphaByteString,
     )
-    assertFailsWith<IllegalStateException> {
-      tester.loader.loadOnce("test", manifestUrl)
-    }
+    val result = tester.loader.loadOnce("test", manifestUrl)
+    assertTrue(result is LoadResult.Failure)
     assertEquals(
       "checksum mismatch for bravo",
       eventListener.takeException().message,
```
JakeWharton: This can use the exception from the result. Also we should still test the type is an `IllegalStateException`.

samples/world-clock/presenters/src/hostMain/kotlin/app/cash/zipline/samples/worldclock/host.kt
JakeWharton: Seems fine. Only other thing you could do would be to log.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
JakeWharton: `channelFlow` is needed when a coroutine other than the one which runs the lambda passed to the `flow { }` factory needs to send data. I don't really see why this change would force using `channelFlow` as the usage of coroutines hasn't changed as far as I can tell.

To answer your question, it won't really affect anyone downstream other than requiring one more dispatch on the event loop before the outgoing event actually is sent downstream.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
JakeWharton: It's certainly not an amazing pattern we got forced into here. Generally you want to handle or rethrow and never both. Maybe the design of `withLifecycleEvents` needs to change a bit in a future change to try and remedy this.

JakeWharton (approved):

adrw:
Going to land this as is and then address channelFlow and #724 in a follow up PR

PR #701, opened 2022-08-23
Replace `map` with `zip` to retain provider chain

JakeWharton:
Otherwise we lose potential dependencies of the `algorithmId` by calling `get()`.

We also _might_ be able to just make these objects not using properties as one could probably provide into the container, but I'm less sure about that...


adrw (approved):

swankjesse:
I was trying to figure out how to express that. This works, but wow it really is a puzzle.

JakeWharton:
I need to research if we even need properties here. I think once you're nested in a container you can push this to the caller.

PR #702, opened 2022-08-23
Make ZiplineCache a top-level type, again

swankjesse:
In integrating this into another project I learned that
some callers will want to use different ZiplineLoader
configurations with the same ZiplineCache instance.

This promotes ZiplineCache to a top-level type that can
be created and closed independently of the ZiplineLoader
instances that use it.

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
@@ -13,9 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package app.cash.zipline.loader.internal.cache
+package app.cash.zipline.loader
 
+import app.cash.zipline.loader.internal.cache.Database
+import app.cash.zipline.loader.internal.cache.FileState
+import app.cash.zipline.loader.internal.cache.Files
+import app.cash.zipline.loader.internal.cache.SqlDriverFactory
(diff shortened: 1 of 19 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
```diff
@@ -170,35 +198,40 @@ internal class ZiplineCache internal constructor(
...
   }
 
-  fun pin(applicationName: String, sha256: ByteString) {
+  internal fun pin(applicationName: String, sha256: ByteString) {
     val fileId = database.filesQueries.get(sha256.hex()).executeAsOneOrNull()?.id ?: return
     createPinIfNotExists(applicationName, fileId)
   }
 
-  fun unpin(applicationName: String, sha256: ByteString) {
+  internal fun unpin(applicationName: String, sha256: ByteString) {
```
adrw, line 206: Nice move to make the class public but all usage internal

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -89,29 +98,9 @@ class ZiplineLoader internal constructor(
   )
 
   fun withCache(
-    fileSystem: FileSystem,
-    directory: Path,
-    maxSizeInBytes: Long,
+    cache: ZiplineCache,
```
adrw, line 101: sweet

adrw (approved):

PR #705, opened 2022-08-23
Add application load report generator

yissachar:
Adds an event listener that will generate an application load report and can be used for performance profiling. I've added a bunch of new events to support this - I'm open to the idea that I'm polluting the interface :)

Assuming this approach is desirable follow-up PRs could enhance this further with more information, e.g. tracking manifest load timing, recording whether a module load resulted in an HTTP call or not, etc.

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -17,6 +17,7 @@ package app.cash.zipline.loader
 
 import app.cash.zipline.EventListener
 import app.cash.zipline.Zipline
+import app.cash.zipline.internal.systemEpochMsClock
 import app.cash.zipline.loader.internal.fetcher.FsCachingFetcher
 import app.cash.zipline.loader.internal.fetcher.FsEmbeddedFetcher
 import app.cash.zipline.loader.internal.fetcher.HttpFetcher
@@ -26,7 +27,6 @@ import app.cash.zipline.loader.internal.getApplicationManifestFileName
 import app.cash.zipline.loader.internal.receiver.FsSaveReceiver
 import app.cash.zipline.loader.internal.receiver.Receiver
 import app.cash.zipline.loader.internal.receiver.ZiplineLoadReceiver
-import app.cash.zipline.loader.internal.systemEpochMsClock
 import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Job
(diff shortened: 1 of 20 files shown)


yissachar:
Looks like https://plugins.gradle.org/ is down at the moment, causing the build to fail

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -333,14 +334,23 @@ class ZiplineLoader internal constructor(
         nowEpochMs = nowEpochMs,
         baseUrl = baseUrl,
         url = module.url,
+        eventListener,
       )!!
       check(byteString.sha256() == module.sha256) {
         "checksum mismatch for $id"
       }
+
+      eventListener.moduleUpstreamFetchStart(applicationName, id)
```
swankjesse, line 343: Mind adding a `startValue` to these? I added that since you last pulled!

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -325,6 +325,7 @@ class ZiplineLoader internal constructor(
      * Fetch and receive ZiplineFile module.
      */
     suspend fun run() {
+      eventListener.moduleLoadStart(applicationName, id)
```
swankjesse, line 328: I suspect this is going to be difficult to act upon well, cause users are going to measure this start/end pair but they really aught to measure download time + receive time.

What would you think about putting the module name into the downloadStart task, and just measuring modules as two independent steps?

(Also that leaks our implementation details less!)

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/fetcher/Fetcher.kt
```diff
@@ -50,24 +51,41 @@ internal suspend fun List<Fetcher>.fetch(
...
+  eventListener: EventListener
+): ByteString? {
+  eventListener.moduleFetchPermitAcquireStart(applicationName, id)
+  return concurrentDownloadsSemaphore.withPermit {
+    eventListener.moduleFetchPermitAcquireEnd(applicationName, id)
+    var firstException: Exception? = null
+    var failedFetcherName = "unknown"
+    for (fetcher in this) {
+      try {
+        eventListener.moduleFetchStart(applicationName, id, fetcher::class.simpleName ?: "unknown")
```
swankjesse, line 63: Should we dedupe this with downloads?

swankjesse (commented):
Let’s get together and design this a bit more? I wanna try to thread the needle and make EventListeners useful for stuff like this, but not expose exactly how our internals work.

yissachar:
Closing this out after discussing with Jesse. This is conflating two different use cases, our internal need to profile Zipline performance and our customer's needs to profile their specific app load times, and change the things they have under their control. 

I'll take a different approach to profiling internal Zipline (maybe just sticking to JVM and manual printing) and possibly put up a separate PR with just the module receive events for the listener.

PR #711, opened 2022-08-26
Endpoint-layer support for shutdown.

swankjesse:
This is mostly just book-keeping of how many outbound services
are present. When the last of them is closed then the overall
endpoint can be closed.

This treats closing as a mechanism that is local to a single
endpoint. In practice this works for our use case, as only the
host application is interested in ever shutting down the QuickJS
VM, it is not interested in shutting itself down.

Note that this keeps counts of client-side outbound interfaces
instead of references to those interfaces directly. That would
be simpler but it would interfere with leak detection as the
Endpoint would always hold a strong reference.


JakeWharton (approved):
What do you think about external visibility of shutdown progress? I'm thinking of the case where you want to shutdown gracefully for N seconds and then forcibly break things.

swankjesse:
Yeah good call out. I think we let you call `shutdown()` or `close()` when you like, and have `EventListener` signal each.

JakeWharton:
Those things are usually far away from each other. Compose exposes a StateFlow so that you can shutdown and then call `.first { it == Shutdown }`. We also could just return a `Deferred<Unit>` and basically do that internally.

zipline/src/engineTest/kotlin/app/cash/zipline/ShutdownTest.kt
```diff
@@ -0,0 +1,225 @@
...
+    endpointB.bind<SuspendingEchoService>("service", service)
+    endpointA.shutdown()
+    assertEquals(endpointA.state, Endpoint.State.SHUTTING_DOWN)
+
+    client.suspendingEcho(EchoRequest("hello"))
+    assertEquals(endpointA.state, Endpoint.State.CLOSED) // Closed when it receives the callback.
+  }
+
+  @Test
+  fun throwingFromOutboundSuspendingFunctionTriggersEndpointClose() = runBlocking {
```
adrw, line 145: great tests

adrw (approved):

swankjesse:
(I’m sitting on this 'cause I’m not satisfied with the cleanup process. I think maybe ZiplineScope might make it better.)

swankjesse:
I ended up focusing on scopes to do reliable shutdown. I think that might be all I need.

PR #712, opened 2022-08-27
Change our wrapper exception to ZiplineException

### LINKED ISSUE #676: Default exception type should be ZiplineException
Instead of `Exception`, which is quite ambiguous.

swankjesse:
This should make it easier to catch exceptions thrown
across the ZiplineService boundary.

Closes: https://github.com/cashapp/zipline/issues/676

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineException.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineException.kt
@@ -23,9 +23,6 @@ package app.cash.zipline
  * process.
  */
 class ZiplineException(
-  message: String?,
-  cause: Throwable?,
-) : RuntimeException(message, cause) {
-  constructor(message: String?) : this(message, null)
-  constructor(cause: Throwable?) : this(null, cause)
-}
+  message: String? = null,
+  cause: Throwable? = null,
+) : RuntimeException(message, cause)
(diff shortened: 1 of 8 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineException.kt
```diff
@@ -0,0 +1,31 @@
...
+ * This is similar to a wrapping exception like Java's `InvocationTargetException`, but the wrapped
+ * exception type is not generally available because its class might not exist in the catching
+ * process.
+ */
+class ZiplineException(
+  message: String?,
+  cause: Throwable?,
+) : RuntimeException(message, cause) {
+  constructor(message: String?) : this(message, null)
+  constructor(cause: Throwable?) : this(null, cause)
```
JakeWharton: Why this and not primary + default argument values? JNI would have to call the two-arg one with explicit nulls, but you get a nicer Kotlin API.

Also anything accessed through JNI needs listed in the shrinker rules.

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineException.kt
swankjesse: Just following the old Java conventions like a chump. Lemme do defaults.

PR #719, opened 2022-08-31
Add local.properties instructions and template

adrw:
For onboarding on a new computer, these instructions and template will make it more intuitive for new contributors.

### DIFF
--- a/local.properties.template
+++ b/local.properties.template
@@ -1,0 +1,1 @@
+sdk.dir=/Users/{your account username}/Library/Android/sdk
--- a/local.properties.template
+++ b/local.properties.template
@@ -1,1 +1,0 @@
-sdk.dir=/Users/{your account username}/Library/Android/sdk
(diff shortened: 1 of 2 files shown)


oldergod (approved):

local.properties.template
```diff
@@ -0,0 +1 @@
+sdk.dir=/Users/{your account username}/Library/Android/sdk
```
oldergod: Is that a common pattern? Otherwise I'd just not include this file

oldergod (approved):

local.properties.template
JakeWharton: AS creates this file the first time you import the project in the IDE. I believe AGP also creates it automatically when you do a CLI build based on your ANDROID_HOME env var. I've never had to create this file myself. Are we sure the instructions aren't to export the env var instead?

adrw: I primarily open Zipline in IntelliJ not Android Studio, and only used Android Studio to get the local SDK downloaded and configured. I don't have a ton of mobile experience so happy to defer to whatever the recommended local setup instructions would be. I'll go ahead and remove this file.

Issue #646, opened 2022-08-11
Try to turn off / avoid using `eval()` as much as possible


swankjesse:
It’s a potential sandbox-escape if we end up running unexpected code.

swankjesse:
Two different projects.

1. Avoid eval() in our libraries
2. Disable eval() in our runtime

swankjesse:
For 1, I can’t find any calls to eval() in our generated JS. Maybe it was eliminated when we upgraded to Kotlin 1.7?

JakeWharton:
It used to be used in development webpack but I believe we turned it off. Should never be used in production Kotlin unless you opt-in somehow or call it explicitly.

swankjesse:
Yep, there’s zero use of eval in our existing Kotlin/JS compiler output.

swankjesse:
I tried disabling eval. Disabling it is relatively straightforward – use `JS_NewContextRaw` and introduce only the intrinsics we want:

```
/**
 * Like JS_NewContext, but without an intrinsic for eval.
 */
JSContext* jsNewContextNoEval(JSRuntime* rt) {
  JSContext *ctx;

  ctx = JS_NewContextRaw(rt);
  if (!ctx) return NULL;

  JS_AddIntrinsicBaseObjects(ctx);
  JS_AddIntrinsicDate(ctx);
  // JS_AddIntrinsicEval(ctx); // TODO(jwilson): remove
  JS_AddIntrinsicStringNormalize(ctx);
  JS_AddIntrinsicRegExp(ctx);
  JS_AddIntrinsicJSON(ctx);
  JS_AddIntrinsicProxy(ctx);
  JS_AddIntrinsicMapSet(ctx);
  JS_AddIntrinsicTypedArrays(ctx);
  JS_AddIntrinsicPromise(ctx);
  // JS_AddIntrinsicBigInt(ctx);
  return ctx;
}
```

Unfortunately actually doing this is currently blocked by a few things:
 * `installFinalizationRegistry` uses `JS_Eval`. We could work around this by compiling first on a separate `JSContext`.
 * QuickJs.compile() uses `JS_Eval`. We could again use a separate `JSContext` for this.
 * ZiplineLoader.loadFromManifest() uses `JS_Eval`. We could precompile a function that takes a `mainModuleId` and `mainFunction`.

swankjesse:
One thing that might work – change `QuickJs.evaluate()` to always work in two steps:
1. compile on an eval-carrying `JSContext`
[...]

PR #730, opened 2022-09-15
[#646] Remove evaluate usages within non-QuickJs.kt main files

adrw:
This continues work to limit and eventually remove `evaluate` as a public API offered on our QuickJs engine.

### DIFF
--- a/zipline/src/engineMain/kotlin/app/cash/zipline/QuickJs.kt
+++ b/zipline/src/engineMain/kotlin/app/cash/zipline/QuickJs.kt
@@ -81,6 +81,20 @@ expect class QuickJs {
 
   internal fun getInboundChannel(): CallChannel
 
+  fun collectModuleDependencies()
+
+  internal fun initModuleLoader()
+
+  fun getGlobalThis(key: String): String?
+
+  fun getModuleDependencies(): List<String>
+
+  internal fun loadJsModule(script: String, id: String)
+
+  internal fun loadJsModule(id: String, bytecode: ByteArray)
+
+  fun runApplication(mainModuleId: String, mainFunction: String)
+
   /**
    * Manually invoke cycle removal. This is intended for testing only and is never necessary to
    * call in regular execution.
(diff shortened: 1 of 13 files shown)


zipline/src/jniMain/kotlin/app/cash/zipline/QuickJs.kt
```diff
@@ -124,6 +130,47 @@ actual class QuickJs private constructor(
     return JniCallChannel(this, instance)
   }
 
+  actual fun collectModuleDependencies() {
```
adrw: @swankjesse do we want these types of functions in QuickJs.kt or another file?

zipline/src/engineMain/kotlin/app/cash/zipline/QuickJs.kt
```diff
@@ -81,6 +81,20 @@ expect class QuickJs {
 
   internal fun getInboundChannel(): CallChannel
 
+  fun collectModuleDependencies()
```
swankjesse: I’d like to find a way to make these non-public API . . . somehow

zipline/src/jniMain/kotlin/app/cash/zipline/QuickJs.kt
swankjesse: Lets put them in another file, that way we get two benefits:
 - that file can be in an `internal` package, even if it isn‘t tagged with `internal` visibility
 - we can put it in engineMain and share it between platforms!

swankjesse (approved):

zipline/src/jniMain/kotlin/app/cash/zipline/QuickJs.kt
```diff
@@ -15,11 +15,17 @@
  */
 package app.cash.zipline
 
+import app.cash.zipline.internal.COLLECT_DEPENDENCIES_DEFINE_JS
```
adrw: rm unused imports

zipline/src/nativeMain/kotlin/app/cash/zipline/QuickJs.kt
```diff
@@ -15,6 +15,10 @@
  */
 package app.cash.zipline
 
+import app.cash.zipline.internal.COLLECT_DEPENDENCIES_DEFINE_JS
```
adrw: rm unused imports

PR #732, opened 2022-09-17
WIP: Compile dependencies from node_modules

### LINKED ISSUE #725: Don't crash compiling modules with NPM JS dependencies
I'm getting the following build failure when trying to use kotlinx-datetime on the JS side of Zipline:

```text
Caused by: java.lang.IllegalArgumentException: No topological ordering is possible for [./88b0986a7186d029-atomicfu-js-ir.js, ./elevated-model.js, ./elevated-worker.js, ./kotlinx-serialization-kotlinx-serialization-json-js-ir.js, ./zipline-root-zipline.js, ./kotlinx.coroutines-kotlinx-coroutines-core-js-ir.js, ./kotlin-kotlinx-atomicfu-runtime-js-ir.js, ./Kotlin-DateTime-library-kotlinx-datetime-js-ir.js, ./kotlin-kotlin-stdlib-js-ir.js, ./kotlinx-serialization-kotlinx-serialization-core-js-ir.js]
	at app.cash.zipline.loader.internal.TopologicalSortKt.topologicalSort(topologicalSort.kt:61)
	at app.cash.zipline.loader.ZiplineManifest$Companion.create(ZiplineManifest.kt:153)
	at app.cash.zipline.loader.ZiplineManifest$Companion.create$default(ZiplineManifest.kt:143)
	at app.cash.zipline.gradle.ZiplineCompiler.writeManifest(ZiplineCompiler.kt:167)
	at app.cash.zipline.gradle.ZiplineCompiler.compile(ZiplineCompiler.kt:53)
	at app.cash.zipline.gradle.ZiplineCompileTask.task(ZiplineCompileTask.kt:101)
```

I ran Gradle in debug mode and the problem library seems to be `kotlinx-datetime`. This library depends on `@js-joda/core` which is not in the module map when creating the manifest. So when building the topological list, the dependencies of `kotlinx-datetime` are never satisfied resulting in the exception.

Is this a known limitation of Zipline? That you cannot use libraries like `kotlinx-datetime` because of such dependencies? I can create a small reproducing project if that is needed.

bnorm:
Resolves #725.

While this all seems to work, it still feels very much like a hack to me as I haven't explored a lot of edge cases. However, I wanted to open a draft PR early just to make sure this is the right approach or if there are other ideas. If you are comfortable with this approach I can start to clean things up and add some test cases.

I was able to get the trivia sample working with passing an `Instant` as a game parameter. However I switched to the world-clock sample for the PR since it was a more natural fit; it compiles but I haven't tested it for real yet.

### DIFF
--- a/samples/kotlin-js-store/yarn.lock
+++ b/samples/kotlin-js-store/yarn.lock
@@ -52,6 +52,11 @@
     "@jridgewell/resolve-uri" "^3.0.3"
     "@jridgewell/sourcemap-codec" "^1.4.10"
 
+"@js-joda/core@3.2.0":
+  version "3.2.0"
+  resolved "https://registry.yarnpkg.com/@js-joda/core/-/core-3.2.0.tgz#3e61e21b7b2b8a6be746df1335cf91d70db2a273"
+  integrity sha512-PMqgJ0sw5B7FKb2d5bWYIoxjri+QlW/Pys7+Rw82jSH0QN3rB05jZ/VrrsUdh1w4+i2kw9JOejXGq/KhDOX7Kg==
+
 "@leichtgewicht/ip-codec@^2.0.1":
   version "2.0.4"
   resolved "https://registry.yarnpkg.com/@leichtgewicht/ip-codec/-/ip-codec-2.0.4.tgz#b2ac626d6cb9c8718ab459166d4bb405b8ffa78b"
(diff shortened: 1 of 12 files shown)


JakeWharton:
Seems like a great first step. Do you want feedback on the individual TODOs?

bnorm:
> Seems like a great first step. Do you want feedback on the individual TODOs?

I would gladly take any feedback you have on TODOs! Quite a few are just notes to myself on what needs to be improved in the next round of changes so don't stress about it if you like the general direction this PR is going.

swankjesse (commented):
Yeah, this is really cool and something we certainly need. Please proceed !

swankjesse:
After some research & experimentation, I think we’re gonna try to support a WASM-based future for this project. That means anything we build to better support NPM & JS ecosystem is gonna have a limited lifespan.

Shabinder:
Is this still active ?
Can see the mention of shift to WASM, and short lifespan on this if ever included until Zipline 2.0 arrives.

Couldn't resist asking, if yes, any eta on that ?

PR #743, opened 2022-10-27
Add websocket for hot reload

yissachar:
Switches from Sun to Jetty server to add websocket serve capabilities. When a Kotlin file is changed a build will be triggered, executing this task. If the server hasn't been started, we will start it. If the server is up already, then we will instead send a websocket message to all connections letting them know that an update is available and they should reload the manifest.

Use of the websocket connection is entirely optional and will just provide a slightly faster hot reload experience. Callers can still poll the manifest (and should fall back to this if the websocket connection fails).

### DIFF
--- a/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
+++ b/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
@@ -16,7 +16,9 @@
 
 package app.cash.zipline.gradle
 
+import java.util.Timer
 import javax.inject.Inject
+import kotlin.concurrent.schedule
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.provider.Property
@@ -31,9 +33,13 @@ import org.http4k.routing.ResourceLoader.Companion.Directory
 import org.http4k.routing.bind
 import org.http4k.routing.routes
 import org.http4k.routing.static
+import org.http4k.routing.websockets
 import org.http4k.server.Http4kServer
-import org.http4k.server.SunHttp
+import org.http4k.server.Jetty
+import org.http4k.server.PolyHandler
 import org.http4k.server.asServer
+import org.http4k.websocket.Websocket
+import org.http4k.websocket.WsMessage
 
 abstract class ZiplineServeTask : DefaultTask() {
 
@@ -50,17 +56,43 @@ abstract class ZiplineServeTask : DefaultTask() {
     val deploymentRegistry = services.get(DeploymentRegistry::class.java)
     val deploymentHandle = deploymentRegistry.get(deploymentId, ZiplineServerDeploymentHandle::class.java)
     if (deploymentHandle == null) {
-      val server = routes(
-        "/" bind static(Directory(inputDir.get().asFile.absolutePath), Pair("zipline", ContentType.TEXT_PLAIN))
-      ).asServer(SunHttp(port.orElse(8080).get()))
-
+      // First time this task is run, start the server
+      val ws = websockets(
+        "/ws" bind { ws: Websocket ->
+          websockets.add(ws)
+          ws.send(WsMessage(RELOAD_MESSAGE))
+          ws.onClose {
+            websockets.remove(ws)
+          }
+        }
+      )
+      val http = routes(
+        "/" bind static(
+          Directory(inputDir.get().asFile.absolutePath),
+          Pair("zipline", ContentType.TEXT_PLAIN)
+        )
+      )
+      val server = PolyHandler(http, ws).asServer(Jetty(8080))
       deploymentRegistry.start(
         deploymentId,
         DeploymentRegistry.ChangeBehavior.BLOCK,
         ZiplineServerDeploymentHandle::class.java,
         server
       )
+
+      // Keep the connection open by sending a message periodically
+      Timer("WebsocketHeartbeat", true).schedule(0, 10000) {
+        websockets.forEach { it.send(WsMessage("heartbeat")) }
+      }
+    } else {
+      // Subsequent task runs, just send a websocket message
+      websockets.forEach { it.send(WsMessage(RELOAD_MESSAGE)) }
     }
+  }
+
+  companion object {
+    const val RELOAD_MESSAGE = "reload"
+    val websockets: MutableList<Websocket> = mutableListOf()
   }
 }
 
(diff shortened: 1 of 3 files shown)


adrw (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
```diff
@@ -50,18 +56,44 @@ abstract class ZiplineServeTask : DefaultTask() {
...
       deploymentRegistry.start(
         deploymentId,
         DeploymentRegistry.ChangeBehavior.BLOCK,
         ZiplineServerDeploymentHandle::class.java,
         server
       )
+
+      // Keep the connection open by sending a message periodically
+      Timer("WebsocketHeartbeat", true).schedule(0, 10000) {
+        websockets.forEach { it.send(WsMessage("heartbeat")) }
```
JakeWharton, line 85: This should probably be a ping instead. OkHttp will actually automatically ping from the other side to determine if the connection is alive, but I suppose we don't want to rely on that.

JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
yissachar: Looks like OkHttp doesn't do this automatically but there is a [setting](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/ping-interval/) to turn it on. The connection was closing every 30 seconds previously which is why I added the heartbeat. 

I'm not sure if proper ping-pong really matters here. In the current implementation the client websocket will be closed if we don't receive a message every 30 seconds, and the server side should receive a close message most of the time when the websocket closes. There may be a slow resource leak over time where the server keeps some connections open that it shouldn't but given that this is a hot reload server we'll be restarting it constantly anyway and clearing out the old connections. That being said, I'm not particularly opposed to to ping-pong instead of heartbeat, but it's just a bit more code complication.

Thoughts?

JakeWharton: Is that our implementation that gives up after 30 seconds? Or is that just a default timeout?

A ping is nice because it never reaches the user code. Both clients and servers must respond automatically to pings with a pong to conform with the spec. And if it's an automatic OkHttp timeout that automatically closes the connection, the ping will keep it alive automatically.

yissachar: Ah, I didn't realize that it was part of the spec, that makes more sense. Apparently http4k does not conform though. It won't respond to OkHttp's pings. Looks like it doesn't [properly map frames ](https://github.com/http4k/http4k/blob/master/http4k-server/jetty/src/main/kotlin/org/http4k/server/Http4kWebSocketFrameHandler.kt#L29)to the backing Jetty. Womp womp.

I think the 30 seconds may be a server idle timeout, but I don't see an easy way to configure. 

So we're left with the heartbeat as the simplest way to make things work, even if it's not quite the proper way of doing things. Definitely less impressed with http4k at this point, though still seems fine for our particular use case. I'll just proceed with this approach unless you have objections.

JakeWharton: Yeah it also doesn't allow sending pings to which OkHttp would automatically reply (the spec is basically symmetric).

> Definitely less impressed with http4k at this point

Yep, same.

> though still seems fine for our particular use case. I'll just proceed with this approach unless you have objections.

Yep. Merge away.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
```diff
@@ -50,18 +56,44 @@ abstract class ZiplineServeTask : DefaultTask() {
...
+      }
+    } else {
+      // Subsequent task runs, just send a websocket message
+      websockets.forEach { it.send(WsMessage(RELOAD_MESSAGE)) }
     }
   }
+
+  companion object {
+    const val RELOAD_MESSAGE = "reload"
+    val websockets: MutableList<Websocket> = mutableListOf()
```
swankjesse, line 95: This needs synchronization for concurrent accesses.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
```diff
@@ -50,18 +56,44 @@ abstract class ZiplineServeTask : DefaultTask() {
...
+      val server = PolyHandler(http, ws).asServer(Jetty(8080))
       deploymentRegistry.start(
         deploymentId,
         DeploymentRegistry.ChangeBehavior.BLOCK,
         ZiplineServerDeploymentHandle::class.java,
         server
       )
+
+      // Keep the connection open by sending a message periodically
+      Timer("WebsocketHeartbeat", true).schedule(0, 10000) {
```
swankjesse, line 84: This is terse, but the right way to do this since ~2004 is ScheduledExecutorService. Or perhaps something with a coroutine.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
```diff
@@ -50,18 +56,44 @@ abstract class ZiplineServeTask : DefaultTask() {
...
+          }
+        }
+      )
+      val http = routes(
+        "/" bind static(
+          Directory(inputDir.get().asFile.absolutePath),
+          Pair("zipline", ContentType.TEXT_PLAIN)
+        )
+      )
+      val server = PolyHandler(http, ws).asServer(Jetty(8080))
```
swankjesse, line 75: Switching from SunHttp to Jetty is a bummer. You probably lose a decent amount of start-up time here.

MAYBE we finish OkHttp 5.x someday and we can use MockWebServer here, which is fast, small, and websockets-aware.

swankjesse (commented):
LGTM!

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
yissachar: Yeah, I would love to use MockWebServer, then we could get good testing too. What is the dependency on OkHttp 5? Can't we add proper WS support in 4.x?

PR #746, opened 2022-11-10
Make events more complete

swankjesse:
This adds a new 'Zipline' parameter to the EventListener where appropriate. I'm hoping we can use this to track which Zipline instance an event came through in applications where there is more than one.

This also adds an event when the load is skipped because that code is unchanged. This fixes a small bug where we emitted a start event but not an end event.

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -35,7 +35,7 @@ import kotlinx.coroutines.channels.ProducerScope
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.channelFlow
-import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.joinAll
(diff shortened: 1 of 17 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -344,26 +409,6 @@ class ZiplineLoader internal constructor(
     }
   }
 
-  /** Wrap [block] in start/end/failed events, recovering gracefully if a flow is canceled. */
-  private suspend fun withLifecycleEvents(
```
swankjesse, line 348: I effectively inlined this function in both its callsites

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/events.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+package app.cash.zipline.internal.bridge
+
+import app.cash.zipline.Call
+import app.cash.zipline.CallResult
+import app.cash.zipline.EventListener
+import app.cash.zipline.Zipline
+import app.cash.zipline.ZiplineService
+
+interface EndpointEventListener {
```
swankjesse, line 24: This is the subset of EventListener that endpoint publishes to, and without a Zipline parameter. I tried wiring that in but it wrecks tests that use Endpoint but not the JS engine.

adrw (approved):

PR #753, opened 2022-11-28
Update ZiplineServeTask inputDir to an input directory.

colinrtwhite:
Pretty sure this is required for compatibility with Gradle 7.6. Updating Redwood to use 7.6 caused `serveZipline` to [not watch any directories](https://github.com/cashapp/redwood/pull/554#issuecomment-1328442887).


JakeWharton (approved):

JakeWharton:
I wonder if we can write an integration test for this. Use GradleRunner to run the task, use OkHttp to hit the server, kill the runner (if that's even possible). At the very least we could do it with bash and curl as a crude fallback.

PR #755, opened 2022-11-30
Add generic val var ZiplineService tests

adrw:
Instead of just `kotlin.Int`, we should also have tests for more complex types like data classes and enums to confirm val/var support.


JakeWharton (approved):
Nice

PR #758, opened 2022-12-02
Workaround missing libtinfo.so.5 on Ubuntu

swankjesse:
We're getting this crash in our cklib builds:

    /home/runner/.cklib/clang-llvm-8.0.0-linux-x86-64/bin/clang: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory

StackOverflow suggests this might fix.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,10 +53,12 @@ jobs:
         if: matrix.os == 'macOS-latest'
         run: ./.github/workflows/build-mac.sh -a ${{ matrix.arch }} -c ${{ matrix.cmake-arch }}
 
+      - name: Install libtinfo5
+        if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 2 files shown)


swankjesse:
Related:
https://youtrack.jetbrains.com/issue/KT-42693/Remove-dependency-on-ncurses5-library

JakeWharton (approved):
This won't affect the ARM builds since they happen in the container, but maybe they're fine? Your error message seems to indicate x86_64 only but maybe there's other lines which were omitted.

swankjesse:
Also related: https://github.com/touchlab/cklib/issues/8

swankjesse:
I don’t think we do Linux + ARM, but it might be broken if we did.

JakeWharton:
It's two steps up in these actions. Ubuntu runner + docker container for cross-compiliation.

PR #759, opened 2022-12-02
Implement toString() more

swankjesse:
Should be useful when interacting with these types in a debugger.


swankjesse:
Screenshot 2022-12-02 at 4 34 03 PM

JakeWharton (approved):

PR #764, opened 2022-12-05
Make functions self-describe as suspending

swankjesse:


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/ziplineFunctions.kt
```diff
@@ -30,6 +30,8 @@ internal abstract class ReturningZiplineFunction<T : ZiplineService>(
   /** A serializer for a `kotlin.Result<T>` which supports success or failure. */
   val kotlinResultSerializer = ResultSerializer(resultSerializer)
 
+  override val isSuspending = false
```
JakeWharton:
```suggestion
  override val isSuspending get() = false
```

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/ziplineFunctions.kt
```diff
@@ -48,6 +50,7 @@ internal abstract class SuspendingZiplineFunction<T : ZiplineService>(
 ) : ZiplineFunction<T> {
   val argsListSerializer = ArgsListSerializer(argSerializers)
 
+  override val isSuspending = true
   override val isClose = false
```
JakeWharton:
```suggestion
  override val isSuspending get() = true
  override val isClose get() = false
```

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/ziplineFunctions.kt
swankjesse: I’d considered this. It generates better bytecode at a cost of _sliiiiightly_ larger sourcecode. Probably a good tradeoff considering that the sourcecode has fewer moving parts.

PR #773, opened 2022-12-16
ZiplineScoped initial plumbing

swankjesse:


swankjesse:
I think this shook out a bug in our JS leak detection. I’ve introduced a reference cycle that involves the object we’re tracking (the outbound service) and its scope, and I suspect it’s unsafe to run code in this situation.

To avoid the cycle I’d need to limit what the scope knows. That might be okay.

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007ff71fa3b759, pid=15812, tid=15814
#
# JRE version: OpenJDK Runtime Environment Zulu11.60+19-CA (11.0.17+8) (build 11.0.17+8-LTS)
# Java VM: OpenJDK 64-Bit Server VM Zulu11.60+19-CA (11.0.17+8-LTS, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [quickjs2619729428427931323.tmp+0x3b759]  __JS_FreeValueRT+0xa9
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /home/runner/work/zipline/zipline/zipline/core.15812)
#
# An error report file with more information is saved as:
# /home/runner/work/zipline/zipline/zipline/hs_err_pid15812.log
#
# If you would like to submit a bug report, please visit:
#   http://www.azul.com/support/
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
app.cash.zipline.LeakedServicesTest[jvm] > jsLeaksService[jvm] SKIPPED
```

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScope.kt
```diff
@@ -0,0 +1,88 @@
...
+
+  internal fun remove(result: ZiplineService) {
+    services -= result
+  }
+
+  fun close() {
+    if (closed) return
+    closed = true
+
+    val servicesCopy = services.toTypedArray() // Because ZiplineService.close() mutates the set.
```
JakeWharton, line 83: What about calling `services.clear()` here so that the calls to `remove()` are a no-op rather than wastefully doing a bunch of set ops?

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package app.cash.zipline
+
+/**
+ * Implement this interface on inbound services to specify which scope service arguments are added
+ * to. See [ZiplineScope] for details.
+ *
+ * Do not extend this in your [ZiplineService] interfaces; it should only be implemented by concrete
+ * implementation classes.
```
JakeWharton, line 23: Can the compiler validate this? At least on dealing with calls to `take` walk its interface hierarchy and validate none do this.

JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
swankjesse: Absolutely. On the list.

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScope.kt
swankjesse: Oooooh I like. Done.

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
swankjesse: https://github.com/cashapp/zipline/issues/775

PR #784, opened 2022-12-19
ZiplineService.withScope narrows a scope

### LINKED ISSUE #781: Mechanism to narrow a ZiplineScope
I’d like to have a `ZiplineScope` that applies during a small number of calls to a service.

Consider this example:

```kotlin
val scope = ZiplineScope()
val stockPriceService = zipline.take<StockPriceService>("stockPriceService", scope)

fun latestStockPrice(symbol: String): Money {
  val flow = stockPriceService.priceFlow(symbol)
  return flow.first().price
}
```

Each invocation of `latestStockPrice()` should get its own scope. We know when to close the scope (when the `latestStockPrice()` returns), but no API with which to do it.

One option:

```kotlin
suspend fun latestStockPrice(symbol: String): Money {
  val childScope = ZiplineScope(scope)
  try {
    val flow = scope.push(childScope) {
      stockPriceService.priceFlow(symbol)
    }
    return flow.first().price
  } finally {
    childScope.close()
  }
}
```

Note that we only `push` for the duration of the service function. We **don’t** push during `flow.first()`, since that’s a suspending operation (and may compete with other calls to `latestStockPrice()`.

swankjesse:
Closes: https://github.com/cashapp/zipline/issues/781

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
@@ -73,7 +73,7 @@ interface ZiplineScoped {
  * ```
  *
  * Note that closing the returned service will close the receiver: the two handles target the same object.
- * Closing [scope] will not close either.
+ * Closing either will not close [scope].
  */
 fun <T : ZiplineService> T.withScope(scope: ZiplineScope): T {
   require(this is OutboundService) { "cannot scope $this; it isn't an outbound service" }
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
@@ -72,7 +72,7 @@ interface ZiplineScoped {
  * scopeB.close() // closes nke.
  * ```
  *
- * Note that closing the returned service will close this: the two handles target the same object.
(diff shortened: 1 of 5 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
```diff
@@ -29,3 +31,51 @@ interface ZiplineScoped {
...
+ * val nke: LiveChart = quoteServiceWithScopeB.chart("NKE")
+ *
+ * scopeA.close() // closes quoteService.
+ *
+ * nke.setPeriod(...)
+ * nke.setPrecision(...)
+ * scopeB.close() // closes nke.
+ * ```
+ *
+ * Note that closing the returned service will close this: the two handles target the same object.
```
JakeWharton: Maybe
```suggestion
 * Note that closing the returned service will close the receiver: the two handles target the same object.
```

zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineScoped.kt
```diff
@@ -29,3 +31,51 @@ interface ZiplineScoped {
...
+ * Closing [scope] will not close either.
```
JakeWharton: This seems wrong, no? Your code example above contradicts this. Did you mean something like
```suggestion
 * Closing either will not close [scope].
```
?

JakeWharton (approved):

Issue #765, opened 2022-12-06
ZiplineService interface must be declared directly


swankjesse:
This doesn’t work as it aught to:

```
interface A : ZiplineService {
  fun a()
}

interface B : A {
  fun b()
}
```

In particular we don’t generate anything for `B` because our code rewriter doesn’t traverse up B’s inheritance hierarchy.

swankjesse:
Best solution here is to fix the generator to not impose this restriction.

PR #785, opened 2022-12-20
Compiler-enforce interfaces cannot extend ZiplineScoped

### LINKED ISSUE #775: Compiler enforcement that interfaces don’t extend ZiplineScoped
We want only implementation classes to implement this special interface.

swankjesse:
Closes: https://github.com/cashapp/zipline/issues/775


swankjesse:
We don’t want any classes to extend both `ZiplineScoped` and `ZiplineService`. If they did, then users’ might reasonably think that `ZiplineScoped` can be bridged, but scopes belong only to the outbound side.

JakeWharton (approved):

JakeWharton:
#765 next? Probably same code.

PR #842, opened 2023-01-10
Fix a bug where flows crash when they're closed

swankjesse:


swankjesse:
I’m inadvertently getting concurrency in my test. Will rework and re-open this PR.

PR #844, opened 2023-01-10
Don’t force a suspend

### LINKED ISSUE #826: callSuspending() should’t force a suspend
Right now InboundService always does a `launch()` and the caller always resumes asynchronously, even if there’s no blocking in the body of the service.

We should attempt to return synchronously when that’s possible.

On the API we’d change `callSuspending` to return _either_ an encoded CancelCallback or the synchronous result.

swankjesse:
Closes: https://github.com/cashapp/zipline/issues/826

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
@@ -125,6 +125,8 @@ internal class InboundService<T : ZiplineService>(
       return encodedSuspendingResult.encodedResult
 
     } else {
+      // Convert our Deferred<Result<T>> into a Result<T>, which will be failure if _either_ the
+      // Deferred failed or if it contains a failure Result.
       val failure = deferred.getCompletionExceptionOrNull()
       val kotlinResult = when {
         failure != null -> Result.failure(failure)
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
@@ -22,6 +22,7 @@ import app.cash.zipline.ZiplineService
 import app.cash.zipline.internal.encodeToStringFast
 import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.launch
+import kotlinx.serialization.KSerializer
 
 /**
  * Inbound calls use this to call into the real service.
@@ -68,9 +69,10 @@ internal class InboundService<T : ZiplineService>(
     externalCall: Call,
     suspendCallback: SuspendCallback<Any?>,
   ): String {
+    @Suppress("UNCHECKED_CAST") // We found this function by matching on its service type T.
+    val function = internalCall.function as SuspendingZiplineFunction<ZiplineService>
+
     val job = endpoint.scope.launch {
-      @Suppress("UNCHECKED_CAST") // We found this function by matching on its service type T.
-      val function = internalCall.function as SuspendingZiplineFunction<ZiplineService>
       val args = internalCall.args
 
       val callStart = endpoint.eventListener.callStart(externalCall)
@@ -108,7 +110,14 @@ internal class InboundService<T : ZiplineService>(
       if (name != null) endpoint.remove(name)
     }
 
-    return endpoint.json.encodeToStringFast(cancelCallbackSerializer, cancelCallback)
+    val suspendingResult = SuspendingResult<Unit>(
+      cancelCallback = cancelCallback,
+    )
+
+    return endpoint.json.encodeToStringFast(
+      function.suspendingResultSerializer as KSerializer<SuspendingResult<*>>,
+      suspendingResult,
+    )
   }
 
   override fun toString() = service.toString()
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
@@ -20,8 +20,9 @@ import app.cash.zipline.CallResult
 import app.cash.zipline.ZiplineFunction
 import app.cash.zipline.ZiplineService
 import app.cash.zipline.internal.encodeToStringFast
-import kotlinx.coroutines.ensureActive
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
+import kotlinx.coroutines.async
+import kotlinx.coroutines.isActive
 import kotlinx.serialization.KSerializer
 
 /**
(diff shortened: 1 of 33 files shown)


JakeWharton:
Requires more than a quick perusal so I'll look later.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
```diff
@@ -296,3 +297,78 @@ internal class ResultSerializer<T>(
     }
   }
 }
+
+/**
+ * Immediate result from invoking a suspending function, that may include either a value (if the
+ * call never suspended) or a cancel callback (if it did suspend).
+ */
+internal class SuspendingResult<T>(
```
swankjesse, line 305: Considering for follow-up . . . 

We have a bunch of result types:
 - kotlin.Result - success or failure
 - app.cash.zipline.CallResult - success or failure, encoded JSON, service names
 - app.cash.zipline.internal.bridge.SuspendingResult - success or failure or callback
 - app.cash.zipline.internal.bridge.EncodedSuspendingResult - success or failure or callback, encoded JSON, service names

We also have serializers for kotlin.Result and SuspendingResult.

Some options to reduce how much code we need:
 * Combine the two serializers to save a bit of code. Rename SuspendingResult to InternalResult for symmetry with InternalCall.
 * Change EncodedSuspendingResult to have a `kotlin.Result` as a field

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
```diff
@@ -68,47 +68,88 @@ internal class InboundService<T : ZiplineService>(
...
+          deferred.cancel()
+        }
+
+        override fun toString() = "CancelCallback/$internalCall"
       }
-    }
 
-    val cancelCallback = object : CancelCallback, HasPassByReferenceName {
-      override var passbyReferenceName: String? = null
+      deferred.invokeOnCompletion {
```
JakeWharton, line 94: I'm a bit worried about the Retrofit case here: where `isActive` returns `true` but by the time you get to this code the `async` has completed and thus this lambda will be invoked synchronously (I'm assuming that's the behavior).

It would mean that we invoke the `suspendCallback` below synchronously from the stackframe which invoked the `suspend` function (and before we allow it to return a value). The normal coroutine machinery defends against this (which is what was happening in Retrofit).

In our case, to defend against it, I think we need something like the following:
```kotlin
var scheduleCallback = true
deferred.invokeOnCompletion {
  // ..

  if (scheduleCallback) {
    ziplineDispatcher.schedule({ /* callback logic */ })
  } else {
    // callback logic
  }
}
scheduleCallback = false

File: zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt [hunk 3]
```diff
@@ -68,47 +68,88 @@ internal class InboundService<T : ZiplineService>(
...
-      override fun toString() = "CancelCallback/$internalCall"
-    }
-    job.invokeOnCompletion {
-      val name = cancelCallback.passbyReferenceName
-      if (name != null) endpoint.remove(name)
-    }
+      val suspendingResult = when {
+        kotlinResult.isFailure -> SuspendingResult<Unit>(failure = kotlinResult.exceptionOrNull()!!)
+        else -> SuspendingResult(success = kotlinResult.getOrNull())
+      }
```
JakeWharton, line 139: These two `when`'s need merged or something. You create the `kotlin.Result` and then unpack it just to match on it, force unwrap, and then re-wrap it.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
```diff
@@ -214,6 +214,7 @@ internal class RealCallSerializer(
       return object : SuspendingZiplineFunction<T>(
         name = functionName,
         argSerializers = listOf(),
+        resultSerializer = Int.serializer(), // Placeholder; we're only encoding failures.
```
JakeWharton, line 217: There's a `NothingSerializer` coming in the next version of kotlinx.coroutines.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
```diff
@@ -296,3 +297,78 @@ internal class ResultSerializer<T>(
...
+) {
+  val serviceNames: List<String> = serviceNames.toList() // Defensive copy.
+}
+
+internal class SuspendingResultSerializer<T>(
+  internal val successSerializer: KSerializer<T>,
+) : KSerializer<SuspendingResult<T>> {
+
+  override val descriptor: SerialDescriptor = buildClassSerialDescriptor("Result") {
+    element("cancelCallback", ThrowableSerializer.descriptor)
```
JakeWharton:
```suggestion
    element("cancelCallback", cancelCallbackSerializer.descriptor)
```

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
```diff
@@ -296,3 +297,78 @@ internal class ResultSerializer<T>(
...
+ */
+internal class SuspendingResult<T>(
+  /** The function suspended. Only non-null for suspend calls. */
+  val cancelCallback: CancelCallback? = null,
+
+  /** The function failed without suspending. */
+  val failure: Throwable? = null,
+
+  /** The function succeeded without suspending. */
+  val success: T? = null,
```
JakeWharton, line 313: Do you think we need an `init { }` which enforces the invariant that one (and only one) is set? Deserialization below could create an instance with all `null`s, for example.

JakeWharton (approved):
Cool stuff

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
swankjesse: It’s single-threaded, so we don‘t need to worry about that.

JakeWharton: As in, the `async` is running on the same dispatcher as this block? That's good.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
swankjesse: Ooooh it’s weirder than that. Added this comment:

```
      // Convert our Deferred<Result<T>> into a Result<T>, which will be failure if _either_ the
      // Deferred failed or if it contains a failure Result.
```

swankjesse: (Is this possible??)

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
swankjesse: Oooh want.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
swankjesse: Good idea. Done.

swankjesse:
Ugh, I’ve got some unwanted concurrency in the tests.
```
> Process 'Gradle Test Executor 1' finished with non-zero exit value 132
```

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
swankjesse: Exactly.

swankjesse:
Okay so eventually I decided to run `java` without Gradle wrapping it, and that was incredibly helpful because I started getting extra information. The JVM printed this before it crashed:

```
Illegal instruction: 4
```

After scouring Stackoverflow for that error I came across [this question](https://stackoverflow.com/questions/60666890/illegal-instruction-4-from-a-jni-application-on-macos-10-13-6), which recommended adding `-Xss2048k` to the Java command. That solved the crashes on x86_64.

It’s Quite Unfortunate that our stacks are deep enough to cause a JVM process crash.

Issue #828, opened 2022-12-23
SuspendCallbacks can leak


swankjesse:
I’ve got a Redwood application that creates and cancels flows, and in some situations I’ve got `SuspendCallback` objects that never get called back, and never get canceled. I suspect it’s a book-keeping issue with ZiplineScope, possibly exacerbated by our flows.

I’m able to create this leak by creating and closing services. The leaked callback is tracked in `Endpoint.inboundServices`; here’s a sample.

```
zipline/host-1029 -> {InboundService@56720} SuspendCallback/Call(receiver=zipline/js-865, function=suspend fun emit(T): kotlin.Unit, args=[HostConfiguration(darkMode=false)])
```

swankjesse:
I traced through and learned what’s happening:

1. Outbound side makes a suspending call to a target service _T_ that passes a service _K_ by-reference, like `FlowZiplineService.collect()`
2. The call is encoded, which puts _K_ in the outbound side’s inboundServices map
3. That coroutine suspends in `OutboundCallHandler.callSuspending()`
4. The target service _T_ is closed by some other coroutine
5. The target service cannot decode the call because _T_ is closed, and therefore doesn’t discover that _K_ needs to be closed

I’m gonna fix #826 first, which will prevent that untimely suspension.

One other gotcha here: the outbound side does receive the `ZiplineApiMismatchException` crash because its coroutine is already canceled by the time that arrives. This is a problem because we don’t get a signal that a leak could have happened.

damianw:
I found a similar issue with using Flows generally - it seems that both `SuspendCallback` and `CancelCallback`s are leaked. I created a sample project based on the `trivial` example here: https://github.com/damianw/zipline-leaky-flow

Running the sample prints some leaks. Using the debugger, we can see that these were `SuspendCallback` and `CancelCallback`. Is this ultimately the same problem as described in this issue, or should I file a new issue?

swankjesse:
I expect it’s likely the same issue. Could I talk you into contributing a test case?

damianw:
Sure, will take a look. 👍

damianw:
Added a test in #850, though I'm not sure that this is really the same issue anymore. Or maybe it is, and nothing is really leaking. Let me know what you think.

swankjesse:
Confirmed fixed by the `ZiplineScope` features.

PR #850, opened 2023-01-12
Add test that 0 leaks are reported after collecting Flow, closing service

damianw:
This adds a test for the behavior being discussed in #828. As I looked over `FlowTest`, I came to the suspicion that this issue is not about whether something leaks, but rather whether it gets reported as leaking. I'm not sure if anything is really leaking, but the listener says so.

The test is currently ignored. To make this work, I made `allReferencesSet` `internal` in order to manually enqueue the references. That's probably not the best, so let me know if you have other suggestions. If you don't think this is the same bug as #828, I can file a separate issue and update the `Ignore` reason.


swankjesse (approved):
Thanks for the work.

Unfortunately I think I duplicated your work with the ZiplineScoped implementation.

PR #853, opened 2023-01-16
Revert AGP to 7.3.1.

colinrtwhite:
[Intellij doesn't support AGP 7.4.0 yet](https://youtrack.jetbrains.com/issue/IDEA-280006/Improve-Android-plugin-merging-cadence).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,7 +14,7 @@ targetSdk = "31"
 
 [libraries]
 android-desugarJdkLibs = { module = "com.android.tools:desugar_jdk_libs", version = "1.2.2" }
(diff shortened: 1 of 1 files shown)


JakeWharton:
Jesse noted the lack of support in his PR to update yet still merged. I"ll defer to him.

JakeWharton:
Actually that was in Okio https://github.com/square/okio/pull/1202

JakeWharton (approved):

swankjesse:
Yeah, I think it’s worth prioritizing IntelliJ support 'cause it handles multiplatform better. Probably a mistake on my part with Okio.

JakeWharton:
I'm not sure that's universally true (Redwood started that way but AS became better for it), but if we don't need 7.4 then there's no reason to jump up so fast either.

PR #862, opened 2023-01-18
Support Webpack builds

swankjesse:
These give up the benefits of modules (independent caching, pipelined download and code-load), in exchange for the benefits of Webpack (minification, fewer downloads).

### DIFF
--- a/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
+++ b/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
@@ -21,6 +21,7 @@ import app.cash.zipline.loader.SignatureAlgorithmId
 import java.io.File
 import java.io.Serializable
 import okio.ByteString
+import okio.ByteString.Companion.decodeHex
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.provider.ListProperty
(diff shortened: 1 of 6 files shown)


zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
```diff
@@ -61,6 +62,35 @@ abstract class ZiplineCompileTask : DefaultTask() {
...
+    )
+
+    mainModuleId.set(extension.mainModuleId)
+    mainFunction.set(extension.mainFunction)
+    version.set(extension.version)
+
+    signingKeys.set(project.provider {
+      extension.signingKeys.asMap.values
+    }.flatMap {
+      it.map { dslKey ->
```
adrw, line 86: Should this be flatMap?

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplinePluginTest.kt
```diff
@@ -60,6 +60,26 @@ class ZiplinePluginTest {
...
+    val taskName = ":lib:jsBrowserProductionWebpackZipline"
+    val result = createRunner(projectDir, "clean", taskName).build()
+    assertThat(SUCCESS_OUTCOMES)
+      .contains(result.task(taskName)!!.outcome)
+
+    val ziplineOut = projectDir.resolve(
+      "lib/build/distributionsZipline"
+    )
+    assertThat(ziplineOut.resolve(MANIFEST_FILE_NAME).exists()).isTrue()
+    assertThat(ziplineOut.resolve("lib.zipline").exists()).isTrue()
```
adrw, line 81: nit: maybe an assertion that only the 2 files were output (manifest and lib.zipline)

adrw (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineCompileTask.kt
adrw: Nvm, tried locally and it wouldn't work in this case

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplinePluginTest.kt
adrw: new commit added with this change

swankjesse: Thanks!

PR #876, opened 2023-02-01
Move websocket development server from Redwood to Zipline

adrw:


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/FastCodeUpdates.kt
```diff
@@ -0,0 +1,56 @@
+package app.cash.zipline.loader
```
swankjesse, line 16: supernit: copyright

(I think we might be due to get Spotless managing this?)

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/FastCodeUpdatesTest.kt
```diff
@@ -0,0 +1,75 @@
+package app.cash.zipline.loader
```
swankjesse, line 16: Copyright!

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/FastCodeUpdatesTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+      // close channel
+      client.closeUpdatesChannel()
+
+      // await URL flow emitting twice: once at time 0, once at 500ms on polling fallback
+      //    from socket failure
+      fastCodeFlow.take(2).toList()
+    }
+
+    // confirm >= 500 ms has elapsed
```
swankjesse, line 118:
```suggestion
    // Confirm 500 ms has elapsed.
```

swankjesse (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/FastCodeUpdates.kt
JakeWharton: I'll do this. I just had to do it for paraphrase and will be updating redwood and friends with the new setup.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
```diff
@@ -61,7 +61,6 @@ abstract class ZiplineServeTask : DefaultTask() {
       val ws = websockets(
         "/ws" bind { ws: Websocket ->
           websockets.add(ws)
-          ws.send(WsMessage(RELOAD_MESSAGE))
```
adrw, line 64: This is a deliberate change from the current implementation in Redwood which assumes that a reload is sent initially. Instead we immediately pass through and don't wait on websocket connect.

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/FastCodeUpdatesTest.kt
adrw: Lawyers! Be thee at ease...

yissachar (approved):

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/FastCodeUpdatesTest.kt
swankjesse: It is 2023 tho

PR #871, opened 2023-01-25
Update Kotlin to 1.8.

colinrtwhite:
Resolves https://github.com/cashapp/zipline/issues/888.

PR #880, opened 2023-02-04
feat: kotlin 1.8 support

epicadk:


JakeWharton:
Hi. Thanks for the PR!

Unfortunately we already have a Kotlin 1.8 upgrade ready to go in #871. We actually landed it once but had to revert to fix some issues only seen in the wild. If you have any improvements to make for 1.8 can you send them as changes which target that branch instead?

epicadk:
Blocked by sqldelight for now.

epicadk:
My bad should have gone through the other prs

PR #884, opened 2023-02-06
Put a copyright header on new files from IDEA

JakeWharton:


.idea/copyright/Square.xml
```diff
@@ -0,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="notice" value="Copyright (C) &amp;#36;{today.year} Square, Inc.&#10;&#10;Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);&#10;you may not use this file except in compliance with the License.&#10;You may obtain a copy of the License at&#10;&#10;     http://www.apache.org/licenses/LICENSE-2.0&#10;&#10;Unless required by applicable law or agreed to in writing, software&#10;distributed under the License is distributed on an &quot;AS IS&quot; BASIS,&#10;WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.&#10;See the License for the specific language governing permissions and&#10;limitations under the License." />
```
adrw: Is this supposed to be Block?

adrw (approved):

.idea/copyright/Square.xml
JakeWharton: Ew, hell no! I can do "Cash App" if that's what we want. Lawyers okayed that, too.

JakeWharton:
Note that once we get Spotless I think we'll have to normalize them all to be the same… we'll see.

.idea/copyright/Square.xml
adrw: Cash App would be dope

swankjesse (commented):
Nice

PR #928, opened 2023-03-24
Don't ClassCastException in the dev server

swankjesse:
There's some kind of isolation happening with --continuous that causes subsequent runs of our task to crash like this:

    Execution failed for task ':playground:treehouse-playground:serveDevelopmentZipline'.
    > Failed to cast object app.cash.zipline.gradle.ZiplineServerDeploymentHandle_Decorated@730ae4d3 of type app.cash.zipline.gradle.ZiplineServerDeploymentHandle_Decorated to target type app.cash.zipline.gradle.ZiplineServerDeploymentHandle

I believe this might fix by not casting to the plugin's type directly.

### DIFF
--- a/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
+++ b/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
@@ -16,9 +16,6 @@
 
 package app.cash.zipline.gradle
 
-import java.util.Timer
-import javax.inject.Inject
-import kotlin.concurrent.schedule
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.DirectoryProperty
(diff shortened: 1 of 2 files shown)


zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplineServeTask.kt
```diff
@@ -55,62 +39,21 @@ abstract class ZiplineServeTask : DefaultTask() {
...
     }
   }
-
-  companion object {
-    const val RELOAD_MESSAGE = "reload"
-    val websockets: MutableList<Websocket> = mutableListOf()
-  }
-}
-
-internal open class ZiplineServerDeploymentHandle @Inject constructor(
```
swankjesse, line 99: I extracted this to a top-level file and put more of the server stuff in it.

JakeWharton (approved):

PR #948, opened 2023-04-17
Fix a longstanding off-by-one in Source Maps rewriting

swankjesse:


zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectReader.kt
```diff
@@ -103,7 +103,7 @@ class JsObjectReader(
     val bytecode = source.readByteString(bytecodeLength.toLong())
     // TODO: fixup atoms within bytecode?
 
-    val hasDebug = flags.bit(10)
+    val hasDebug = flags.bit(11)
```
swankjesse, line 106: If you’re curious where 11 comes from, it’s this line:
https://github.com/cashapp/zipline/blob/trunk/zipline/native/quickjs/quickjs.c#L34299-L34311

What I missed was this call, which increments the index by 2:
```
    bc_set_flags(&flags, &idx, b->func_kind, 2);
```

JakeWharton (approved):

colinrtwhite (approved):

PR #949, opened 2023-04-18
Work-around Dokka Gradle 8 issue.

colinrtwhite:
Works around an [implicit dependency issue](https://github.com/cashapp/zipline/actions/runs/4727842290/jobs/8389453947).


JakeWharton (approved):
Not sure GFM is appropriate to publish over HTML, but it's what's already configured so...

PR #965, opened 2023-04-27
Add `OptIn` annotation for engine-specific APIs

### LINKED ISSUE #964: `Zipline`'s `quickJs` property and `QuickJs` type should have an opt-in annotation
Something to indicate you're mucking around with the low-level engine which is dangerous, can break Zipline, and isn't the engine we want to commit to long-term. It's an implementation detail that we're required to expose publicly because there are very good reasons for accessing it, but you still need to take care.

JakeWharton:
These are usable, but dangerous.

Closes #964

### DIFF
--- a/zipline/src/engineMain/kotlin/app/cash/zipline/EngineApi.kt
+++ b/zipline/src/engineMain/kotlin/app/cash/zipline/EngineApi.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Square, Inc.
+ * Copyright (C) 2023 Cash App
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


zipline/src/engineMain/kotlin/app/cash/zipline/EngineApi.kt
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package app.cash.zipline
+
+@RequiresOptIn(
+  message = "This API exposes the low-level JS engine and should be used with extreme care",
```
swankjesse, line 20: ‘should be used with extreme care’ –> ‘won’t work when the engine changes’

swankjesse:
```suggestion
  message = "This API exposes the low-level JS engine. This API will change when the engine changes.",
```

swankjesse (approved):

PR #967, opened 2023-05-03
Cache ziplineFunctions

chrisbanes:
`ziplineFunctions` fetches all of the serializers necessary, but currently it can be called a lot, meaning we end up doing a lot of work repeatedly. This PR caches the result of `ziplineFunctions`, enabling us to only do the work once.

### DIFF
--- a/zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
+++ b/zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
@@ -249,6 +249,9 @@ internal class AdapterGenerator(
     val serialNameProperty = irSerialNameProperty(adapterClass, constructor)
     adapterClass.declarations += serialNameProperty
 
+    val simpleNameProperty = irSimpleNameProperty(adapterClass)
+    adapterClass.declarations += simpleNameProperty
+
     val serializersProperty = irSerializersProperty(adapterClass, constructor)
     adapterClass.declarations += serializersProperty
 
@@ -295,8 +298,8 @@ internal class AdapterGenerator(
   }
 
   /**
-   * Override `ZiplineServiceAdapter.serialName`. The constant value is the service's simple name,
-   * like "SampleService".
+   * Override `ZiplineServiceAdapter.serialName`. The constant value is the service's full
+   * typed name like "my.package.SampleService<package.Type1>".
    */
   private fun irSerialNameProperty(adapterClass: IrClass, value: IrConstructor): IrProperty {
     // override val serialName: String = serialName
@@ -308,6 +311,23 @@ internal class AdapterGenerator(
       overriddenProperty = ziplineApis.ziplineServiceAdapterSerialName,
     ) {
       irExprBody(irGet(value.valueParameters[1]))
+    }
+  }
+
+  /**
+   * Override `ZiplineServiceAdapter.simpleName`. The constant value is the service's simple name,
+   * like "SampleService".
+   */
+  private fun irSimpleNameProperty(adapterClass: IrClass): IrProperty {
+    // override val simpleName: String = "MyClass"
+    return irVal(
+      pluginContext = pluginContext,
+      propertyType = pluginContext.symbols.string.defaultType,
+      declaringClass = adapterClass,
+      propertyName = ziplineApis.ziplineServiceAdapterSimpleName.owner.name,
+      overriddenProperty = ziplineApis.ziplineServiceAdapterSimpleName,
+    ) {
+      irExprBody(irString(original.name.identifier))
     }
   }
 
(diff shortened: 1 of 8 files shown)


zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/CallAdapterConstructorRewriter.kt
```diff
@@ -54,6 +55,9 @@ internal class CallAdapterConstructorRewriter(
     val kClassArgument = original.getValueArgument(0) ?: return original
     val serializersListExpression = original.getValueArgument(1) ?: return original
 
+    // my.package.AdapterClass<package.Type1, package.Type2>
+    val serialName = (original.getTypeArgument(0) as IrSimpleType).asString()
```
chrisbanes, line 59: Not sure how safe this is. I saw it used in other places.

chrisbanes (commented):
I couldn't find a great way to test this change. I logged the generated `serialName` locally during tests and they looked as expected:

`app.cash.zipline.testing.EchoZiplineService`

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -115,6 +116,8 @@ class Endpoint internal constructor(
     }
   }
 
+  private val functionsCache = mutableMapOf<String, List<ZiplineFunction<*>>>()
```
chrisbanes, line 119: Might want to use a LruCache? Probably fine for now though.

swankjesse: Fine for now. There’s no LruCache in Kotlin/JS!

zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
```diff
@@ -282,24 +298,41 @@ internal class AdapterGenerator(
...
+    ) {
+      irExprBody(irGet(value.valueParameters[1]))
+    }
+  }
+
+  /**
+   * Override `ZiplineServiceAdapter.simpleName`. The constant value is the service's simple name,
+   * like "SampleService".
+   */
+  private fun irSimpleNameProperty(adapterClass: IrClass): IrProperty {
```
swankjesse, line 321: Drop this? I don’t think the benefit is worth the extra code.

swankjesse: Maybe a `name` property that’s used for both caching and toString() ?

zipline-kotlin-plugin/tests/src/test/java/app/cash/zipline/kotlin/ZiplineTestInternals.java
```diff
@@ -149,10 +155,14 @@ public Object call(EchoService service, List<?> args) {
       extends ZiplineServiceAdapter<GenericEchoService<String>> {
     public static final GenericEchoServiceAdapter INSTANCE = new GenericEchoServiceAdapter();
 
-    @Override public String getSerialName() {
+    @Override public String getSimpleName() {
       return "GenericEchoService";
     }
 
+    @Override public String getSerialName() {
+      return "app.cash.zipline.kotlin.GenericEchoService<kotlin.String>";
```
swankjesse, line 163: yay

zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
chrisbanes: That was my initial plan, but the simple name was returned from `toString()`. I wasn't sure if anything else was using `toString()` as some sort of key (it was named `serialName` so I guessed so). Felt safer to keep it, but happy to remove it.

zipline/src/engineTest/kotlin/app/cash/zipline/SampleService.kt
```diff
@@ -61,8 +63,9 @@ interface SampleService<T> : ZiplineService {
      */
     internal class ManualAdapter<TX>(
       override val serializers: List<KSerializer<*>>,
+      override val serialName: String,
     ) : ZiplineServiceAdapter<SampleService<TX>>() {
-      override val serialName: String = "SampleService"
+      override val simpleName: String = "ManualAdapter"
```
swankjesse:
```suggestion
      override val simpleName: String = "SampleService"
```

swankjesse: (this should be the name of the service it’s adapting, not the adapter type)

swankjesse (approved):

swankjesse:
This is wonderful

zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
chrisbanes: I'm going to stick with this as-is for now, but we can revisit later if you feel strong about this. I think `toString()` should behave as if we running _with_ the JVM's normal type erasure, and `serialName` definitely doesn't do that 😄

PR #975, opened 2023-05-06
Group Compose compiler into Kotlin update PRs

JakeWharton:
This works really well on paraphrase. Copying to here!

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -9,5 +9,13 @@
       "matchPackageNames": ["co.touchlab:cklib-gradle-plugin"],
       "allowedVersions": "!/1\\.5\\.31\\.2/",
     },
+    {
+      // Compose compiler is tightly coupled to Kotlin version
+      "groupName": "Kotlin and Compose",
+      "matchPackagePrefixes": [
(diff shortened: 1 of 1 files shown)


JakeWharton:
Hmm we actually might need a dummy version catalog entry for the Compose compiler as only the version is present now.

chrisbanes:
Yeah I had to do similar in Tivi: https://github.com/chrisbanes/tivi/pull/1216

renovate.json5
```diff
@@ -9,5 +9,13 @@
       "matchPackageNames": ["co.touchlab:cklib-gradle-plugin"],
       "allowedVersions": "!/1\\.5\\.31\\.2/",
     },
+    {
+      // Compose compiler is tightly coupled to Kotlin version
+      "groupName": "Kotlin and Compose",
+      "matchPackagePrefixes": [
+        "androidx.compose.compiler",
+        "org.jetbrains.kotlin",
```
JakeWharton:
```suggestion
        "org.jetbrains.kotlin:kotlin",
```

swankjesse (approved):

PR #989, opened 2023-05-18
Promote ZiplineManifest to the main zipline module

swankjesse:
I'd like to either add ZiplineManifest as a parameter to EventListener, or add it as an optional member on Zipline. In either case promoting it to zipline is working towards that goal.

Related: https://github.com/cashapp/redwood/issues/640


zipline/src/engineMain/kotlin/app/cash/zipline/ZiplineManifest.kt
swankjesse, line 1: This is in `engineMain`, not `commonMain` !

zipline/src/engineMain/kotlin/app/cash/zipline/internal/serializers.kt
```diff
@@ -0,0 +1,74 @@
...
+
+  // Because new releases may change default values, it's best to encode them.
+  encodeDefaults = true
+}
+
+/**
+ * Confirm the manifest is of a reasonable size before proceeding to operate on it. 10 KiB is a
+ * typical size for our test applications. 640K ought to be enough for anybody.
+ */
+const val MANIFEST_MAX_SIZE = 640 * 1024
```
JakeWharton, line 57: Would be nice to hide these from the actual public API. What about using `internal` visibility and then suppressing the error from the loader (until friend modules are easier to use).

JakeWharton (approved):

zipline/src/engineMain/kotlin/app/cash/zipline/internal/serializers.kt
swankjesse: Ooooh I didn’t even know that was a thing. Coming up!

swankjesse: https://github.com/cashapp/zipline/pull/995

Issue #645, opened 2022-08-11
ZiplineFunction.name is the signature, not the name


swankjesse:
It’s a string like this:
```
fun ping(app.cash.zipline.SampleRequest): app.cash.zipline.SampleResponse
```

I think the name should be just `ping` and the rest should be the signature?

swankjesse:
Today we use this signature string as a unique identifier for a function. I’d like to avoid serializing this long thing so much, and so propose the following:

1. Create a 53-bit hash ID of each function. I’d create this by hashing the existing signature, then taking the first 53 bits of that as a integer-valued Double. (53-bit double and not 64-bit long because JavaScript.)
2. Use that hash ID in `RealCallSerializer` instead of the full function signature.
3. Change `unknownFunction()` to omit the called function signature on the receiving side. (It’ll still be available in the stack trace.)

Thoughts?

PR #1010, opened 2023-05-26
Introduce ZiplineServiceType

### LINKED ISSUE #556: ZiplineClass
I’m not sure of the name. We should consider adding a type to reflectively describe a `ZiplineService`.

```
interface ZiplineClass {
  val functions: List<ZiplineFunction>
}
```

Then we can have this:
```
class Zipline {
  val services: Map<String, ZiplineClass>
  val clients: Map<String, ZiplineClass>
}
```

One challenge: would we serialize the `ZiplineFunction` instances so they always described the inbound side? This is interesting if inbound and outbound have different functions for the same service.

swankjesse:
This is a new reflection-like API to introspect what functions are available on the other side of the bridge.

Working towards this: https://github.com/cashapp/zipline/issues/645

Closes: https://github.com/cashapp/zipline/issues/556

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
@@ -33,7 +33,7 @@ import kotlinx.serialization.modules.SerializersModule
  * An outbound channel for delivering calls to the other platform, and an inbound channel for
  * receiving calls from the other platform.
  */
-class Endpoint internal constructor(
+internal class Endpoint internal constructor(
   internal val scope: CoroutineScope,
   internal val userSerializersModule: SerializersModule,
   internal val eventListener: EventListener,
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
@@ -17,7 +17,6 @@ package app.cash.zipline.internal.bridge
 
 import app.cash.zipline.Call
 import app.cash.zipline.CallResult
-import app.cash.zipline.ZiplineFunction
 import app.cash.zipline.ZiplineScope
 import app.cash.zipline.ZiplineService
 import app.cash.zipline.internal.EndpointService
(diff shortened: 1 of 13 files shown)


zipline/src/hostTest/kotlin/app/cash/zipline/ZiplineServiceTypeTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    val functions = type.functions.sortedBy { it.name }
+    assertEquals(2, functions.size)
+
+    assertEquals("fun close(): kotlin.Unit", functions[0].name)
+    assertFalse(functions[0].isSuspending)
+    assertTrue(functions[0].isClose)
+
+    assertEquals(
+      "fun echo(app.cash.zipline.testing.EchoRequest): app.cash.zipline.testing.EchoResponse",
```
swankjesse, line 62: I haven’t yet taken the next step of making this accessible through the public API.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -112,7 +111,7 @@ class Endpoint internal constructor(
     }
   }
 
-  private val functionsCache = mutableMapOf<String, List<ZiplineFunction<*>>>()
+  private val serviceTypeCache = mutableMapOf<String, RealZiplineServiceType<*>>()
```
swankjesse, line 114: I’m changing the cache from `List<ZiplineFunction>` to `ZiplineServiceTypes`. Big benefit there is we don’t need to build the HashMap organized by name every single time.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -153,8 +146,8 @@ class Endpoint internal constructor(
     // Detect leaked old services when creating new services.
     detectLeaks()
 
-    val functions = adapter.ziplineFunctions(json.serializersModule)
-    val callHandler = OutboundCallHandler(name, this, adapter, scope, functions)
+    val type = serviceType(adapter)
```
swankjesse, line 149: Now we’re caching on both sides of the endpoint.

JakeWharton (approved):
Good red bits

PR #1014, opened 2023-05-29
Move API mismatch details to the caller

swankjesse:
This should make it possible to not send the developer-facing function name across the bridge.

### DIFF
--- a/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineApiMismatchException.kt
+++ b/zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineApiMismatchException.kt
@@ -15,4 +15,18 @@
  */
 package app.cash.zipline
 
-class ZiplineApiMismatchException(override val message: String) : Exception(message)
+class ZiplineApiMismatchException(override val message: String) : Exception(message) {
+  internal companion object {
+    /**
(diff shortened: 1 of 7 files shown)


zipline/src/commonMain/kotlin/app/cash/zipline/ZiplineApiMismatchException.kt
```diff
@@ -15,4 +15,18 @@
...
 package app.cash.zipline
 
-class ZiplineApiMismatchException(override val message: String) : Exception(message)
+class ZiplineApiMismatchException(override val message: String) : Exception(message) {
+  internal companion object {
+    /**
+     * This magic value is detected by the bridge and replaced with an actionable list of available
+     * functions.
+     */
+    internal const val UNKNOWN_FUNCTION = "<unknown function>"
```
swankjesse, line 24: Today I learned – even though the companion is `internal`, if these aren’t `internal` then they leak in the public API!

JakeWharton: Only to Java, although Java can see all internal things with their names mangled.

JakeWharton (approved):

PR #1023, opened 2023-06-01
Flatten Gradle modules

swankjesse:
Nesting 'testing' inside 'zipline' and 'tests' inside 'zipline-kotlin-plugin' make it difficult to use the binary-compatibility-validator and dokka plugins.

The binary-compatibility-validator plugin applies itself to nested modules. We don't want this.

The dokka plugin crashes on multi-module builds.
https://github.com/Kotlin/dokka/issues/2954


zipline/testing/api/testing.api
swankjesse, line 1: Deleting this file is one immediate benefit of this PR.

JakeWharton (approved):

PR #1024, opened 2023-06-01
Use Dokka HtmlMultiModule

swankjesse:


build.gradle.kts
```diff
@@ -102,19 +100,35 @@ subprojects {
   }
 }
 
+tasks.named("dokkaHtmlMultiModule", DokkaMultiModuleTask::class.java).configure {
+  moduleName.set("Zipline")
+  outputDirectory.set(project.file("${project.rootDir}/docs/0.x"))
```
JakeWharton: This should be inside the build directory or .gitignore'd. Otherwise you dirty the working tree by running this task locally.

Additionally, both the `build.yaml` and `release.yaml` currently expect the output of this task to go into `build/dokka/htmlMultiModule/` (the default location) so that it can be deployed to static hosting.

JakeWharton (approved):

build.gradle.kts
swankjesse: Lemme leave the destination directory as it was.

PR #1035, opened 2023-06-05
Run tests in a background thread on native

swankjesse:
This is difficult because the default stack size for background threads is different from the default stack size on the main thread: 512 KiB for background vs. 8 MiB for main.

This is setting infrastructure in place to promote more tests from jvmTest to hostTest.

### DIFF
--- a/zipline-testing/src/appleMain/kotlin/app/cash/zipline/testing/CoroutineDispatchersApple.kt
+++ b/zipline-testing/src/appleMain/kotlin/app/cash/zipline/testing/CoroutineDispatchersApple.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2023 Cash App
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


JakeWharton (approved):

zipline-testing/src/appleMain/kotlin/app/cash/zipline/testing/CoroutineDispatchersApple.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.CloseableCoroutineDispatcher
+import kotlinx.coroutines.Runnable
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.runBlocking
+import platform.Foundation.NSThread
+
+actual fun singleThreadCoroutineDispatcher(
+  name: String,
+  stackSize: Int,
+): CloseableCoroutineDispatcher {
```
JakeWharton, line 29: It's a shame there's not a better way to do this where we don't have to manage the queue and lose access to things like `Delay` support. I suspect while the native dispatchers are `Worker`-based we don't have a choice. Maybe once they start exposing threads we can revisit this implementation.

swankjesse: Agreed. https://github.com/cashapp/redwood/issues/1181

PR #1040, opened 2023-06-11
Fix collision in serviceTypeCache keys

### LINKED ISSUE #1039: ServiceTypeCache keys not distinct
I saw some ClassCastExceptions internal to Zipline and investigated. We were attempting to serialize one type (`Foo`) with a serializer of another type (`Bar`). What the two types have in common is the application has a `Flow<Foo>` in one place, and a `Flow<Bar>` in another.

When inspecting the cache keys used for these, I would expect fully-qualified type names:
 * `app.cash.zipline.internal.bridge.FlowZiplineCollector<com.example.Foo>`
 * `app.cash.zipline.internal.bridge.FlowZiplineCollector<com.example.Bar>`

Instead, the generic types were omitted:
 * `app.cash.zipline.internal.bridge.FlowZiplineCollector<T>`

A good fix for this won’t reduce the hit rate of the cache.

swankjesse:
We were selecting cache keys at build-time, which resulted in multiple flows using the same serializer instances, despite having different runtime types.

This changes key construction to happen at runtime instead.

Closes: https://github.com/cashapp/zipline/issues/1039

### DIFF
--- a/zipline/src/hostTest/kotlin/app/cash/zipline/FlowTest.kt
+++ b/zipline/src/hostTest/kotlin/app/cash/zipline/FlowTest.kt
@@ -16,6 +16,8 @@
 package app.cash.zipline
 
 import app.cash.zipline.testing.newEndpointPair
+import assertk.assertThat
+import assertk.assertions.containsExactly
 import kotlin.test.Test
 import kotlin.test.assertContains
 import kotlin.test.assertEquals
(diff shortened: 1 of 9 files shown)


zipline/src/hostTest/kotlin/app/cash/zipline/FlowTest.kt
```diff
@@ -241,4 +244,28 @@ internal class FlowTest {
     assertEquals(setOf(), endpointA.serviceNames)
     assertEquals(setOf(), endpointB.serviceNames)
   }
+
+  /** https://github.com/cashapp/zipline/issues/1039 */
```
swankjesse, line 248: This test crashed on the old code attempting to use an Int serializer on a string.

JakeWharton (approved):

PR #1053, opened 2023-06-20
Wire up the ZiplineDumpTask to the Zipline Gradle plugin

swankjesse:


zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -74,6 +76,12 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
         it.dependsOn(kotlinWebpack)
       }
     }
+
+    target.tasks.withType(KotlinCompile::class.java) { kotlinCompile ->
+      if (kotlinCompile.name == "compileKotlinJvm") {
```
swankjesse, line 81: I’m not totally sure how to bootstrap this

JakeWharton: I suspect the check task will need to run on every target to ensure the interface is the same everywhere. Otherwise you could expect/actual and break yourself, or have completely different interfaces in different targets.

Not sure about the task to write it out, then. Maybe we just pick the first target? We're not guaranteed to have a JVM target (and doing things based on task name is generally discouraged anyway).

JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: I like that a lot. Added this advice as a note on the tracking issue.

PR #1058, opened 2023-06-23
Fix API checks for Android builds

swankjesse:


.github/workflows/build.yaml
```diff
@@ -57,10 +57,15 @@ jobs:
...
-      # Build unless this is a cross-compile.
-      - run: ./gradlew build --stacktrace
+      # Build unless this is a cross-compile (we can't test cross-compiled code)
+      - name: Build Zipline
+        run: ./gradlew build --stacktrace
         if: matrix.arch == 'amd64' || matrix.arch == 'x86_64'
 
+      - name: Build samples
+        run: ../gradlew check --stacktrace
+        working-directory: ./samples
```
JakeWharton: If you want
```suggestion
        run: ./gradlew -p samples check --stacktrace
```

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -77,11 +79,15 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
+    target.tasks.named("check").configure { checkTask ->
+      checkTask.dependsOn(ziplineApiCheck)
+    }
+    val ziplineApiDump = target.tasks.register("ziplineApiDump")
     target.tasks.withType(KotlinCompile::class.java) { kotlinCompile ->
-      if (kotlinCompile.name == "compileKotlinJvm") {
-        registerZiplineApiTask(target, kotlinCompile, ZiplineApiValidationTask.Mode.Check)
-        registerZiplineApiTask(target, kotlinCompile, ZiplineApiValidationTask.Mode.Dump)
-      }
+      if ("Test" in kotlinCompile.name) return@withType
```
JakeWharton, line 88: I think we can get at the main compilation task through
```kotlin
kotlinExtension.targets.all { target -> .. }
```
but that can be a follow-up. At the very least it reduces the risk of name-based task logic. Someone could, in theory, have "Test" in their target name.

JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: Yeah, lemme do

PR #1059, opened 2023-06-24
[WIP] feat: @ZiplineId

### LINKED ISSUE #1048: @ZiplineId
Lets build an annotation that can be applied to a member of a `ZiplineService` to override its ID. We’ll use this instead of the signature hash when it exists.

This will be useful to support renaming functions or their argument types without breaking backwards compatibility.

```
interface EchoService : ZiplineService {
  @ZiplineId("az1234sb")
  fun echo(request: EchoRequest): EchoResponse
}
```

epicadk:
closes: #1048


swankjesse:
I’ll take a look at this! But please in the future let us know before you proceed to build features? We’re not actively seeking external contributors.

epicadk:
Thanks, I think we can close, can't get the new services tests to run. It always throws method not found `serviceNamesArray` on `Zipline.create` for some reason.

PR #1063, opened 2023-06-28
Set JDK versions when running the compiler frontend

swankjesse:
This is necessary to load JPMS modules from the JDK.

Unfortunately there's no nice API to fetch the JDK version from the Gradle KotlinCompile task. It exposes an API to write this data but not to read it.


zipline-api-validator/src/main/kotlin/app/cash/zipline/api/validator/fir/KotlinFirLoader.kt
```diff
@@ -76,10 +79,10 @@ internal class KotlinFirLoader(
     configuration.put(CommonConfigurationKeys.MODULE_NAME, targetName)
     configuration.put(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY, messageCollector)
     configuration.put(CommonConfigurationKeys.USE_FIR, true)
+    configuration.put(JVMConfigurationKeys.JDK_HOME, javaHome)
+    configuration.put(JVMConfigurationKeys.JDK_RELEASE, jdkRelease)
```
swankjesse, line 83: @JakeWharton setting these two properties satisfies the removed TODO!

JakeWharton (approved):

PR #1080, opened 2023-07-18
Fix flaky SetTimeoutTest

### LINKED ISSUE #1073: SetTimeoutTest is flaky
```
> Task :zipline:iosX64Test

app.cash.zipline.SetTimeoutTest.happyPath[iosX64] FAILED
    kotlin.AssertionError at /opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Throwable.kt:25
```

Seeing this frequently. With 1hr+ CI builds this is a huge pain 😭

gnawf:
Attempts to fix #1073

Bit of a weird conundrum though, does `eventually` work correctly? Debugging it tells me so, but do I test the test code?

Any reason why this project don't use Kotest or so?

### DIFF
--- a/zipline/src/hostTest/kotlin/app/cash/zipline/SetTimeoutTest.kt
+++ b/zipline/src/hostTest/kotlin/app/cash/zipline/SetTimeoutTest.kt
@@ -21,6 +21,9 @@ import kotlin.test.AfterTest
 import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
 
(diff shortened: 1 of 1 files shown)


zipline/src/hostTest/kotlin/app/cash/zipline/SetTimeoutTest.kt
```diff
@@ -68,4 +73,25 @@ class SetTimeoutTest {
...
+      try {
+        assertCode()
+        return // Success, exit
+      } catch (e: AssertionError) {
+        lastException = e // Try again
+      }
+      delay(interval)
+    }
+
+    throw lastException ?: throw IllegalStateException("How could we reach here without an exception?")
```
colinrtwhite:
```suggestion
    throw lastException!!
```

zipline/src/hostTest/kotlin/app/cash/zipline/SetTimeoutTest.kt
```diff
@@ -51,8 +54,10 @@ class SetTimeoutTest {
     )
 
     assertEquals("hello", zipline.quickJs.evaluate("greeting"))
-    delay(200L)
-    assertEquals("goodbye", zipline.quickJs.evaluate("greeting"))
+
```
colinrtwhite, line 57:
```suggestion

    // TODO: Figure out why this requires retries.
```

colinrtwhite (commented):
@JakeWharton Looks like this works around the flaky test issue. Should we merge + add a TODO to figure out the root cause?

JakeWharton:
I'll look today. I wanted to give a shot at fixing the test first, presumably by moving to `runTest` so the virtual clock and dispatcher are shared.

JakeWharton:
> Any reason why this project don't use Kotest or so?

We generally think this library is not well designed and causes more problems than it solves.

PR #1081, opened 2023-07-20
Update Kotlin to 1.9.0.

colinrtwhite:
https://kotlinlang.org/docs/whatsnew19.html

Did a quick check with the world clock sample and everything seems to be working correctly.


JakeWharton (approved):

JakeWharton:
What caused that? Please tell me it wasn't Hermit...

colinrtwhite:
@JakeWharton Whenever I run the `kotlinUpgradeYarn` task it replaces it with our Artifactory references - not sure why. It happens with Redwood and other repos so it's not specific to Zipline.

zipline-api-validator/src/main/kotlin/app/cash/zipline/api/validator/fir/KotlinFirLoader.kt
```diff
@@ -104,10 +105,13 @@ internal class KotlinFirLoader(
 
     val input = ModuleCompilerInput(
       targetId = TargetId(JvmProtoBufUtil.DEFAULT_MODULE_NAME, targetName),
+      groupedSources = GroupedKtSources(
+        platformSources = files.map(::KtVirtualFileSourceFile),
+        commonSources = emptyList(),
+        sourcesByModuleName = emptyMap(),
+      ),
```
JakeWharton:
```suggestion
    val sourceFiles = files.map(::KtVirtualFileSourceFile).toSet()
    val input = ModuleCompilerInput(
      targetId = TargetId(JvmProtoBufUtil.DEFAULT_MODULE_NAME, targetName),
      groupedSources = GroupedKtSources(
        platformSources = sourceFiles,
        commonSources = emptyList(),
        sourcesByModuleName = mapOf(JvmProtoBufUtil.DEFAULT_MODULE_NAME to sourceFiles)
      ),
```

zipline-testing/src/appleMain/kotlin/app/cash/zipline/testing/CoroutineDispatchersApple.kt
```diff
@@ -16,13 +16,15 @@
...
 import kotlin.coroutines.CoroutineContext
+import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.cinterop.convert
 import kotlinx.coroutines.CloseableCoroutineDispatcher
 import kotlinx.coroutines.Runnable
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.runBlocking
 import platform.Foundation.NSThread
 
+@OptIn(ExperimentalForeignApi::class)
```
JakeWharton, line 28: This actual opts-in, but the expect and the other actuals propagate the annotation. Should they all opt-in instead?

zipline-gradle-plugin/src/test/projects/basic/kotlin-js-store/yarn.lock
JakeWharton, line 1: I think we should `.gitignore` these and set
```kotlin
plugins.withType(org.jetbrains.kotlin.gradle.targets.js.yarn.YarnPlugin::class.java) {
    the<YarnRootExtension>().yarnLockAutoReplace = true
}
```
per https://kotlinlang.org/docs/whatsnew18.html#new-settings-for-reporting-that-yarn-lock-has-been-updated

This way we never have to update them since they're test fixtures.

zipline-testing/src/appleMain/kotlin/app/cash/zipline/testing/CoroutineDispatchersApple.kt
colinrtwhite: It's actually separate experimental annotations: the expects propagate `@ExperimentalCoroutinesApi` as it exposes `CloseableCoroutineDispatcher`, but this opts into `@ExperimentalForeignApi`.

colinrtwhite:
The last failing test is a weird one:

```
app.cash.zipline.SetTimeoutTest.happyPath[macosX64] FAILED
    kotlin.AssertionError at null:-1
```

it only fails on Darwin platforms on CI. It passes locally on my M1 machine.

EDIT: Nevermind the test is [known to be flaky](https://github.com/cashapp/zipline/issues/1073).

colinrtwhite:
I think this PR is good to merge once we can also update Redwood.

JakeWharton:
Weird failure. Do we need to downgrade the JDK to one that's more tolerant of Kotlin's reflective shenanigans?

JakeWharton:
By fixing one flake did I break a bunch of others? I can fix those too, I guess...

colinrtwhite:
@JakeWharton Looks like your fix on https://github.com/cashapp/zipline/pull/1083 fixes the remaining failing tests.

JakeWharton:
Excellent! The console one was proper broken so I'm surprised the timeout one was the one which failed all the time.

JakeWharton:
Let's try whack-a-mole downgrading the JDK version by one until one works, I guess?

colinrtwhite:
@JakeWharton I think we're already using JDK 11 (via `jvmToolchain(11)`), though, which is the minimum for AGP.

JakeWharton:
JVM toolchains only apply to tasks which fork a JVM, like `JavaCompile` or `Test`. The JVM in which Gradle itself is running is whatever is in `JAVA_HOME` which looks to be 20. I don't see any YouTrack issues about it, though, which is weird. Let me pull and look locally.

JakeWharton:
I cannot reproduce locally. Also, I misread the underlying problem it's actually this:
```
 Execution failed for task ':zipline:commonizeCInterop'.
> java.io.FileNotFoundException: /Users/runner/work/zipline/zipline/zipline/build/kotlinTransformedCInteropMetadataLibraries/nativeMain/.zipline-nativeMain.cinteropLibraries (No such file or directory)
```

colinrtwhite:
Hmm JDK 17 also fails. I'll try JDK 11. Could it also be a cklib issue given it's related to `cinteropLibraries`?

EDIT: Actually AGP now requires JDK 17 so that won't work.

EDIT2: Seems like this might work: https://github.com/Kotlin/dokka/issues/2977#issuecomment-1571550626. It did!

JakeWharton:
Yeah the JDK issue was a red herring caused by me reading the failure on mobile.

Nice find on the Dokka thing, though. Also, Dokka 🤦‍♂️

JakeWharton (approved):
Let's do this!

IgnatBeresnev:
> Also, Dokka 🤦‍♂️

we're not doing it on purpose, I swear :smiling_face_with_tear:

PR #1083, opened 2023-07-25
Migrate more tests to `runTest`

JakeWharton:


kotlin-js-store/yarn.lock
JakeWharton, line 1: Don't know why I had to update this, but I did

zipline/src/jvmTest/kotlin/app/cash/zipline/ConsoleTest.kt
```diff
@@ -20,29 +20,28 @@ import assertk.assertThat
...
 
 class ConsoleTest {
-  @Rule @JvmField
-  val ziplineTestRule = ZiplineTestRule()
-  private val dispatcher = ziplineTestRule.dispatcher
+  private val dispatcher = StandardTestDispatcher()
   private val zipline = Zipline.create(dispatcher)
 
-  private val logRecords = LinkedBlockingDeque<LogRecord>()
+  private val logRecords = Channel<LogRecord>(UNLIMITED)
```
JakeWharton, line 41: This is half the problem. Calling `take()` blocks the dispatcher and prevents Zipline from posting a coroutine to append to the queue. A channel allows the test to suspend while waiting.

zipline/src/jvmTest/kotlin/app/cash/zipline/ConsoleTest.kt
```diff
@@ -124,14 +139,18 @@ class ConsoleTest {
    * Note that this test is checking our expected behavior, but our behavior falls short of what
    * browsers implement. In particular, we don't do string replacement for `%s`, `%d`, etc.
    */
-  @Test fun logWithArguments() = runBlocking(dispatcher) {
+  @Test fun logWithArguments() = runTest(dispatcher) {
     zipline.quickJs.evaluate("testing.app.cash.zipline.testing.consoleLogWithArguments()")
-    assertEquals("INFO: this message for %s is a %d out of %d Jesse 8 10", takeLogMessage())
+
+    val record = logRecords.receive()
```
JakeWharton, line 145: This is the other half of the problem. By using `takeLogMessage()` before, an empty deque would result in returning null rather than waiting. By switching to `receive()` we now suspend when we expect an item.

colinrtwhite (approved):

PR #1101, opened 2023-08-15
Use min-sdk from toml for world-clock.

Shabinder:
- Fix for compile-sdk being used as min-sdk in world-clock android sample

### DIFF
--- a/samples/world-clock/android/build.gradle.kts
+++ b/samples/world-clock/android/build.gradle.kts
@@ -11,7 +11,7 @@ android {
 
   defaultConfig {
     applicationId = "com.example.zipline.worldclock"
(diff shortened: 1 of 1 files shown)


Shabinder:
Have bumped the min-sdk for android target, because:
 - Deps such and Android.Material require minSdk 21 as target.
 - since above pre-requisite 21 as minSdk, didn't needed to enable multi-dex explicitly.

JakeWharton:
I have no idea why we even try support down to 18. That was probably just inherited from the Duktape days. Next minor release I'll try to remember to bump everything to at least 21.

JakeWharton (approved):

PR #1131, opened 2023-09-18
Configure a 6 MiB stack size by default

swankjesse:
Previously this was unbounded, which meant that Zipline would scribble over the heap if the stack size was exceeded.


swankjesse:
We use an 8 MiB thread at the callsite on iOS in Redwood.
https://github.com/cashapp/redwood/blob/trunk/redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/IosTreehouseDispatchers.kt#L59

JakeWharton (approved):

veyndan (approved):
Sweet!

PR #1146, opened 2023-10-16
Serialize `null` when passed to a `ZiplineService` function with a nullable `@Contextual` parameter

veyndan:


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
```diff
@@ -235,7 +237,7 @@ internal class ArgsListSerializer(
     encoder.encodeStructure(descriptor) {
       for (i in serializers.indices) {
         @Suppress("UNCHECKED_CAST") // We don't have a type argument T for each parameter.
-        encodeSerializableElement(descriptor, i, serializers[i] as KSerializer<Any?>, value[i])
+        encodeNullableSerializableElement(descriptor, i, serializers[i] as KSerializer<Any?>, value[i])
```
veyndan, line 240: An alternative approach is to edit `AdapterGenerator` such that nullable parameters have their `KSerializer`'s made nullable via https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.builtins/nullable.html. Went with the approach in the PR as it was quicker/easier and is comparable to the other approach.

swankjesse (approved):

swankjesse:
Nice! Thanks for fixing.

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/calls.kt
JakeWharton: This needs added to the README as an additional unstable API that we rely on. Is the `nullable` API stable? It may be worth the extra effort to reduce our reliance on unstable API in an otherwise stable library.

veyndan: > Is the `nullable` API stable?

Yep!

I'll have a go at it then in a follow up.

veyndan: Started a PR at https://github.com/cashapp/zipline/pull/1149

PR #1149, opened 2023-10-17
Replace `[en|de]codeNullableSerializableElement` with `nullable`

veyndan:
`encodeNullableSerializableElement` and `decodeNullableSerializableElement` are both experimental APIs, whereas `nullable` isn't.

### DIFF
--- a/zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/BridgedInterface.kt
+++ b/zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/BridgedInterface.kt
@@ -31,12 +31,15 @@ import org.jetbrains.kotlin.ir.declarations.IrValueParameter
 import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
+import org.jetbrains.kotlin.ir.symbols.IrPropertySymbol
 import org.jetbrains.kotlin.ir.symbols.IrSimpleFunctionSymbol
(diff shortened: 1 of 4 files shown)


zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/BridgedInterface.kt
```diff
@@ -234,6 +234,8 @@ internal class BridgedInterface(
           )
           putValueArgument(1, parameterList)
         }
+        ziplineApis.nullableSerializer // TODO Invoke this on [contextualSerializerExpression] if the [type] is nullable.
```
veyndan: @JakeWharton / @swankjesse Any ideas on how to do this? I've been searching for some API that'd take an `IrPropertySymbol` with `contextualSerializerExpression` as a receiver, but I'm a bit lost. I failed at finding some prior art of this being done within Zipline and online.

JakeWharton: Here's how the built-in kotlinx.serialization plugin does it: https://github.com/JetBrains/kotlin/blob/ec79d8cd2711e78883c77f2e6c4ec377398aecb9/plugins/kotlinx-serialization/kotlinx-serialization.backend/src/org/jetbrains/kotlinx/serialization/compiler/backend/ir/BaseIrGenerator.kt#L359-L378

veyndan: Oh damn, that's perfect. Thanks!

JakeWharton (approved):

PR #1153, opened 2023-10-25
Support arbitrary metadata in the manifest

swankjesse:


JakeWharton (approved):
Binary dump tool is a pain in the ass.

PR #1162, opened 2023-11-04
Hack the klib manifest version to work around KT-62515

swankjesse:
This is a big regression in Kotlin 1.9.20.


swankjesse:
https://youtrack.jetbrains.com/issue/KT-62515/Interop-klib-of-concurrent-version-is-not-accepted-when-building-dependent-project-The-library-versions-dont-match

swankjesse:
(I manually confirmed that this fixes the problem for downstream builds)

JakeWharton (approved):

PR #1166, opened 2023-11-13
Use Kotlin/Multiplatform's hierarchy template

swankjesse:
Much nicer.

### DIFF
--- a/zipline-loader-testing/build.gradle.kts
+++ b/zipline-loader-testing/build.gradle.kts
@@ -21,8 +21,6 @@ kotlin {
   tvosSimulatorArm64()
   tvosX64()
 
-  applyDefaultHierarchyTemplate()
-
   sourceSets {
     val commonMain by getting {
       dependencies {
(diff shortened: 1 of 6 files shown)


zipline-loader/build.gradle.kts
```diff
@@ -85,9 +77,6 @@ kotlin {
         implementation(libs.turbine)
       }
     }
-    val androidInstrumentedTest by getting {
```
swankjesse: This was forbidden!

zipline-loader-testing/build.gradle.kts
```diff
@@ -22,6 +21,8 @@ kotlin {
   tvosSimulatorArm64()
   tvosX64()
 
+  applyDefaultHierarchyTemplate()
+
```
JakeWharton:
```suggestion
```
This project doesn't have any custom source sets. Shouldn't be needed explicitly.

zipline-profiler/build.gradle.kts
```diff
@@ -22,6 +21,8 @@ kotlin {
   tvosSimulatorArm64()
   tvosX64()
 
+  applyDefaultHierarchyTemplate()
+
```
JakeWharton:
```suggestion
```
Don't think it's needed here either

JakeWharton (approved):

PR #1175, opened 2023-11-24
Don't mangle production builds

swankjesse:
We want to remove dead code, but not much more. There's not much benefit to minifying code before we send it into the QuickJS compiler.


swankjesse:
Turns out there’s a cost to this!

For a non-trivial application, this causes a 42% increase in the download size.

```
                       NO DEBUG DATA    DEBUG DATA    DELTA
        bytecode size:     2,856,719     3,871,128     +35%
gzipped bytecode size:       941,474     1,346,230     +42%
```

I’ll see if I can tighten that up...

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/WriteWebpackConfigTask.kt
```diff
@@ -0,0 +1,70 @@
...
+      |      },
+      |      mangle: false,
+      |      format: {
+      |        beautify: true,
+      |        braces: true,
+      |      }
+      |    },
+      |  }),
+      |];
+      """.trimMargin(),
```
JakeWharton, line 65:
```suggestion
      |""".trimMargin(),
```
Trailing newline always a good idea

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/WriteWebpackConfigTask.kt
```diff
@@ -0,0 +1,70 @@
...
+    )
+  }
+
+  companion object {
+    val generatedFilePath = "webpack.config.d/generated-zipline-webpack-config.js"
```
JakeWharton, line 70: I thought these had to be prefixed with a number like `23-whevs.js`. Maybe that's just a convention for ordering.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -67,17 +67,21 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
     }
 
     target.tasks.withType(KotlinWebpack::class.java) { kotlinWebpack ->
```
JakeWharton, line 69: Unrelated to this change, but I'm pretty sure this causes all tasks to be configured. Gradle docs will have the lazy pattern. Separate PR for that...

zipline-testing/build.gradle.kts
```diff
@@ -68,7 +68,6 @@ kotlin {
 }
 
 tasks {
-  // TODO upstream this to ZiplinePlugin
   withType<KotlinJsCompile>().all {
     kotlinOptions {
       freeCompilerArgs += listOf("-Xir-per-module")
```
JakeWharton, line 73: Delete it all. This is the default!

JakeWharton (approved):

swankjesse:
More analysis of file size costs:

```
                                    .zipline size        gzipped size
A NO DEBUG DATA                    2,856,719             941,474
B MANGLE CLASS + FUNCTION NAMES    3,400,044 (+19%)    1,193,241 (+27%)
C OMIT INSTRUCTION LINE NUMBERS    3,690,174 (+29%)    1,218,497 (+29%)
D TRUNCATE FILE PATH               3,642,791 (+28%)    1,338,461 (+42%)
E (B + C)                          3,218,986 (+13%)    1,064,318 (+13%)
F FULL DEBUG DATA                  3,871,128 (+35%)    1,346,230 (+42%)
```

Item B changes the Terser config to enable mangling, which will prevent true function names from occurring in stacktraces:
```
 new TerserPlugin({
   terserOptions: {
     compress: {
       sequences: false,
     },
     mangle: {
     },
     format: {
       beautify: true,
       braces: true,
     }
   },
 }),
```

Item C clears the pc2line entry from the bytecode, which will prevent true line numbers from occurring in stacktraces.

I think our best option for production use is B+C which still gives us bad stacktraces (file names only, no line numbers or function names), but with dramatically lower cost.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/WriteWebpackConfigTask.kt
swankjesse: ktlint really has no taste.

```
> The following files had format violations:
      zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/WriteWebpackConfigTask.kt
          @@ -60,7 +60,8 @@
           ······|····},
           ······|··}),
           ······|];
          -······|""".trimMargin(),
          +······|
          +""".trimMargin(),
           ····)
           ··}
```

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/WriteWebpackConfigTask.kt
swankjesse: Yep. I don’t expect that to be useful for this use.

zipline-testing/build.gradle.kts
swankjesse: Follow-up coming up.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: Follow-up coming up.

JakeWharton:
The source map doesn't/can't help at all here?

JakeWharton:
Or what if we rewrote/appended to the R8 mapping file so the short, mangled names would get deobfuscated by crash reporting tools entirely on the server-side?

JakeWharton:
Can be future investigations

swankjesse:
Yep, we could do it server-side, though we’ll introduce some complexity to support that. The biggest challenge I see is that our QuickJS stack frames don’t consistently have a class+function, and I assume that’s how deobfuscators are keyed.

JakeWharton:
They are, but since R8 might put everything into a single class we only need to use a fake one to get that working.

zipline-testing/build.gradle.kts
swankjesse: https://github.com/cashapp/zipline/pull/1179

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: https://github.com/cashapp/zipline/issues/1180

PR #1193, opened 2023-12-13
Add FreshnessChecker into zipline loading process

jingwei99:
This PR

1) Added a new parameter to `ZiplineLoader#load`, and `ZiplineLoader#loadOnce`
2) Changed behaviors of `ZiplineLoader#load`, and `ZiplineLoader#loadOnce`
3) `ZiplineLoader#load`, and `ZiplineLoader#loadOnce` with existing parameter list are marked as `@Deprecated`

Follow up changes: 
1) add a informational eventListener.applicationLoadSkipped for reason
2) add retry when fetching from network fails

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/FreshnessChecker.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/FreshnessChecker.kt
@@ -16,17 +16,30 @@
 package app.cash.zipline.loader
 
 import app.cash.zipline.ZiplineManifest
-import kotlin.time.Duration
-import kotlin.time.Duration.Companion.hours
 
-
-/** Checks if a given ZiplineManifest is considered fresh, therefore safe to be served. */
+/**
+ * Checks if a given ZiplineManifest is considered fresh.
+ *
+ * For local development that requires continuous loading (or hot loading), supply a
+ * FreshnessChecker that always returns false.
+ */
 interface FreshnessChecker {
 
-  /** Returns true if the manifest is considered fresh. */
+  /**
+   * Decides whether the [manifest] is eligible to be used.
+   *
+   * Returns true to launch the [manifest] immediately; false to download a fresh
+   * ZiplineManifest and launch that.
+   */
   fun isFresh(
     manifest: ZiplineManifest,
     freshAtEpochMs: Long,
-    //shelfLife: Duration = 168.hours
   ): Boolean
+}
+
+/** A FreshnessChecker that always returns true. */
+object DefaultFreshnessCheckerNotFresh : FreshnessChecker {
+  override fun isFresh(manifest: ZiplineManifest, freshAtEpochMs: Long): Boolean {
+    return false
+  }
 }
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/FreshnessChecker.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/FreshnessChecker.kt
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2023 Cash App
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.loader
+
+import app.cash.zipline.ZiplineManifest
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.hours
+
+
+/** Checks if a given ZiplineManifest is considered fresh, therefore safe to be served. */
+interface FreshnessChecker {
+
+  /** Returns true if the manifest is considered fresh. */
+  fun isFresh(
+    manifest: ZiplineManifest,
+    freshAtEpochMs: Long,
+    //shelfLife: Duration = 168.hours
+  ): Boolean
+}
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -157,20 +157,20 @@ class ZiplineLoader internal constructor(
    * network is unreachable: returns null, and emits a LoadResult.Failure
    *
    * @param manifestUrlFlow a flow that emits whenever by the downstream service.
-   * @param freshnessChecker checks if a cached ZiplineManifest is considered fresh.
+   * @param freshnessChecker checks if a cached ZiplineManifest is considered fresh. Defaulted to
+   * always return false, which means always load from network when [manifestUrlFlow] emits.
    */
   fun load(
     applicationName: String,
-    freshnessChecker: FreshnessChecker,
+    freshnessChecker: FreshnessChecker = DefaultFreshnessCheckerNotFresh,
     manifestUrlFlow: Flow<String>,
     serializersModule: SerializersModule = EmptySerializersModule(),
     initializer: (Zipline) -> Unit = {},
   ): Flow<LoadResult> {
     return channelFlow {
       var previousManifest: ZiplineManifest? = null
-      val now = nowEpochMs()
       val localManifest = loadFromLocal(
-        now,
+        nowEpochMs(),
         applicationName,
         freshnessChecker,
         serializersModule,
@@ -188,7 +188,7 @@ class ZiplineLoader internal constructor(
         } else {
           val loadedFromNetwork = loadFromNetwork(
             previousManifest,
-            now,
+            nowEpochMs(),
             applicationName,
             manifestUrl,
             serializersModule,
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -147,62 +147,63 @@ class ZiplineLoader internal constructor(
   private val moduleFetchers = listOfNotNull(embeddedFetcher, cachingFetcher ?: httpFetcher)
 
   /**
-   * Loads code from [manifestUrlFlow] each time it emits, skipping loads if the code to load is
-   *    the same as what's already loaded.
+   * Each time when [manifestUrlFlow] emits: Emits a fresh ZiplineManifest and loads Zipline with
+   * the fresh ZiplineManifest.
    *
-   * If the network is unreachable and there is local code newer, then this
-   *   will load local code, either from the embedded directory or the cache.
+   * If the local ZiplineManifest is fresh, emits it and load Zipline without downloading from the
+   * network.
    *
-   * @param manifestUrlFlow a flow that should emit each time a load should be attempted. This
-   *     may emit periodically to trigger polling. It should also emit for loading triggers like
-   *     app launch, app foregrounding, and network connectivity changed.
+   * If the local ZiplineManifest is not fresh, download a new ZiplineManifest from network.
+   *
+   * Newly downloaded ZiplineManifest is always considered fresh.
+   *
+   * If the network is unreachable : returns null, and emits a LoadResult.Failure
+   *
+   * @param manifestUrlFlow a flow that emits whenever a copy of ZiplineManifest is requested by the
+   *  downstream service.
+   * @param freshnessChecker checks if a cached ZiplineManifest is considered fresh.
    */
   fun load(
     applicationName: String,
+    freshnessChecker: FreshnessChecker,
     manifestUrlFlow: Flow<String>,
     serializersModule: SerializersModule = EmptySerializersModule(),
     initializer: (Zipline) -> Unit = {},
   ): Flow<LoadResult> {
     return channelFlow {
-      var isFirstLoad = true
       var previousManifest: ZiplineManifest? = null
-
       manifestUrlFlow.collect { manifestUrl ->
-        // Each time a manifest URL is emitted, download and initialize a Zipline for that URL.
-        //  - skip if the manifest hasn't changed from the previous load.
-        //  - pin the application if the load succeeded; unpin if it failed.
         val now = nowEpochMs()
-
-        val loadedFromNetwork = loadFromNetwork(
-          previousManifest,
+        val localManifest = loadFromLocal(
           now,
           applicationName,
-          manifestUrl,
+          freshnessChecker,
           serializersModule,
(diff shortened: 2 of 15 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -147,62 +147,63 @@ class ZiplineLoader internal constructor(
...
-   *     app launch, app foregrounding, and network connectivity changed.
+   * If the local ZiplineManifest is not fresh, download a new ZiplineManifest from network.
+   *
+   * Newly downloaded ZiplineManifest is always considered fresh.
+   *
+   * If the network is unreachable : returns null, and emits a LoadResult.Failure
+   *
+   * @param manifestUrlFlow a flow that emits whenever a copy of ZiplineManifest is requested by the
+   *  downstream service.
+   * @param freshnessChecker checks if a cached ZiplineManifest is considered fresh.
```
jingwei99: first draft, if it looks alright, more test cases will be added 

also a few thoughts I have while working on this: 

1) for local development (of Treehouse), it'll need the continuous loading, one way (which is implied by this change) to support continuous loading is to pass in a `freshnessChecker` that always returns false 

2) related to 1), right now with the current implementation, it's implied that a new download from the network is always considered as fresh and therefore loaded (into Zipline), the question is: is there a danger/potential harm for this assumption? I thought over a few different scenarios, it seems a safe assumption (meaning downstream layers, like Treehouse, would be mitigating with error handling, if the version served by server is not the version intended to be served)

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/FreshnessChecker.kt
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package app.cash.zipline.loader
+
+import app.cash.zipline.ZiplineManifest
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.hours
+
+
+/** Checks if a given ZiplineManifest is considered fresh, therefore safe to be served. */
```
swankjesse: This is great.

I’m tempted to suggest another word for _safe_ . . . . maybe _eligible_ ? Would you consider putting a short discussion of what this is for?

> Decides whether cached code is eligible to be used.
> Return true to launch the cached code immediately; false to
> download fresh code and launch that.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -147,62 +147,63 @@ class ZiplineLoader internal constructor(
...
    */
   fun load(
     applicationName: String,
+    freshnessChecker: FreshnessChecker,
```
swankjesse: This needs a new `load()` overload due to backwards compatibility

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
swankjesse: 1. Great idea. 
2. Lets document that FreshnessChecker is only used to check freshness of cached things. Downloaded things are fresh by definition.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -147,62 +147,63 @@ class ZiplineLoader internal constructor(
...
-
       manifestUrlFlow.collect { manifestUrl ->
-        // Each time a manifest URL is emitted, download and initialize a Zipline for that URL.
-        //  - skip if the manifest hasn't changed from the previous load.
-        //  - pin the application if the load succeeded; unpin if it failed.
         val now = nowEpochMs()
-
-        val loadedFromNetwork = loadFromNetwork(
-          previousManifest,
+        val localManifest = loadFromLocal(
```
swankjesse: I’ve got a weird idea . . . what if we hoist this out of the `collect()` call? Ie. we load from the cache before the flow emits for the first time?

In theory if we load from the cache we don’t ever need to collect ?!

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -257,57 +258,68 @@ class ZiplineLoader internal constructor(
...
-      send(LoadResult.Success(zipline, loadedManifest.manifest, loadedManifest.freshAtEpochMs))
-      return loadedManifest.manifest
-    } catch (e: CancellationException) {
-      // If emit() threw a CancellationException, consider that emit to be successful.
-      // That's 'cause loadOnce() accepts an element and then immediately cancels the flow.
-      throw e
-    } catch (e: Exception) {
-      send(LoadResult.Failure(e))
-      eventListener.applicationLoadFailed(applicationName, null, e, startValue)
+    if (freshnessChecker.isFresh(loadedManifest.manifest, loadedManifest.freshAtEpochMs)) {
```
swankjesse: Invert this `if()` to return early?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -257,57 +258,68 @@ class ZiplineLoader internal constructor(
...
+        // If emit() threw a CancellationException, consider that emit to be successful.
+        // That's 'cause loadOnce() accepts an element and then immediately cancels the flow.
+        throw e
+      } catch (e: Exception) {
+        send(LoadResult.Failure(e))
+        eventListener.applicationLoadFailed(applicationName, null, e, startValue)
+        return null
+      }
+    } else {
+      // TODO: consider to add a informational eventListener.applicationLoadSkipped for reason
```
swankjesse: Ooooh I like this idea a lot. Should help us to refine our various tuning parameters.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -257,57 +258,68 @@ class ZiplineLoader internal constructor(
...
       return null
     }
   }
 
   suspend fun loadOnce(
     applicationName: String,
+    freshnessChecker: FreshnessChecker,
```
swankjesse, line 346: No signature changes in existing APIs. Gotta add new APIs

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/LoaderTester.kt
```diff
@@ -310,7 +312,24 @@ class LoaderTester(
...
+      applicationName = applicationName,
+      freshnessChecker = FakeFreshnessCheckerFresh(),
+      manifestUrlFlow = flowOf(manifestUrl),
+      initializer = initializer,
+    )
     zipline = (results.last() as LoadResult.Success).zipline
   }
 }
+
+class FakeFreshnessCheckerFresh : FreshnessChecker {
```
swankjesse: class -> object

swankjesse (commented):
Love it.

swankjesse (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
jingwei99: so I thought about it, and here's my thought process: if we need to satisfy the existing expectation from downstream, which is: emits whenever a manifestUrl emits, then we need to keep it here. If we don't need to satisfy that expectation, then we can update the behavior here. 

I didn't get a full picture (comprehensive) answer within myself on: if it's a good/helpful expectation from the downstream to get a new emission whenever the URL flow emits, one use case I thought about was hot loading for local development (but there are for sure other ways to support that, doesn't have to be relying on this behavior here)

jingwei99:
~not sure if the CI failure is a flake or it's real, task `:zipline-loader:iosX64Test` passes locally~

update, it's not a flake, it happens on CI for specific OS: `tvosX64`

```
app.cash.zipline.loader.LoaderEventsTest.manifestDownloadFailsWithoutFreshLocal[tvosX64] FAILED
    kotlin.AssertionError at /opt/buildAgent/work/2fed3917837e7e79/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:14
```

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -147,62 +147,96 @@ class ZiplineLoader internal constructor(
...
+      val now = nowEpochMs()
+      val localManifest = loadFromLocal(
+        now,
+        applicationName,
+        freshnessChecker,
+        serializersModule,
+        initializer,
+      )
+      if (localManifest != null) {
+        previousManifest = localManifest
```
swankjesse, line 180: Unnecessary now, because it returns?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -147,62 +147,96 @@ class ZiplineLoader internal constructor(
...
-        if (loadedFromNetwork != null) {
-          previousManifest = loadedFromNetwork
-        }
-
-        // If network loading failed (due to network error, or bad code), attempt to load from the
-        // cache or embedded file system. This doesn't update pins!
-        if (previousManifest == null && isFirstLoad) {
-          isFirstLoad = false
-          val loadedFromLocal = loadFromLocal(
+        if (previousManifest != null &&
```
swankjesse, line 184: Unreachable now?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -147,62 +147,96 @@ class ZiplineLoader internal constructor(
...
+    freshnessChecker: FreshnessChecker,
     manifestUrlFlow: Flow<String>,
     serializersModule: SerializersModule = EmptySerializersModule(),
     initializer: (Zipline) -> Unit = {},
   ): Flow<LoadResult> {
     return channelFlow {
-      var isFirstLoad = true
       var previousManifest: ZiplineManifest? = null
-
+      val now = nowEpochMs()
```
swankjesse: Move this into `collect` so it gets the timestamp when the manifestUrl was emitted, not the timestamp that load() was called?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
swankjesse: Should we default this?

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/LoaderEventsTest.kt
```diff
@@ -95,13 +116,15 @@ class LoaderEventsTest {
 
   @Test
   fun loadAlreadyCached() = runBlocking {
-    assertEquals("apple", tester.success("red", "apple"))
+    // TODO
```
swankjesse: ?

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/LoaderEventsTest.kt
```diff
@@ -119,26 +142,14 @@ class LoaderEventsTest {
   }
 
   @Test
-  fun manifestDownloadFails() = runBlocking {
-    tester.seedEmbedded("red", "firetruck")
-    assertEquals("firetruck", tester.failureManifestFetchFails("red"))
+  fun manifestDownloadFailsWithoutFreshLocal() = runBlocking {
```
swankjesse, line 145: I like how much less work this is doing

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/LoaderTester.kt
```diff
@@ -130,6 +142,44 @@ class LoaderTester(
...
       "$baseUrl/$applicationName/$seed.zipline" to ziplineFileByteString,
     )
+    return loader.load(
+      applicationName = applicationName,
+      manifestUrlFlow = List(count) { manifestUrl }.asFlow(),
+      freshnessChecker = freshnessChecker,
+    )
+  }
+
+  // Used to test the deprecated load function, and should be deleted when the deprecated function
```
swankjesse, line 152: Nice

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/LoaderTester.kt
```diff
@@ -227,6 +289,18 @@ class LoaderTester(
     )
   }
 
+  suspend fun failureCacheNotFresh(applicationName: String): LoadResult {
```
swankjesse, line 292: YASS!

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/ZiplineLoaderTest.kt
```diff
@@ -89,13 +89,97 @@ class ZiplineLoaderTest {
   }
 
   @Test
-  fun loadManifestFromUrl() = runBlocking {
+  fun loadsOnceWhenLocalNotFreshWithOneManifestUrlEmission() = runBlocking {
```
swankjesse, line 92: Nice

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/ZiplineLoaderTest.kt
```diff
@@ -89,13 +89,97 @@ class ZiplineLoaderTest {
...
+    ).test {
+      assertEquals(
+        "firetruck",
+        (
+        (awaitItem() as LoadResult.Success).zipline.quickJs.evaluate(
+          "globalThis.log",
+          "assert.js",
+        ) as String
+        ).removeSuffix(" loaded\n"),
+      )
```
swankjesse:
```suggestion
      assertEquals(
        "firetruck loaded\n",
        (awaitItem() as LoadResult.Success).zipline.quickJs.evaluate(
          "globalThis.log",
          "assert.js",
        )
      )
```

swankjesse (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
jingwei99: took several looks, it is reachable, `manifestUrlFlow` might emit several times

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/LoaderEventsTest.kt
jingwei99: forgot to delete, was a note to myself, deleted.

PR #1199, opened 2023-12-15
Add applicationLoadSkipped to EventListener

jingwei99:
Follow up from https://github.com/cashapp/zipline/pull/1193

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -308,7 +308,7 @@ class ZiplineLoader internal constructor(
       ?: return null
     val startValue = eventListener.applicationLoadStart(applicationName, null)
     if (!freshnessChecker.isFresh(loadedManifest.manifest, loadedManifest.freshAtEpochMs)) {
-      // TODO: add an informational eventListener.applicationLoadSkipped for reason
+      eventListener.applicationLoadSkippedNotFresh(applicationName, null, startValue)
       return null
     } else {
       try {
--- a/zipline/src/hostMain/kotlin/app/cash/zipline/EventListener.kt
+++ b/zipline/src/hostMain/kotlin/app/cash/zipline/EventListener.kt
@@ -86,6 +86,19 @@ abstract class EventListener {
   }
 
   /**
+   * Invoked when an application load was skipped because the code is stale.
+   *
+   * @param startValue the value returned by [applicationLoadStart] for the start of this call. This
+   *   is null unless [applicationLoadStart] is overridden to return something else.
+   */
+  open fun applicationLoadSkippedNotFresh(
+    applicationName: String,
+    manifestUrl: String?,
+    startValue: Any?,
+  ) {
+  }
+
+  /**
    * Invoked when an application load succeeds.
    *
    * @param startValue the value returned by [applicationLoadStart] for the start of this call. This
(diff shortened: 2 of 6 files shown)


zipline/src/hostMain/kotlin/app/cash/zipline/EventListener.kt
```diff
@@ -72,6 +72,21 @@ abstract class EventListener {
...
+   * Invoked when an application load was skipped because the code is unchanged.
+   *
+   * @param info additional info on the load-skipping
+   * @param startValue the value returned by [applicationLoadStart] for the start of this call. This
+   *   is null unless [applicationLoadStart] is overridden to return something else.
+   */
+  open fun applicationLoadSkipped(
+    applicationName: String,
+    manifestUrl: String?,
+    info: String,
```
swankjesse: Is there a more typesafe way to do this?

zipline/src/hostMain/kotlin/app/cash/zipline/EventListener.kt
```diff
@@ -72,6 +72,21 @@ abstract class EventListener {
     return null
   }
 
+  /**
+   * Invoked when an application load was skipped because the code is unchanged.
```
swankjesse: Because the code is unchanged? Or because the code is stale?

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -308,7 +308,12 @@ class ZiplineLoader internal constructor(
       ?: return null
     val startValue = eventListener.applicationLoadStart(applicationName, null)
     if (!freshnessChecker.isFresh(loadedManifest.manifest, loadedManifest.freshAtEpochMs)) {
-      // TODO: add an informational eventListener.applicationLoadSkipped for reason
+      eventListener.applicationLoadSkipped(
```
swankjesse: Maybe rename this to `applicationLoadSkippedNotFresh` and delete the string?

swankjesse (approved):

zipline/src/hostMain/kotlin/app/cash/zipline/EventListener.kt
jingwei99: ah it's because it's stale!

zipline/src/hostMain/kotlin/app/cash/zipline/EventListener.kt
jingwei99: definitely, what are you thinking about? (there are different ways!)

jingwei99: ok, realized this: https://github.com/cashapp/zipline/pull/1199#discussion_r1430487746

zipline/src/hostMain/kotlin/app/cash/zipline/EventListener.kt
```diff
@@ -85,6 +85,19 @@ abstract class EventListener {
...
 
+  /**
+   * Invoked when an application load was skipped because the code is stale.
+   *
+   * @param startValue the value returned by [applicationLoadStart] for the start of this call. This
+   *   is null unless [applicationLoadStart] is overridden to return something else.
+   */
+  open fun applicationLoadSkippedNotFresh(
+    applicationName: String,
+    manifestUrl: String?,
```
JakeWharton, line 96: Do we call out what null means for this anywhere? I see it's nullable on other functions too without docs. I'm guessing it's for local?

jingwei99: yea this is for local, I was a bit conflicted when I noticed the null, when I was working on this PR: 

1) what if later we'll use this API somewhere else 
2) but where? or later when? 

also because I'm not fully aware of all the implications of changing an API (is it costly, or it's not), so I left it this way 

if changing API (later adding a function in event listener or so) is not a big deal, I'd be happy if the decision is to remove this field, given what we know currently about use case.

PR #1211, opened 2024-01-03
Replace kapt usage with ksp.

colinrtwhite:


JakeWharton (approved):
Seems like a wholly lateral move. Ship it!

PR #1212, opened 2024-01-04
Support K2.

colinrtwhite:
~Took a quick pass at supporting K2. It might be possible to support both K2 and K1 if that's something we want to do. Else, we could wait to make these changes until K2 is released.~

Updated so we can support both K1 and K2. This PR also contains a couple smaller changes that aren't strictly necessary unless we're also compiling Zipline with K2. Can revert if preferred.


zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
```diff
@@ -233,4 +233,6 @@ internal class Endpoint internal constructor(
     open fun callEnd(call: Call, result: CallResult, startValue: Any?) {
     }
   }
+
+  internal companion object
```
colinrtwhite: Some of the generated code attempts to access `Endpoint`'s companion object. Need to figure out why

zipline-testing/build.gradle.kts
```diff
@@ -69,19 +68,23 @@ kotlin {
...
+  withType<KotlinNativeCompile>().configureEach {
+    compilerOptions {
+      freeCompilerArgs.addAll("-Xpartial-linkage-loglevel=ERROR")
     }
   }
 
   // https://youtrack.jetbrains.com/issue/KT-56025
   // https://youtrack.jetbrains.com/issue/KT-57203
-  named("jsBrowserProductionWebpack").configure {
+  findByName("jsBrowserProductionWebpack")?.apply {
```
swankjesse, line 80: Comment // K1, // K2

swankjesse (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt
colinrtwhite: This'll be an issue when we start compiling Zipline itself with K2. Until then we don't need this.

zipline/src/jniMain/kotlin/app/cash/zipline/QuickJs.kt
```diff
@@ -174,3 +174,5 @@ actual class QuickJs private constructor(
   private external fun gc(context: Long)
   private external fun setMaxStackSize(context: Long, stackSize: Long)
 }
+
+internal expect fun loadNativeLibrary()
```
colinrtwhite, line 178: This expect declaration is only necessary if we're compiling the zipline runtime with K2 (which won't happen until Kotlin 2.0.0) - can revert if preferred.

build.gradle.kts
```diff
@@ -167,7 +167,7 @@ allprojects {
       repositories {
         maven {
           name = "testMaven"
-          url = file("${rootProject.buildDir}/testMaven").toURI()
+          url = file("${rootProject.layout.buildDirectory.asFile.get()}/testMaven").toURI()
```
JakeWharton:
```suggestion
          url = rootProject.layout.buildDirectory.dir("testMaven").get().asFile.toURI()
```

JakeWharton (approved):
Sweet

PR #1215, opened 2024-01-08
Adding zipline-cryptography package 

AlexSzlavik:
Adding zipline-security package to provide cryptographic APIs for guest code

Adding a new package to support cryptography primitives in zipline. This PR adds project layout and an implementation of a Random (SecureRandom specifically) interface. This commit only adds end to end testing via zipline.ZiplineSecurityTest. This PR does not change existing usage of cryptographic abstractions (ie. for code signing).

Paired with @swankjesse, @alvinsee


zipline-security/build.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+    val nativeMain by getting {
+      dependsOn(hostMain)
+    }
+
+    val jvmMain by getting {
+      dependsOn(hostMain)
+    }
+  }
+
+  // TODO(jessewilson): move this to a common build module, shared with zipline-testing.
```
AlexSzlavik: @swankjesse fyi that we left this as a follow up

swankjesse: 👍🏻

zipline/src/hostTest/kotlin/app/cash/zipline/ZiplineSecurityTest.kt
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
```
swankjesse:
```suggestion
 * Copyright (C) 2024 Block, Inc.
```

swankjesse (approved):

swankjesse:
Test failure is exactly the one we want:

```
app.cash.zipline.ZiplineSecurityTest.secureRandomWorksWithNativeImplementation[linuxX64] FAILED
    app.cash.zipline.ZiplineException at null:-1
There were failing tests. See the report at: file:///home/runner/work/zipline/zipline/zipline/build/reports/tests/linuxX64ReleaseTest/index.html
```

Best mitigation is to ignore that test. Something like this in the test itself:

```
  if (isLinux()) return
```

(Unfortunately there’s no annotation like `@Ignore` that’ll apply to Linux but not other platforms, and there’s also no `AssumptionViolatedException` for Kotlin/Native tests, yet)

swankjesse:
... and to test if it’s Linux or not, you actually need more expect/actual stuff like [IoTestingNative.kt](https://github.com/cashapp/zipline/blob/trunk/zipline-testing/src/nativeMain/kotlin/app/cash/zipline/testing/IoTestingNative.kt#L30
).

zipline/src/hostTest/kotlin/app/cash/zipline/ZiplineSecurityTest.kt
AlexSzlavik: Other files in this directory are attributed to Square. Follow up to clean these up and change to Block? (some are 2022 when block existed)

zipline-security/src/appleMain/kotlin/app/cash/zipline/security/RealZiplineSecurityService.kt
```diff
@@ -0,0 +1,50 @@
...
+    val status = SecRandomCopyBytes(
+      kSecRandomDefault,
+      result.size.convert(),
+      bytesPointer,
+    )
+    pin.unpin()
+
+    require(status == errSecSuccess) {
+      "failed to generate random bytes."
+    }
```
JakeWharton:
```suggestion
    if (size != 0) {
      val status = result.usePinned {
        SecRandomCopyBytes(
          kSecRandomDefault,
          result.size.convert(),
          it.addressOf(0),
        )
      }

      check(status == errSecSuccess) {
        "failed to generate random bytes."
      }
    }
```
Summary:
* Avoid dealing with a `null` pointer and invoking the copy function for size of zero
* Use `usePinned` to avoid having to remember to manually balance `unpin()` calls
* Switch to `check` which throws `IllegalStateException` as `require` throws `IllegalArgumentException` and a failure to copy is not related to the input argument

zipline-security/src/jsMain/kotlin/app/cash/zipline/security/SecureRandom.kt
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.security
+
+class SecureRandom internal constructor(
```
JakeWharton:
```suggestion
internal class SecureRandom(
```
If the class cannot be externally constructed there's no reason to put it into the public API. Seems like we should _only_ be exposing a single random type that is always secure.

zipline-security/src/linuxMain/kotlin/app/cash/zipline/security/RealZiplineSecurityService.kt
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.security
+
+class RealZiplineSecurityService : ZiplineSecurityService {
```
JakeWharton:
```suggestion
internal class RealZiplineSecurityService : ZiplineSecurityService {
```

zipline-security/src/jsMain/kotlin/app/cash/zipline/security/ZiplineSecurity.kt
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.security
+
+import app.cash.zipline.Zipline
+
+/** Guest-side API to Zipline's security services. */
+class ZiplineSecurity(zipline: Zipline) {
```
JakeWharton: Don't think this is the API we should offer. The `take` is hidden internally meaning I cannot have two unrelated parts of my codebase instantiate this type.

Can `nextBytes` and `nextLong` be extension functions on the service type which is taken directly by user code instead?

zipline-security/src/jsMain/kotlin/app/cash/zipline/security/Random.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.security
+
+interface Random {
+  fun nextBytes(sink: ByteArray)
```
JakeWharton:
```suggestion
  fun nextBytes(sink: ByteArray, offset: Int = 0, count: Int = sink.size - offset)
```

zipline-security/README.md
```diff
@@ -0,0 +1,3 @@
+# Zipline Security
```
JakeWharton: "Security" as a name is a bit nebulous to me. What are the plans for this module in the future such that it deserves to be its own module (as opposed to built-in to Zipline's core)? Maybe a future tink dependency or something?

Based on the sentence below, I would have expected "Cryptography" to be the name.

gadphly: +1 on changing to cryptography.

zipline-security/api/zipline-security.api
```diff
@@ -0,0 +1,4 @@
+public final class app/cash/zipline/security/ZiplineSecurityJvmKt {
+	public static final fun installSecurityService (Lapp/cash/zipline/Zipline;)V
+}
```
gadphly, line 3: curious if the V is supposed to be there at the end of the line?

zipline-security/README.md
AlexSzlavik: Yea, I agree with this. We started with cryptography but changed to security on a whim. Changing

zipline-security/src/jsMain/kotlin/app/cash/zipline/security/SecureRandom.kt
swankjesse: I’d like a Random interface to support deterministic testing, and a secure implementation for production use.

AlexSzlavik: I agree that a Random interface needs to have a way to be mocked / faked in a deterministic way for testing. Can we not make the js type internal only and still fake the backend implementation?

zipline-security/src/jsMain/kotlin/app/cash/zipline/security/ZiplineSecurity.kt
AlexSzlavik: I *think* that's fine and don't have strong opinions myself. What you suggest certainly sounds like the idiomatic way to expose this in zipline, i think (I'm still learning). Maybe this is an abstraction that belongs in consumer code (higher level)

zipline-security/src/jsMain/kotlin/app/cash/zipline/security/SecureRandom.kt
JakeWharton: Right so that still necessitates only a single type which is mostly what I'm getting at. There's just one `Random` and you can either use factory functions to create a cryptographically-secure instance, or a seeded fake.

zipline-security/src/jsMain/kotlin/app/cash/zipline/security/ZiplineSecurity.kt
swankjesse: Lets land this as-is and then @JakeWharton  and I can figure out a good pattern for end-users in follow-up?

zipline-cryptography/src/hostMain/kotlin/app/cash/zipline/cryptography/ZiplineSecurity.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package app.cash.zipline.cryptography
+
+import app.cash.zipline.Zipline
+
+expect fun Zipline.installCryptographyService()
+
+internal fun Zipline.installCryptographyServiceInternal(ziplineCryptographyService: ZiplineCryptographyService) {
+  bind("zipline/security", ziplineCryptographyService)
```
swankjesse, line 23: for follow-up: change to "zipline/cryptography"

zipline-cryptography/src/jsMain/kotlin/app/cash/zipline/cryptography/SecureRandom.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+package app.cash.zipline.cryptography
+
+class SecureRandom internal constructor(
+  private val securityService: ZiplineCryptographyService,
+) : Random {
+  override fun nextBytes(sink: ByteArray, offset: Int, count: Int) {
+    if (sink.size < count) {
+      throw IllegalArgumentException("count: $count > sink.size: ${sink.size}")
+    }
```
swankjesse:
```suggestion
    require(offset + count <= sink.size) {
      "offset: $offset + count: $count > sink.size: ${sink.size}"
    }
```

zipline-cryptography/src/jsMain/kotlin/app/cash/zipline/cryptography/ZiplineCryptography.kt
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.zipline.cryptography
+
+import app.cash.zipline.Zipline
+
+/** Guest-side API to Zipline's security services. */
+class ZiplineCryptography(zipline: Zipline) {
+  private val cryptographyService = zipline.take<ZiplineCryptographyService>("zipline/security")
```
swankjesse:
```suggestion
  private val cryptographyService = zipline.take<ZiplineCryptographyService>("zipline/cryptography")
```

zipline-cryptography/src/hostMain/kotlin/app/cash/zipline/cryptography/ZiplineSecurity.kt
swankjesse:
```suggestion
  bind("zipline/cryptography", ziplineCryptographyService)
```

swankjesse (approved):

swankjesse:
Merged to get things going. We wanna do some follow-ups:

 * Linux on Kotlin/Native
 * Fix the API to get access so it doesn’t need to be globally coordinated
 * Figure out how to make Randoms testable

PR #1217, opened 2024-01-10
Draft for discussion: on sending a signal when networking fails

jingwei99:


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -281,6 +282,9 @@ class ZiplineLoader internal constructor(
       // If emit() threw a CancellationException, consider that emit to be successful.
       // That's 'cause loadOnce() accepts an element and then immediately cancels the flow.
       throw e
+    } catch (e: IOException) {
+      send(LoadResult.Failure(e))
+      return null
```
jingwei99, line 287: I started working on the idea of "surfacing networking failure", and realized that for the `catch (e: Exception)` branch, does `send(LoadResult.Failure(e))` already, and the downstream can check the type of the exception and if it's an `IOException`, the downstream will take it as an networking failure. So the change needed might be adding a documentation about `IOException` for `LoadResult.Failure`. 

Or maybe I'm missing something / forgot something

jingwei99: talked over video, we don't need this, closing the draft

PR #1245, opened 2024-02-07
Better error messages when topological sort fails

swankjesse:
No topological ordering is possible for these items:
      ./Kotlin-DateTime-library-kotlinx-datetime-js-ir.js (@js-joda/core)

### DIFF
--- a/zipline/src/hostMain/kotlin/app/cash/zipline/internal/topologicalSort.kt
+++ b/zipline/src/hostMain/kotlin/app/cash/zipline/internal/topologicalSort.kt
@@ -62,18 +62,11 @@ internal fun <T> List<T>.topologicalSort(sourceToTarget: (T) -> Iterable<T>): Li
     buildString {
       append("No topological ordering is possible for these items:")
 
-      val unorderedItems = toMutableSet()
-        .apply {
-          removeAll(result.toSet())
-        }
-
+      val unorderedItems = this@topologicalSort.toSet() - result.toSet()
       for (unorderedItem in unorderedItems) {
         append("\n  ")
         append(unorderedItem)
-        val unsatisfiedDeps = sourceToTarget(unorderedItem).toMutableSet()
-          .apply {
-            removeAll(result.toSet())
-          }
+        val unsatisfiedDeps = sourceToTarget(unorderedItem).toSet() - result.toSet()
         unsatisfiedDeps.joinTo(this, separator = ", ", prefix = " (", postfix = ")")
       }
     }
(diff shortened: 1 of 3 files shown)


zipline/src/hostMain/kotlin/app/cash/zipline/internal/topologicalSort.kt
```diff
@@ -59,7 +59,24 @@ internal fun <T> List<T>.topologicalSort(sourceToTarget: (T) -> Iterable<T>): Li
...
 
   require(result.size == this.size) {
-    "No topological ordering is possible for $this"
+    buildString {
+      append("No topological ordering is possible for these items:")
+
+      val unorderedItems = toMutableSet()
+        .apply {
+          removeAll(result.toSet())
+        }
```
JakeWharton: Not that it matters, so feel free to ignore, but since we don't care about performance right here...
```suggestion
      val unorderedItems = toSet() - result.toSet()
```

zipline/src/hostMain/kotlin/app/cash/zipline/internal/topologicalSort.kt
```diff
@@ -59,7 +59,24 @@ internal fun <T> List<T>.topologicalSort(sourceToTarget: (T) -> Iterable<T>): Li
...
+
+      for (unorderedItem in unorderedItems) {
+        append("\n  ")
+        append(unorderedItem)
+        val unsatisfiedDeps = sourceToTarget(unorderedItem).toMutableSet()
+          .apply {
+            removeAll(result.toSet())
+          }
```
JakeWharton:
```suggestion
        val unsatisfiedDeps = sourceToTarget(unorderedItem).toSet() - result.toSet()
```

JakeWharton (approved):

zipline/src/hostMain/kotlin/app/cash/zipline/internal/topologicalSort.kt
swankjesse: That’s what I did first, and for whatever reason it inferred the wrong type...

swankjesse: Ahhh . ..  `buildString()` has a receiver that defines `toSet()`. Fix is a little ugly, but okay:
```
      val unorderedItems = this@topologicalSort.toSet() - result.toSet()
```

zipline/src/hostMain/kotlin/app/cash/zipline/internal/topologicalSort.kt
swankjesse: Done.

zipline/src/hostMain/kotlin/app/cash/zipline/internal/topologicalSort.kt
JakeWharton: I don't feel strongly. Leave it as is

swankjesse: I prefer your recommendation!

Issue #1278, opened 2024-04-02
Zipline.close() should be calling Endpoint.disconnect()


swankjesse:
Looks like we might have a leak otherwise?

jingibus:
Would this fix address the root cause of the leaks? `inboundChannel` does retain a bunch of services, but `inboundChannel` *itself* is also leaking. Small potatoes compared to the services, but y'know... always better to strike at the root...

swankjesse:
Fixed by https://github.com/cashapp/zipline/pull/1315

PR #1280, opened 2024-04-02
Disconnected inbound services on close

jingibus:
Fixes part of #1278. `Zipline`'s `inboundChannel` itself is still leaked, however....

### DIFF
--- a/zipline/src/hostMain/kotlin/app/cash/zipline/Zipline.kt
+++ b/zipline/src/hostMain/kotlin/app/cash/zipline/Zipline.kt
@@ -143,6 +143,12 @@ actual class Zipline private constructor(
     closed = true
 
     scope.cancel()
+
+    // The inboundChannel is leaking for reasons unknown. Clean up all bound services
+    // to limit the blast radius.
+    for (serviceName in endpoint.inboundServices.keys.toTypedArray()) {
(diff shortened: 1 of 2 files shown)


jingibus:
> `Zipline`'s `inboundChannel` itself is still leaked, however....

I can't repro this under unit test. Going to let this one rest until I get this fix in and ingest the updated redwood + zipline artifacts in Cash Android.

zipline/src/hostMain/kotlin/app/cash/zipline/Zipline.kt
```diff
@@ -143,6 +143,9 @@ actual class Zipline private constructor(
     closed = true
 
     scope.cancel()
+    for (serviceName in endpoint.inboundServices.keys.toList()) {
```
JakeWharton: Why the copy? Does disconnect mutate? Can we use toTypedArray then if so? It's cheaper to create and cheaper to iterate. Plus a comment, please!

JakeWharton (approved):

zipline/src/hostMain/kotlin/app/cash/zipline/Zipline.kt
jingibus: Yep, disconnect mutates inboundServices.

PR #1284, opened 2024-04-08
Path should come after FileSystem in the parameters

### LINKED ISSUE #1283: Function parameters should be (FileSystem, Path) and not the opposite
Parameters should be ordered from widest context to narrowest.

For Okio’s types, the FileSystem should always precede the Path.

Goooler:
Closes #1283.

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -83,7 +83,7 @@ class ZiplineLoader internal constructor(
 
   @Deprecated(
     message = "Use `withEmbedded` which has FileSystem came first instead.",
-    level = DeprecationLevel.HIDDEN
+    level = DeprecationLevel.HIDDEN,
   )
   fun withEmbedded(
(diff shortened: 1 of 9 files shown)


JakeWharton (commented):
This breaks binary compatibility. We need to retain the old order as an overload, probably with hidden depreciation level.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -464,29 +473,40 @@ class ZiplineLoader internal constructor(
...
+  @Deprecated(
+    message = "Use `download` which has FileSystem came first instead.",
+    level = DeprecationLevel.HIDDEN
+  )
   suspend fun download(
     applicationName: String,
     downloadDir: Path,
     downloadFileSystem: FileSystem,
     manifestUrl: String,
+  ) = download(applicationName, downloadFileSystem, downloadDir, manifestUrl,)
```
JakeWharton:
```suggestion
  ) = download(applicationName, downloadFileSystem, downloadDir, manifestUrl)
```

JakeWharton (approved):

PR #1288, opened 2024-04-12
Don't leak Zipline instances

### LINKED ISSUE #1287: Zipline’s leak detector . . . leaks
In `leakCanaryJni.kt` we keep a global list of services we’re worried about leaking.

 * allReferencesSet (global singleton)
   * ZiplineServiceReference
     * OutboundCallHandler
       * Endpoint 

This is useful to detect leaks where the lifetime of a bound service is narrower than the lifetime of a `Zipline` instance.

But it also creates a leak when the `Zipline` instance is closed.

We should evict items from `allReferencesSet` in `Zipline.close()`.

swankjesse:
Closes: https://github.com/cashapp/zipline/issues/1287

### DIFF
--- a/zipline/src/jniMain/kotlin/app/cash/zipline/internal/bridge/leakCanaryJni.kt
+++ b/zipline/src/jniMain/kotlin/app/cash/zipline/internal/bridge/leakCanaryJni.kt
@@ -21,18 +21,29 @@ import java.lang.ref.ReferenceQueue
 import java.util.Collections.synchronizedSet
 
 internal actual fun trackLeaks(
-  eventListener: Endpoint.EventListener,
+  endpoint: Endpoint,
   serviceName: String,
   callHandler: OutboundCallHandler,
   service: ZiplineService,
 ) {
-  allReferencesSet += ZiplineServiceReference(eventListener, serviceName, callHandler, service)
+  allReferencesSet += ZiplineServiceReference(endpoint, serviceName, callHandler, service)
 }
 
 internal actual fun detectLeaks() {
   while (true) {
     val reference = allReferencesQueue.poll() ?: break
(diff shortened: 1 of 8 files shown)


zipline/src/jniMain/kotlin/app/cash/zipline/internal/bridge/leakCanaryJni.kt
```diff
@@ -36,22 +36,33 @@ internal actual fun detectLeaks() {
...
 
+internal actual fun stopTrackingLeaks(
+  endpoint: Endpoint,
+) {
+  val i = allReferencesSet.iterator()
+  while (i.hasNext()) {
+    if (i.next().endpoint == endpoint) {
+      i.remove()
+    }
+  }
```
JakeWharton:
```suggestion
  allReferencesSet.removeAll { it.endpoint == endpoint }
```

JakeWharton (approved):

zipline/src/jniTest/kotlin/app/cash/zipline/ZiplineTest.kt
```diff
@@ -78,16 +77,6 @@ class ZiplineTest {
...
   }
 
-  @Test fun inboundServicesReleasedOnClose() = runTest(dispatcher) {
-    zipline.quickJs.evaluate("testing.app.cash.zipline.testing.prepareJsBridges()")
-    val service = JvmEchoService("echo, echo, echo")
-    zipline.bind<EchoService>("helloService", service)
-
-    zipline.close()
-
-    assertThat(zipline.serviceNames).doesNotContain("helloService")
```
swankjesse, line 88: This test now fails! We actually free the whole Zipline though so we get what we want.

jingibus (approved):

PR #1298, opened 2024-04-24
Track host OS changes in GitHub actions

swankjesse:


.github/workflows/build.yaml
```diff
@@ -105,7 +105,7 @@ jobs:
           script: ./gradlew connectedCheck
 
   ios:
-    runs-on: macos-latest
+    runs-on: macos-latest-large
```
JakeWharton: Why switch this? Once code signing is disabled it should work fine on ARM, no?

.github/workflows/release.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
 
   android-emulator:
     # We build on a Mac to get hardware acceleration for the Android emulator.
-    runs-on: macos-latest
+    runs-on: macos-latest-large
```
JakeWharton: This should be moved to ubuntu which is faster than both architectures of macs

.github/workflows/release.yaml
```diff
@@ -84,7 +84,7 @@ jobs:
           script: ./gradlew connectedCheck
 
   publish:
-    runs-on: macos-latest
+    runs-on: macos-latest-large
```
JakeWharton: This can probably be arm

samples/world-clock/ios/app/WorldClock.xcodeproj/project.pbxproj
```diff
@@ -284,6 +303,7 @@
 				CLANG_WARN_UNREACHABLE_CODE = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
 				COPY_PHASE_STRIP = NO;
+				CODE_SIGNING_ALLOWED = NO;
```
JakeWharton, line 306: Ooo I was specifying this on the command line but I like it better in this garbage file where no one has to see it

JakeWharton (approved):

.github/workflows/build.yaml
swankjesse: Just attempting to preserve the old behavior. I think ARM is an improvement, I’ll switch

.github/workflows/release.yaml
swankjesse: Fixed.

.github/workflows/release.yaml
swankjesse: Fixed.

PR #1332, opened 2024-06-04
Publish only to Maven Central on CI

JakeWharton:
This avoids publishing to the test-local repo which, while cheap, does a bunch of file I/O which is slow on Macs.


colinrtwhite (approved):

colinrtwhite:
Copying this! https://github.com/coil-kt/coil/pull/2302

PR #1346, opened 2024-06-11
Don't run ./build-mac.sh on Linux

swankjesse:
The script assumes it's only ever executed on a Mac. When it runs on Linux it builds a Linux library.


JakeWharton (approved):

swankjesse:
This reverts a bit of my earlier PR: https://github.com/cashapp/zipline/pull/1298/files

JakeWharton:
CI failure seems legitimate. Something else is broken.

PR #1351, opened 2024-06-12
Update how we collect output JS files

swankjesse:
This is updated with Kotlin 2.0, and our previous approach isn't compatible with the configuration cache.

### DIFF
--- a/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
+++ b/zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
@@ -60,11 +60,14 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
 
     kotlinExtension.targets.withType(KotlinJsIrTarget::class.java).all { kotlinTarget ->
       kotlinTarget.binaries.withType(JsIrBinary::class.java).all { kotlinBinary ->
-        registerCompileZiplineTask(
+        val ziplineCompileTask = registerCompileZiplineTask(
           project = target,
           jsProductionTask = kotlinBinary.asJsProductionTask(),
           extension = ziplineExtension,
         )
+        ziplineCompileTask.configure {
+          it.dependsOn(kotlinBinary.linkTask)
+        }
       }
     }
 
@@ -141,6 +144,7 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
       createdTask.description = "Serves Zipline files"
       createdTask.inputDir.set(ziplineCompileTask.flatMap { it.outputDir })
       createdTask.port.set(extension.httpServerPort)
+      createdTask.dependsOn(ziplineCompileTask)
     }
 
     return ziplineCompileTask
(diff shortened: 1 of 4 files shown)


zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineGradlePluginTest.kt
```diff
@@ -519,6 +519,7 @@ class ZiplineGradlePluginTest {
   ): GradleRunner {
     val gradleRoot = projectDir.resolve("gradle").also { it.mkdir() }
     File("../gradle/wrapper").copyRecursively(gradleRoot.resolve("wrapper"), true)
+    File(projectDir, "kotlin-js-store/yarn.lock").delete()
```
swankjesse, line 522: Sneaking this in. My test was failing locally because I had an old yarn.lock file that the new dependencies didn’t like.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -60,11 +60,14 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
...
     kotlinExtension.targets.withType(KotlinJsIrTarget::class.java).all { kotlinTarget ->
       kotlinTarget.binaries.withType(JsIrBinary::class.java).all { kotlinBinary ->
-        registerCompileZiplineTask(
+        val ziplineCompileTask = registerCompileZiplineTask(
           project = target,
           jsProductionTask = kotlinBinary.asJsProductionTask(),
           extension = ziplineExtension,
         )
+        ziplineCompileTask.configure {
+          it.dependsOn(kotlinBinary.linkTask)
```
swankjesse, line 69: The error I’m attempting to fix:

```
InvalidUserCodeException: Querying the mapped value of flatmap(provider(task 'compileDevelopmentExecutableKotlinJs', class org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrLink)) before task ':activity:treehouse:compileDevelopmentExecutableKotlinJs' has completed is not supported
```

zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineGradlePluginTest.kt
JakeWharton: I think the other tests have configuration in their build.gradle to always overwrite the file which prevents this problem.

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
JakeWharton: The use of `dependsOn` means either there's a Gradle bug with the use of `Property` or a Kotlin feature request with a lack of `Property` being exposed.

JakeWharton (approved):

zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
swankjesse: Agreed.

PR #1352, opened 2024-06-13
Special case cancellation exceptions

swankjesse:


JakeWharton (approved):

zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt
```diff
@@ -86,7 +86,7 @@ internal class InboundService<T : ZiplineService>(
         override var passByReferenceName: String? = null
 
         override fun cancel() {
-          deferred.cancel()
+          deferred.cancel(theOnlyCancellationException)
```
swankjesse, line 89: FYI I added these also!

zipline/src/hostMain/kotlin/app/cash/zipline/Zipline.kt
```diff
@@ -145,7 +146,7 @@ actual class Zipline private constructor(
 
     var thrown: Throwable? = null
 
-    scope.cancel()
+    scope.cancel(theOnlyCancellationException)
```
swankjesse, line 149: and here.

PR #1370, opened 2024-07-08
Build directly on Ubuntu 20.04

JakeWharton:
CentOS 7 was EOL July 1, 2024.


JakeWharton:
I don't know how to test if this is old enough or not. 5-year-old software should be old enough for anyone... right?

JakeWharton:
#1374

PR #1371, opened 2024-07-08
Make ZiplineCache resilient to disk write failures

### LINKED ISSUE #1369: Cache needs to be more resilient to a full disk
We’ve seen some crashes where the cache behaves badly when the disk isn’t writable because it’s full.

swankjesse:
This pulls the try/catch block up to the public API of ZiplineCache. Exceptions put the cache in a degraded mode where all reads return null and all writes do nothing. This yields bad caching, but that's our best option given an unwritable disk.

This also introduces a very aggressive test case to confirm that the cache can recover from any write failure.

Closes: https://github.com/cashapp/zipline/issues/1369

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
@@ -21,7 +21,6 @@ import app.cash.zipline.loader.internal.cache.FileState
 import app.cash.zipline.loader.internal.cache.Files
 import app.cash.zipline.loader.internal.cache.SqlDriverFactory
 import app.cash.zipline.loader.internal.cache.createDatabase
-import app.cash.zipline.loader.internal.cache.isSqlException
 import app.cash.zipline.loader.internal.fetcher.LoadedManifest
 import okio.ByteString
 import okio.ByteString.Companion.decodeHex
@@ -52,6 +51,7 @@ class ZiplineCache internal constructor(
   private val directory: Path,
   private val maxSizeInBytes: Long,
 ) : Closeable {
+  private var hasWriteFailures = false
 
   /*
    * Files are named by their SHA-256 hashes. We use a SQLite database for file metadata: which
@@ -80,27 +80,23 @@ class ZiplineCache internal constructor(
     driver.close()
   }
 
-  internal fun write(
+  private fun write(
     applicationName: String,
     sha256: ByteString,
     content: ByteString,
     nowEpochMs: Long,
     isManifest: Boolean = false,
     manifestFreshAtMs: Long? = null,
-  ): Files? {
-    try {
-      val metadata = openForWrite(
-        applicationName = applicationName,
-        sha256 = sha256,
-        nowEpochMs = nowEpochMs,
-        isManifest = isManifest,
-        manifestFreshAtMs = manifestFreshAtMs,
-      ) ?: return null
-      write(metadata, content, nowEpochMs)
-      return metadata
-    } catch (ignored: IOException) {
-      return null // Silently ignore write failures; the disk might be full?
-    }
+  ): Files {
+    val metadata = openForWrite(
+      applicationName = applicationName,
+      sha256 = sha256,
+      nowEpochMs = nowEpochMs,
+      isManifest = isManifest,
+      manifestFreshAtMs = manifestFreshAtMs,
+    )
+    write(metadata, content, nowEpochMs)
+    return metadata
   }
 
   private fun write(metadata: Files, content: ByteString, nowEpochMs: Long) {
@@ -121,28 +117,40 @@ class ZiplineCache internal constructor(
     applicationName: String,
     sha256: ByteString,
     nowEpochMs: Long,
-    download: suspend () -> ByteString?,
-  ): ByteString? {
-    val read = read(sha256, nowEpochMs)
-    if (read != null) return read
+    download: suspend () -> ByteString,
+  ): ByteString {
+    if (hasWriteFailures) return download()
 
-    val content = download() ?: return null
-    write(
-      applicationName = applicationName,
-      sha256 = sha256,
-      content = content,
-      isManifest = false,
-      nowEpochMs = nowEpochMs,
-    )
+    try {
+      val read = read(sha256, nowEpochMs)
+      if (read != null) return read
+    } catch (e: Exception) {
+      hasWriteFailures = true // Mark this cache as broken.
+      return download()
+    }
+
+    val content = download()
+    try {
+      write(
+        applicationName = applicationName,
+        sha256 = sha256,
+        content = content,
+        isManifest = false,
+        nowEpochMs = nowEpochMs,
+      )
+    } catch (e: Exception) {
+      hasWriteFailures = true // Mark this cache as broken.
+    }
+
     return content
   }
 
-  internal fun getOrPutManifest(
+  private fun getOrPutManifest(
     applicationName: String,
     content: ByteString,
     putFreshAtMs: Long,
     nowEpochMs: Long,
-  ): Files? {
+  ): Files {
     val sha256 = content.sha256()
     val metadata = getOrNull(sha256)
     return metadata ?: write(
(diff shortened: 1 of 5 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/fetcher/FsCachingFetcher.kt
```diff
@@ -34,9 +34,9 @@ internal class FsCachingFetcher(
     nowEpochMs: Long,
     baseUrl: String?,
     url: String,
-  ): ByteString? {
+  ): ByteString {
     return cache.getOrPut(applicationName, sha256, nowEpochMs) {
-      delegate.fetch(applicationName, eventListener, id, sha256, nowEpochMs, baseUrl, url)
+      delegate.fetch(applicationName, eventListener, id, sha256, nowEpochMs, baseUrl, url)!!
```
swankjesse, line 39: The concrete implementation of `Delegate` is an HttpFetcher that never returns null. (Though it does throw exceptions).

swankjesse:
CI failure is unrelated but does need a fix!

```

+ yum -y install centos-release-scl
Loaded plugins: fastestmirror, ovl
Determining fastest mirrors
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"


 One of the configured repositories failed (Unknown),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).

     3. Run the command with the repository temporarily disabled
            yum --disablerepo=<repoid> ...

     4. Disable the repository permanently, so yum won't use it by default. Yum
        will then just ignore the repository until you permanently enable it
        again or use --enablerepo for temporary usage:

            yum-config-manager --disable <repoid>
        or
            subscription-manager repos --disable=<repoid>

     5. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:

            yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true

Cannot find a valid baseurl for repo: base/7/x86_64
```

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
```diff
@@ -80,27 +80,23 @@ class ZiplineCache internal constructor(
...
-  internal fun write(
+  private fun write(
     applicationName: String,
     sha256: ByteString,
     content: ByteString,
     nowEpochMs: Long,
     isManifest: Boolean = false,
     manifestFreshAtMs: Long? = null,
-  ): Files? {
-    try {
```
swankjesse, line 91: lots of the diffs in this file are moving try/catch blocks outside of implementation-detail functions and towards the public API. They’re more coarse-grained that way and that’s okay.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
```diff
@@ -121,28 +117,40 @@ class ZiplineCache internal constructor(
...
+    val content = download()
+    try {
+      write(
+        applicationName = applicationName,
+        sha256 = sha256,
+        content = content,
+        isManifest = false,
+        nowEpochMs = nowEpochMs,
+      )
+    } catch (e: Exception) {
```
swankjesse, line 141: Discarded exception is recovered here:
https://github.com/cashapp/zipline/pull/1375

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/fetcher/FsCachingFetcher.kt
JakeWharton: Can we move that invariant up to the property type? Then presumably its covariant override will eliminate the need for this double bang.

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
```diff
@@ -0,0 +1,187 @@
...
+ * expects from the cache: it accepts a wide variety of download counts. But regardless of whether
+ * the file system accepts writes, the cache should never return invalid results.
+ */
+class ZiplineCacheFaultsTest {
+  @Test
+  fun openMissHitClose(): Unit = runBlocking {
+    // How many writes are attempted in this test in the happy path. Determined experimentally!
+    val noFailuresWriteCount = 20
+
+    for (i in 0 until noFailuresWriteCount + 1) {
```
JakeWharton, line 40: until is going to do `- 1`
```suggestion
    for (i in 0..noFailuresWriteCount) {
```

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
```diff
@@ -0,0 +1,187 @@
...
+      assertThat(tester.fileSystemWriteCount).isLessThanOrEqualTo(noFailuresWriteCount)
+    }
+  }
+
+  @Test
+  fun openMissClose_OpenHitClose(): Unit = runBlocking {
+    // How many writes are attempted in this test in the happy path. Determined experimentally!
+    val noFailuresWriteCount = 20
+
+    for (i in 0 until noFailuresWriteCount + 1) {
```
JakeWharton, line 69:
```suggestion
    for (i in 0..noFailuresWriteCount) {
```

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
```diff
@@ -0,0 +1,187 @@
...
+      assertThat(tester.fileSystemWriteCount).isLessThanOrEqualTo(noFailuresWriteCount)
+    }
+  }
+
+  @Test
+  fun openFailClose_OpenMissClose_OpenHitClose(): Unit = runBlocking {
+    // How many writes are attempted in this test in the happy path. Determined experimentally!
+    val noFailuresWriteCount = 27
+
+    for (i in 0 until noFailuresWriteCount + 1) {
```
JakeWharton, line 100:
```suggestion
    for (i in 0..noFailuresWriteCount) {
```

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  /** This test confirms that stale cached results are pruned. */
+  @Test
+  fun openMissClose_openStaleClose(): Unit = runBlocking {
+    // How many writes are attempted in this test in the happy path. Determined experimentally!
+    val noFailuresWriteCount = 31
+
+    for (i in 0 until noFailuresWriteCount + 1) {
```
JakeWharton, line 139:
```suggestion
    for (i in 0..noFailuresWriteCount) {
```

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
```diff
@@ -0,0 +1,187 @@
...
+        assertThat(tester.downloadCount).isEqualTo(2)
+
+        // The subsequent launch should have between 2 and 4 cache misses, depending on what was
+        // successfully written to the cache last time.
+        loadApp()
+        assertThat(tester.downloadCount).isLessThanOrEqualTo(4)
+      }
+
+      // Assert more strictly if we didn't need to inject any write failures.
+      if (i >= noFailuresWriteCount) {
```
JakeWharton, line 56: Not that it matters, but this should never be able to be greater than, right? It could only ever be equal to.

JakeWharton (approved):

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
JakeWharton: Got 'em in the follow-up.

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
swankjesse: ooooh yeah lemme fix that.

swankjesse: https://github.com/cashapp/zipline/pull/1375

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/internal/fetcher/FsCachingFetcher.kt
swankjesse: https://github.com/cashapp/zipline/pull/1375

PR #1374, opened 2024-07-08
Change Linux build from amd64/centos:7 to amd64/oraclelinux:8

swankjesse:
Our downstream consumers no longer need builds for CentOS 7 (2014) and are now happy with Oracle Linux 8 (2019).

### DIFF
--- a/.github/workflows/build-linux-amd64.sh
+++ b/.github/workflows/build-linux-amd64.sh
@@ -1,22 +1,11 @@
 #!/bin/bash
 
-# Run this in Docker with the Zipline project mounted at /zipline.
-#
-# We use Docker because GitHub actions doesn't offer virtual environments for old distros, and we
-# need to build on old distros to run on old distros.
-
 set -e
 set -x
-
-cd /zipline
-
-dnf install gcc-c++ cmake3 java-11-openjdk-devel
 
 mkdir -p build/jni/amd64/
 mkdir -p zipline/src/jvmMain/resources/jni/amd64/
 cmake3 -S zipline/src/jvmMain/ -B build/jni/amd64/ \
   -DQUICKJS_VERSION="$(cat zipline/native/quickjs/VERSION)" \
-  -DCMAKE_SYSTEM_PROCESSOR=x86_64 \
-  -DCMAKE_C_COMPILER=/usr/bin/gcc \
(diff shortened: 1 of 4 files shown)


.github/workflows/build-linux-amd64.sh
```diff
@@ -2,31 +2,21 @@
...
 
 mkdir -p build/jni/amd64/
 mkdir -p zipline/src/jvmMain/resources/jni/amd64/
 cmake3 -S zipline/src/jvmMain/ -B build/jni/amd64/ \
   -DQUICKJS_VERSION="$(cat zipline/native/quickjs/VERSION)" \
   -DCMAKE_SYSTEM_PROCESSOR=x86_64 \
-  -DCMAKE_C_COMPILER=/opt/rh/devtoolset-7/root/usr/bin/gcc \
-  -DCMAKE_CXX_COMPILER=/opt/rh/devtoolset-7/root/usr/bin/c++
+  -DCMAKE_C_COMPILER=/usr/bin/gcc \
+  -DCMAKE_CXX_COMPILER=/usr/bin/g++
```
JakeWharton: Just delete? These should be discovered

.github/workflows/build-linux-amd64.sh
```diff
@@ -2,31 +2,21 @@
...
 
 cd /zipline
-yum -y install centos-release-scl
-yum -y install devtoolset-7 devtoolset-7-toolchain
-. /opt/rh/devtoolset-7/enable
-yum -y install epel-release
-yum -y install cmake3
-yum -y install java-1.8.0-openjdk-devel
+
+dnf install gcc-c++ cmake3 java-11-openjdk-devel
```
JakeWharton: We should move this to the action YAML so that this script works on bare metal linux the same way the mac one works on macs. The Docker comment should also move there, I think.

JakeWharton (approved):

.github/workflows/build-linux-amd64.sh
swankjesse: Done.

.github/workflows/build-linux-amd64.sh
swankjesse: Done. I made a basic `Dockerfile`.

PR #1375, opened 2024-07-08
Publish events in ZiplineLoader

swankjesse:


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/LoaderEventListener.kt
```diff
@@ -0,0 +1,42 @@
...
+   *     write error or a metadata database write error.
+   */
+  open fun cacheStorageFailed(
+    applicationName: String?,
+    e: Exception,
+  ) {
+  }
+
+  companion object {
+    val NONE: LoaderEventListener = object : LoaderEventListener() {
```
JakeWharton:
```suggestion
    val None: LoaderEventListener = object : LoaderEventListener() {
```

JakeWharton: This is the nit-iest of nits, but I've been preferring to use a named class here instead.
```kotlin
val None: LoaderEventListener = EmptyLoaderEventListener()

...

private class EmptyLoaderEventListener : LoaderEventListener {
  ...
}
```
because its `toString()` includes the fact that it's empty which is useful when `println` debugging or in the actual debugger rather than seeing `LoaderEventListener$1@1ab4bfd`.

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
```diff
@@ -35,9 +35,9 @@ class ZiplineCacheFaultsTest {
   @Test
   fun openMissHitClose(): Unit = runBlocking {
     // How many writes are attempted in this test in the happy path. Determined experimentally!
-    val noFailuresWriteCount = 20
+    val noFailuresWriteCount = 21
 
-    for (i in 0 until noFailuresWriteCount + 1) {
+    for (i in 0..noFailuresWriteCount) {
```
JakeWharton, line 40: Ah, got it here!

JakeWharton (approved):

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/LoaderEventListener.kt
```diff
@@ -0,0 +1,39 @@
...
+   * @param e the exception that triggered this failure. This is typically either a file system
+   *     write error or a metadata database write error.
+   */
+  open fun cacheStorageFailed(
+    applicationName: String?,
+    e: Exception,
+  ) {
+  }
+
+  object None : LoaderEventListener()
```
swankjesse, line 38: The toString is `app.cash.zipline.loader.LoaderEventListener$None@15eebbff`

PR #1377, opened 2024-07-09
Build all projects only on ARM Mac

JakeWharton:
We only need to test the QuickJS build is working for the architecture. We can assume a single test of everything else on any architecture is sufficent.


JakeWharton:
This is specifically to avoid the incessant flakes on the macos-latest-large runner (which is in fact smaller than macos-latest).

swankjesse (approved):

Issue #1385, opened 2024-07-12
Thread per app


swankjesse:
ZiplineLoader currently assumes a single dispatcher shared by all the apps it loads. We should change this to be a thread-per-app, for greater parallelism.

swankjesse:
Implemented!

PR #1386, opened 2024-07-15
Use different dispatchers for cache vs code

swankjesse:
The cache needs to be thread-isolated.

QuickJS needs to be thread-isolated.

But they don't need to be the same threads. This makes that possible.

This is working towards https://github.com/cashapp/zipline/issues/1385

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -54,6 +54,19 @@ import okio.Path
  *
  * Loader attempts to load code as quickly as possible with concurrent network downloads and code
  * loading.
+ *
+ * # Concurrency
+ *
+ * This class uses two different coroutine dispatchers:
+ *
+ *  - The [dispatcher] is the only dispatcher permitted to call functions on the returned [Zipline]
+ *    instance. Each instance is thread-confined and this dispatcher must implement that
+ *    enforcement.
+ *  - The [cacheDispatcher] is the only dispatcher permitted to call functions on [ZiplineCache].
+ *    The cache is also thread-confined, and the two threads may be different.
+ *
+ * In applications where multiple [Zipline] instances are used, but the instances all share a cache,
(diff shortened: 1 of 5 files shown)


swankjesse:
I started writing a test for this and decided that the easiest way to implement this is via new EventListener events for cache hits. I’m going to do that in a separate PR as it requires its own tests!

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/LoaderTester.kt
```diff
@@ -68,9 +70,9 @@ class LoaderTester(
     systemFileSystem.createDirectories(tempDir, mustCreate = true)
     systemFileSystem.createDirectories(embeddedDir, mustCreate = true)
     cache = testZiplineCache(
-      systemFileSystem,
-      cacheDir,
-      cacheMaxSizeInBytes.toLong(),
+      fileSystem = systemFileSystem,
```
adrw, line 73: nice named params

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -54,6 +54,19 @@ import okio.Path
  *
  * Loader attempts to load code as quickly as possible with concurrent network downloads and code
  * loading.
+ *
+ * # Concurrency
```
adrw, line 58: thanks for docs

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -101,9 +116,20 @@ class ZiplineLoader internal constructor(
 
   fun withCache(
     cache: ZiplineCache,
+    cacheDispatcher: CoroutineDispatcher? = null,
   ): ZiplineLoader = copy(
     cache = cache,
+    cacheDispatcher = cacheDispatcher,
+  )
+
+  @Deprecated(
```
adrw, line 125: nice backwards compatability

adrw (approved):
LGTM

PR #1431, opened 2024-09-06
Add compile subcommand to zipline-cli

JakeWharton:
Update Gradle plugin to invoke CLI to perform compilation. This allows compilation to not be subject to the version of Kotlin on the Gradle buildscript classpath.

Should unblock #1420


zipline-cli/src/main/kotlin/app/cash/zipline/cli/Compile.kt
```diff
@@ -0,0 +1,104 @@
...
+    )
+    .convert {
+      val (idName, name, keyHex) = it.split(':', limit = 3)
+      val id = SignatureAlgorithmId.valueOf(idName)
+      val key = keyHex.decodeHex()
+      Triple(id, name, key)
+    }
+    .multiple()
+
+  private val metadata by option().associate()
```
swankjesse, line 69: This API is so powerful

zipline-cli/src/main/kotlin/app/cash/zipline/cli/Compile.kt
```diff
@@ -0,0 +1,104 @@
...
+      outputDir = outputDir,
+      mainFunction = mainFunction,
+      mainModuleId = mainModuleId,
+      manifestSigner = manifestSigner,
+      version = version,
+      metadata = metadata,
+      stripLineNumbers = stripLineNumbers,
+    )
+
+    if (addedFiles.size or removedFiles.size or modifiedFiles.size != 0) {
```
swankjesse, line 94: is this bitwise or ? Sneaky! I’d complain but it passes the less-code test

JakeWharton: Yep. The normal isNotEmpty and boolean conditionals wrapped and it annoyed me so... I figure it's only in one place.

PR #1475, opened 2024-12-09
Version code defaults to null, fix comment on manifest version

thekirankumar:
The default value of "version" is null. Git commit SHA is a recommended way to populate the version, but is not the default


JakeWharton:
This is not referring to the default expression as specified directly on the argument, but instead the default which is injected by our Gradle plugin. If that is not clear (which I would agree that it is _not_ clear), we should adding clarification words, not removing this phrase.

PR #1554, opened 2025-04-11
Use Jetty instead of http4k

squarejesse:
The http4k Commercial License warning turned me off.


zipline-gradle-plugin/src/test/kotlin/app/cash/zipline/gradle/ZiplineDevelopmentServerTest.kt
```diff
@@ -114,13 +120,65 @@ class ZiplineDevelopmentServerTest {
...
+    val downloadedManifest2 = httpClient.call(manifestUrl).use { response ->
+      assertThat(response.networkResponse!!.code).isEqualTo(304)
+      ZiplineManifest.decodeJson(response.body!!.string())
+    }
+
+    assertThat(downloadedManifest2).isEqualTo(downloadedManifest1)
+  }
+
+  @Test
+  fun webSocketReload() {
```
squarejesse, line 152: I saw this hang locally once but I’m not sure why. I MIGHT have put in enough concurrency stuff to prevent that, but I might have also missed something. If this is flaky in the future, we can `@Ignore` it then

JakeWharton (approved):

PR #1611, opened 2025-07-09
Generate es2015 JavaScript

squarejesse:
We get better stack traces this way.


zipline-gradle-plugin/src/main/kotlin/app/cash/zipline/gradle/ZiplinePlugin.kt
```diff
@@ -69,6 +70,10 @@ class ZiplinePlugin : KotlinCompilerPluginSupportPlugin {
     )
 
     kotlinExtension.targets.withType(KotlinJsIrTarget::class.java).all { kotlinTarget ->
+      kotlinTarget.compilerOptions {
+        this.target.set("es2015")
+        this.moduleKind.set(MODULE_UMD)
```
```suggestion
        // Target latest JS to get classes, arrow functions, etc.
        this.target.set("es2015")
        // But our loader requires we still use the old module format.
        this.moduleKind.set(MODULE_UMD)
```

JakeWharton (approved):

PR #1616, opened 2025-07-15
Flip the endianness of the function flags

swankjesse:
This value was mostly pass-through: we read it
and wrote it back, without interpretting it, so
it was easy for this bug to sneak through.

The one catch we didn't handle is the debug
flag is always true in practice, but with
ES2015 we now see the incorrect flag triggering
incorrect behavior.

This was annoying to isolate.

### DIFF
--- a/zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectReader.kt
+++ b/zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectReader.kt
@@ -16,14 +16,24 @@
 package app.cash.zipline.bytecode
 
 import okio.Buffer
-import okio.BufferedSource
 import okio.Closeable
 import okio.IOException
 
 class JsObjectReader(
-  private val source: BufferedSource,
+  private val source: Buffer,
 ) : Closeable by source {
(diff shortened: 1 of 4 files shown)


zipline-bytecode/src/main/kotlin/app/cash/zipline/bytecode/JsObjectReader.kt
```diff
@@ -73,12 +84,12 @@ class JsObjectReader(
...
+      else -> throw IOException("unsupported tag: $tag at 0x${objectOffset.toString(16)}")
     }
   }
 
   private fun readFunction(): JsFunctionBytecode {
-    val flags = source.readShort().toInt()
+    val flags = source.readShortLe().toInt()
     val jsMode = source.readByte()
     val functionName = readAtomString()
     val argCount = source.readLeb128()
```
swankjesse, line 95: 20 lines down is this:

```
val hasDebug = flags.bit(11)
```

QuickJS always emits debug info in our current configuration. And the incorrect flag we were reading was also always true, so we got away with this bug.

Until we started doing ES2015 stuff, which caused the incorrect flag to be false sometimes!

JakeWharton (approved):

PR #1626, opened 2025-07-22
Stop using deprecated Kotlin compiler APIs

swankjesse:


zipline-kotlin-plugin/src/main/kotlin/app/cash/zipline/kotlin/AdapterGenerator.kt
```diff
@@ -430,7 +425,7 @@ internal class AdapterGenerator(
 
       val serializers = bridgedInterface.declareSerializerTemporaries(
         statementsBuilder = this@irFunctionBody,
-        serializersModuleParameter = ziplineFunctionsFunction.valueParameters[0],
+        serializersModuleParameter = ziplineFunctionsFunction.parameters[1],
```
swankjesse, line 428: (many of these shifted because the receiver is parameter 0)

zipline-kotlin-plugin/build.gradle.kts
```diff
@@ -17,6 +17,14 @@ dependencies {
   compileOnly(libs.auto.service.annotations)
 }
 
+kotlin {
+  sourceSets {
+    all {
```
JakeWharton: I believe `all` eagerly configures (not that it really matters much here)
```suggestion
    configureEach {
```

JakeWharton (approved):

PR #1657, opened 2025-09-08
Don't crash if creating a cache directory fails

swankjesse:
I'd previously attempted to recover when writes fail, but I missed this case.

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
@@ -27,7 +27,6 @@ import okio.ByteString.Companion.decodeHex
 import okio.Closeable
 import okio.FileNotFoundException
 import okio.FileSystem
-import okio.IOException
 import okio.Path
 
 /**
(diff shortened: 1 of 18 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
```diff
@@ -201,19 +200,14 @@ class ZiplineCache internal constructor(
...
     }
 
     return result
   }
 
-  internal fun pin(applicationName: String, sha256: ByteString) {
-    val fileId = database.filesQueries.get(sha256.hex()).executeAsOneOrNull()?.id ?: return
-    createPinIfNotExists(applicationName, fileId)
-  }
-
```
swankjesse, line 216: Was unused

zipline-loader/src/commonTest/kotlin/app/cash/zipline/loader/internal/cache/ZiplineCacheFaultsTest.kt
```diff
@@ -35,7 +35,7 @@ class ZiplineCacheFaultsTest {
   @Test
   fun openMissHitClose(): Unit = runBlocking {
     // How many writes are attempted in this test in the happy path. Determined experimentally!
-    val noFailuresWriteCount = 21
+    val noFailuresWriteCount = 23
```
swankjesse, line 38: Experimentally!

JakeWharton (approved):

PR #1664, opened 2025-09-12
Fix build failures: Spotless formatting and dependency compatibility

arora-vishal:
- Applied Spotless Kotlin formatting to resolve 111 formatting violations
- Upgraded JVM toolchain from Java 11 to Java 17 for Jetty 12.1.1 compatibility Jetty 12.1.1 requires Java 17+ and was causing plugin validation failures
- Downgraded dev.zacsweers.kctfork:core from 0.9.0 to 0.8.0 Version 0.9.0 was bringing in kotlin-compiler-embeddable 2.2.20 which is incompatible with the project's Kotlin 2.2.10

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -150,7 +150,7 @@ allprojects {
 
   plugins.withId("org.jetbrains.kotlin.multiplatform") {
     configure<KotlinMultiplatformExtension> {
-      jvmToolchain(11)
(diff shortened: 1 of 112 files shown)


JakeWharton:
Thanks, but this doesn't address the underlying issue of missing CI coverage. I'm working on that in #1682.

PR #1666, opened 2025-09-18
Add Windows Build Script

crowforkotlin:
# Changelog

## Updates

1. **Update `build.zig` for zig 0.16.0 Compatibility**
   - The build script is now compatible with zig-0.16.0.
   - Deprecated and removed APIs in the new zig version have been properly handled.

2. **Add Windows x86_64 QuickJS DLL Build Support**
   - `build.zig` now supports building `quickjs.dll` for Windows x86_64.
   - Folder naming conventions for Linux and macOS have been revised to prevent output conflicts, ensuring that `zig build` no longer causes macOS artifacts to overwrite Linux ones.

3. **Enhanced Build Output**
   - In addition to outputting to `zig-out`, library artifacts are now also copied to `src/jvmMain/resources` to support direct execution of test files on Windows.

4. **Update `QuickJsNativeLoader.kt`**
   - Added Windows support for the native loader, improving cross-platform compatibility.

5. **Testing**
   - Successfully passed tests in `zipline/src/hostTest/kotlin/app/cash/zipline/QuickJsTest.kt`, confirming the effectiveness of the above changes.

---
- I will continue to test the compatibility of zipline windows and quickjs and multiple tests on this branch. For now, I will submit it to the windows-dev branch and keep the latest code synchronized from the trunk branch.
---

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
 
       - uses: mlugg/setup-zig@v2
         with:
-          version: 0.14.0
+          version: 0.16.0-dev.238+580b6d1fa
 
       - run: zig build -p src/jvmMain/resources/jni
         working-directory: zipline
--- a/zipline/build.zig
+++ b/zipline/build.zig
@@ -5,28 +5,32 @@ pub fn build(b: *std.Build) !void {
   const deleteLib = b.addRemoveDirTree(.{ .cwd_relative = b.getInstallPath(.prefix, "lib") });
   b.getInstallStep().dependOn(&deleteLib.step);
 
-  try setupTarget(b, &deleteLib.step, .linux, .aarch64, "aarch64");
-  try setupTarget(b, &deleteLib.step, .linux, .x86_64, "amd64");
-  try setupTarget(b, &deleteLib.step, .macos, .aarch64, "aarch64");
-  try setupTarget(b, &deleteLib.step, .macos, .x86_64, "x86_64");
+  try setupTarget(b, &deleteLib.step, .linux, .aarch64, "linux_aarch64");
+  try setupTarget(b, &deleteLib.step, .linux, .x86_64, "linux_amd64");
+  try setupTarget(b, &deleteLib.step, .macos, .aarch64, "macos_aarch64");
+  try setupTarget(b, &deleteLib.step, .macos, .x86_64, "macos_x86_64");
+  try setupTarget(b, &deleteLib.step, .windows, .x86_64, "windows_amd64");
 }
 
 fn setupTarget(b: *std.Build, step: *std.Build.Step, tag: std.Target.Os.Tag, arch: std.Target.Cpu.Arch, dir: []const u8) !void {
-  const lib = b.addSharedLibrary(.{
+  const lib = b.addLibrary(.{
     .name = "quickjs",
-    .target = b.resolveTargetQuery(.{
-      .cpu_arch = arch,
-      .os_tag = tag,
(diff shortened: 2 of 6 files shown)


.github/workflows/build.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
 
       - uses: mlugg/setup-zig@v2
         with:
-          version: 0.14.0
+          version: 0.16.0-dev.238+580b6d1fa
```
JakeWharton, line 23: Is there a reason you are bumping to a development version to support Windows? If something is required, then we have to wait for 0.16.0 stable.

However, if you want to upgrade to 0.15.1 first (#1646), please do that in its own PR separate from Windows support.

zipline/build.zig
```diff
@@ -71,6 +75,13 @@ fn setupTarget(b: *std.Build, step: *std.Build.Step, tag: std.Target.Os.Tag, arc
   });
 
   const install = b.addInstallArtifact(lib, .{
+    .dest_dir = .{
+      .override = .{
+        .custom = b.fmt("../src/jvmMain/resources/jni/{s}", .{dir}),
+      },
+    },
+  });
```
JakeWharton, line 83: We're already specifying `-p` when running the Zig build to direct the resources to the right place. I think I would prefer we either stick with that or just change the single default location to be the right thing.

Is there some other reason to install to both places?

zipline/build.zig
```diff
@@ -5,28 +5,32 @@ pub fn build(b: *std.Build) !void {
...
+      }),
+      .optimize = .ReleaseSmall,
     }),
-    .optimize = .ReleaseSmall,
   });
 
   var version_buf: [11]u8 = undefined;
   const version = try readVersionFile(&version_buf);
-  var quoted_version_buf: [12]u8 = undefined;
+  var quoted_version_buf: [14]u8 = undefined;
```
JakeWharton, line 33: Why this change? There's only 12 characters total.

.github/workflows/build.yaml
crowforkotlin: Okay, I understand. Upgrading the version could have a widespread impact on the build artifacts. Therefore, we should first add Windows support to the existing version. After the PR is approved, we can then proceed to the next step of upgrading the version, rather than upgrading the version and adding Windows support all at once. Moreover, this should follow the Single Responsibility Principle for PRs.

zipline/build.zig
crowforkotlin: Thanks for the feedback
I see now that my change introduces redundancy and mixes a local workflow concern into the main build logic. The reason I added it was because I was struggling to run QuickJsTest.kt directly from the IDE. The test requires the compiled JNI library to be in the resources directory, and I was having to manually copy it over after every zig build. This change was a shortcut to automate that manual step for my local debugging.
I agree that this is not the right place for this fix. I will remove the .dest_dir override.

zipline/build.zig
crowforkotlin: When testing this, I tried using a 12-byte size and there was a problem during the build. I will adjust this, it was my negligence.

JakeWharton: If the size is too small for Windows I'd rather just bump to an arbitrary number like 50.

crowforkotlin: got it👌

crowforkotlin:
OK, the changes are very minor. I tested using zipline in Compose Desktop, and it loaded very smoothly. Next, I will take a look at the project source code written by the author to learn from it. I will continue to follow up! ! Thank you!!!
@JakeWharton 
image

PR #1679, opened 2025-09-30
Build tests once, run on multiple platforms

JakeWharton:
via test-distribution-gradle-plugin. Also add Linux ARM runner.


JakeWharton:
Our tests require the packaged JS, and while I could package those up too it's getting a bit ridiculous.

PR #1678, opened 2025-09-27
Add Windows Jvm build support

crowforkotlin:
## PR
- **This pull request primarily adds support for compiling and packaging on the Windows JVM platform. Compilation has also been verified on a Windows WSL Ubuntu system. It has been released to Maven and can be included in Compose Desktop. Zipline support on Windows platforms works as expected.**

## ChangeList
- **Added: JvmNativeLoader.kt adds Windows support.**
- **Modified: Changed QuickJS Library naming to resolve overwriting of folders on different platforms.**
- **Modified: Fixed an error in Zig reading the QuickJS version number on Windows.**

# Fix
image

# Release
Maven

Issue #1360, opened 2024-06-22
Unsupported OS: windows


TelephoneTan:
I am trying to use zipline in win11 desktop target of KMP, but it failed with:
```
Exception in thread "main" java.lang.ExceptionInInitializerError
	at app.cash.zipline.Zipline$Companion.create(Zipline.kt:198)
	at app.cash.zipline.loader.ZiplineLoader.loadFromManifest$zipline_loader(ZiplineLoader.kt:437)
	at app.cash.zipline.loader.ZiplineLoader.loadFromNetwork(ZiplineLoader.kt:305)
	at app.cash.zipline.loader.ZiplineLoader.access$loadFromNetwork(ZiplineLoader.kt:57)
	at app.cash.zipline.loader.ZiplineLoader$loadFromNetwork$1.invokeSuspend(ZiplineLoader.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:277)
	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:95)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:69)
	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:48)
	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source)
	at MainKt.main(main.kt:5)
	at MainKt.main(main.kt)
Caused by: java.lang.IllegalStateException: Unsupported OS: windows 11
	at app.cash.zipline.QuickJsNativeLoaderKt.loadNativeLibrary(QuickJsNativeLoader.kt:33)
	at app.cash.zipline.QuickJs.<clinit>(QuickJs.kt:36)
	... 15 more
```

JakeWharton:
There is no Windows support at this time and we have no plans of adding it ourselves. See #1342 for more

If someone wants to fix it and send a PR I'd be happy to merge it.

JakeWharton:
Now that we use Zig to build the native library, this should be easier to support.

PR #1680, opened 2025-09-30
Add Windows JVM support

JakeWharton:
Half of #1360. Closes #1678.


crowforkotlin:
@JakeWharton 
1. The error reported above, build/platform-tests (windows-2025) (pull_request), requires merging PR #1677 first to resolve the Windows CL error. Because I'm concerned that the separate PRs for CL changes and support additions might not be approved, I created two PRs for the two Windows changes (one for Windows CL fixes and the other for Windows JVM Library builds and unit testing support).

2. Secondly, I'm seeing an error with the CL jvmTest macOsX64Test on Mac. I checked the actions in the current trunk branch of Zipline and found that some actions are skipped when executing this task, resulting in the CL passing, while others are not skipped and fail. This is odd. I've reproduced the issue locally and found that a large amount of test code in the test case implementation points to this issue. I've analyzed the build artifacts and debugged them, but I've forgotten the specific details. I'll investigate further tomorrow morning.

The specific error reported by jvmTest macOsX64Test on Mac is in this code implementation.
https://github.com/cashapp/zipline/blob/7cf5c0fade23ed65dfff3327eea90da546df2179/zipline/src/jvmTest/kotlin/app/cash/zipline/ADummyTest.kt#L36

JakeWharton:
Fixing the error does not require your PR. We should not be running `apiCheck` on each shard of platform-tests because it forces native compilation of all targets.

crowforkotlin:
@JakeWharton Thanks for the explanation. I agree that running apiCheck on each platform-tests shard is inefficient and triggers unnecessary native compilations, which is something we should definitely optimize.
However, I believe the direct cause of the windows-2025 check failure is a more fundamental build error. Looking at the logs, the task fails at :zipline:linuxX64Quickjs with a clang error:
code

* What went wrong:
Execution failed for task ':zipline:linuxX64Quickjs'.
> Process 'command 'C:\Users\runneradmin/.cklib/llvm-11.1.0-windows-x64-2/bin/clang'' finished with non-zero exit value 1
This error indicates that the native compilation task itself is broken on the Windows environment, and it will fail whenever it's triggered, regardless of apiCheck. My other PR, #1677, is specifically intended to fix these kinds of Windows command-line (CL) build issues.
So, my understanding is that we first need to merge #1677 to fix the underlying build problem and get the Windows build to a healthy state. After that, we can proceed with optimizing the CI configuration by removing the redundant apiCheck calls to improve efficiency.

JakeWharton:
The Windows CI shard should not be triggering Linux compilation. Whether or not the build failure itself should be fixed is an orthogonal problem.

crowforkotlin:
Ah, that makes perfect sense. The issue is apiCheck forcing cross-compilation everywhere.

.github/workflows/.java-version
```diff
@@ -1 +1 @@
-24
+21
```
JakeWharton, line 1: There's only "LTS" builds for Windows ARM, and we can't jump to 25 yet

.github/workflows/build.yaml
```diff
@@ -66,6 +66,8 @@ jobs:
           - macos-15
           - ubuntu-24.04
           - ubuntu-24.04-arm
+          - windows-2025
+          - windows-11-arm
```
JakeWharton, line 70: GitHub just continually crushing it on naming these things...

swankjesse (approved):

PR #1682, opened 2025-09-30
Overhaul CI to add coverage

JakeWharton:
This runs the check task on the root project (sans tests we run elsewhere) which should start catching everything. It removes apiCheck from the platform test shards as that requires building all targets, so they should get faster and work on Windows.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,7 +10,7 @@ on:
       - '**'
 
 env:
-  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false -Dorg.gradle.logging.stacktrace=full"
 
 jobs:
   zig:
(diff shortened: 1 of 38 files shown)


JakeWharton:
Failing check is good. Shows our missing coverage today.

JakeWharton:
Good first start. Going to split up the JVM and native platform tests next, since we can run JVM on more platforms than native and we can do so without rebuilding them.

.github/workflows/build.yaml
```diff
@@ -127,14 +173,30 @@ jobs:
           pod install
           xcodebuild -workspace WorldClock.xcworkspace -scheme WorldClock -destination 'platform=iOS Simulator,name=iPhone 16,OS=latest'
 
-  build:
-    runs-on: macos-latest
+  final-status:
```
swankjesse, line 176: I like this pattern

build.gradle.kts
```diff
@@ -71,9 +57,27 @@ allprojects {
...
+      // Avoid 'build' folders within test fixture projects which may contain generated sources.
+      targetExclude("src/test/fixture/**/build/**")
+
+      ktlint()
+        .editorConfigOverride(
+          mapOf(
+            "ktlint_standard_comment-spacing" to "disabled", // TODO Re-enable
+            "ktlint_standard_filename" to "disabled",
+            "ktlint_standard_indent" to "disabled", // TODO Re-enable
+            // Making something an expression body should be a choice around readability.
```
swankjesse, line 74: 👍🏻

swankjesse (approved):

PR #1686, opened 2025-10-06
Split JVM and native tests

JakeWharton:
This lets us run them on different subsets of platforms. Add Linux ARM to JVM tests.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -31,20 +31,10 @@ jobs:
           path: zipline/src/jvmMain/resources/jni
           if-no-files-found: error
 
-  platform-tests:
+  jvm-test-binary:
     needs:
       - zig
-    strategy:
-      fail-fast: false
-      matrix:
-        platform:
-          - os: macos-13
-            task: macosX64Test
-          - os: macos-latest
-            task: macosArm64Test
-          - os: ubuntu-latest
-            task: linuxX64Test
-    runs-on: ${{ matrix.platform.os }}
+    runs-on: macos-latest
     steps:
       - uses: actions/checkout@v5
       - uses: actions/setup-java@v5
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -31,20 +31,10 @@ jobs:
           path: zipline/src/jvmMain/resources/jni
           if-no-files-found: error
 
-  platform-tests:
+  jvm-test-binary:
```
JakeWharton, line 34: This shard builds the JVM tests once, and they embed the necessary JS files in resources.

.github/workflows/build.yaml
```diff
@@ -57,25 +47,61 @@ jobs:
...
+      - uses: actions/download-artifact@v5
+        with:
+          name: jvm-tests
+
+      - name: Run JVM tests
+        run: |
+          chmod +x bin/zipline-test
+          bin/zipline-test
+
+  native-tests:
```
JakeWharton, line 89: We can't do the same with native because it reads the JS files directly from the file system. Maybe eventually I'll package them up so we can run the native tests on platforms we support but for which KGP does not support (e.g., Linux ARM).

zipline/build.gradle.kts
```diff
@@ -20,6 +20,7 @@ plugins {
   id("co.touchlab.cklib")
   id("com.github.gmazzo.buildconfig")
   id("binary-compatibility-validator")
+  id("com.jakewharton.test-distribution")
```
swankjesse, line 23: neat

.github/workflows/build.yaml
swankjesse: Yeah I really like that.

swankjesse (approved):

PR #1697, opened 2025-10-08
Create a wider gap between date captures

JakeWharton:
This avoids cases where the JVM returns the same timestamp for both captures (on Windows).

    1) dateNow(app.cash.zipline.QuickJsJvmTest)
    java.lang.AssertionError: 1.759945097065E12 <= 1.759945097066E12 <= 1.759945097065E12
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at app.cash.zipline.QuickJsJvmTest.dateNow(QuickJsJvmTest.kt:60)

### DIFF
--- a/zipline/src/jniTest/kotlin/app/cash/zipline/QuickJsJvmTest.kt
+++ b/zipline/src/jniTest/kotlin/app/cash/zipline/QuickJsJvmTest.kt
@@ -55,7 +55,9 @@ class QuickJsJvmTest {
 
   @Test fun dateNow() {
     val beforeMillis = System.currentTimeMillis().toDouble()
+    Thread.sleep(100)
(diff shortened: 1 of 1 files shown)


squarejesse (approved):

squarejesse:
The middle value is _greater_ than before or after? Wild. I gotta assume before and after are being truncated somehow?!

JakeWharton:
It's slightly greater than the after, which leads me to believe the JVM is perhaps caching calls. Or they're using different APIs on Windows, leading to the JS VM getting a more precise and thus slightly higher value.

JakeWharton:
The before and after are returning equal values.

Issue #1694, opened 2025-10-08
`ZiplineDevelopmentServerTest.webSocketReload` flaky


JakeWharton:
Seeing this _very_ frequently

```
ZiplineDevelopmentServerTest > webSocketReload FAILED
    org.opentest4j.AssertionFailedError: expected:<[1]> but was:<[0]>
        at app//app.cash.zipline.gradle.ZiplineDevelopmentServerTest.webSocketReload(ZiplineDevelopmentServerTest.kt:165)
```

PR #1698, opened 2025-10-08
Do not assert on number of sockets sent to

JakeWharton:
This is implied by the following assertion which asserts we receive the message.

This _changes_ how #1694 happens so that hopefully we see a better exception to know what's going on.


squarejesse (approved):

JakeWharton:
Still failing, but I guess with a better message indicating what we knew was happening...

PR #1720, opened 2025-11-03
Work around another bug around Unit return values

swankjesse:
In this case I believe a suspending function should return 'Unit' but is instead returning a different value.

The test involving GenericUnitService reproduces a crash where the Unit return value is promised but not returned. This fixes that test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Unreleased
 
-Nothing yet!
+ * Fix: Work-around the Kotlin compiler returning a non-`Unit` value for a function that declares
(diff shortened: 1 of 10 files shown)


squarejesse (commented):
We already have a work around for a similar issue, but that workaround assumes the `Unit` return type is known statically.

In this case we’re getting `Unit` through a generic, and our existing workaround doesn’t activate.

The generic function we’re hitting in our crash is likely `SuspendCallback.success()`.

JakeWharton (approved):
Well this sucks, but at least we're tightly coupled with the Kotlin version so we can drop this when the fix is released.

JakeWharton:
Change log too, please!

PR #1789, opened 2026-03-04
Use fresher manifest between cached and embedded

dellisd:
Previously a cached manifest would _always_ take precedence over any embedded manifests.

In practice we often have users who don't open the app for months, but whose app does update in the background from their app store. Their app launch experience is bad because the loader chooses to load potentially months-old code instead of the embedded code that may be very fresh.

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
@@ -22,6 +22,7 @@ import app.cash.zipline.loader.internal.fetcher.FsCachingFetcher
 import app.cash.zipline.loader.internal.fetcher.FsEmbeddedFetcher
 import app.cash.zipline.loader.internal.fetcher.HttpFetcher
 import app.cash.zipline.loader.internal.fetcher.LoadedManifest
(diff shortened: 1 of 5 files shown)


zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineLoader.kt
```diff
@@ -664,9 +665,15 @@ class ZiplineLoader internal constructor(
...
   ): LoadedManifest? {
-    val result = cachingFetcher?.loadPinnedManifest(applicationName, nowEpochMs)
-      ?: embeddedFetcher?.loadEmbeddedManifest(applicationName)
-      ?: return null
+    val pinnedManifest = cachingFetcher?.loadPinnedManifest(applicationName, nowEpochMs)
+    val embeddedManifest = embeddedFetcher?.loadEmbeddedManifest(applicationName)
+
+    // If both are available, pick the freshest of the two
+    val result = if (pinnedManifest != null && embeddedManifest != null) {
+      maxOf(pinnedManifest, embeddedManifest, LoadedManifestComparator)
```
dnagler, line 673: Love it!

dnagler (approved):
🙏

PR #1801, opened 2026-03-17
Disable gradle parallel builds on native tests

dellisd:
Maybe fix a concurrent modification exception?


JakeWharton:
This will still have parallelism through workers. You might need to set max workers to 1 to truly make it serial.

PR #1806, opened 2026-03-30
Guard ZiplineCache against use-after-close crashes

jhollida24:
## Summary

- Adds a `@Volatile` `closed` flag to `ZiplineCache` (as a class body property, not a constructor parameter) that is set **before** the SQLite driver is closed, ensuring visibility to concurrent threads
- Guards all database-accessing methods to bail out gracefully when the cache is closed:
  - `getOrPut` — falls back to `download()`
  - `read` — returns `null`
  - `unpin` — no-op
  - `getPinnedManifest` — returns `null`
  - `pinManifest` — no-op
  - `unpinManifest` — no-op
  - `updateManifestFreshAt` — no-op
  - `prune` — no-op
  - `countFiles` — returns `0`
  - `countPins` — returns `0`
- Prevents `sqlite3_reset` from being called on a destroyed connection when concurrent coroutines race against cache closure (root cause of iOS use-after-free crashes)

The pattern follows the existing `hasWriteFailures` guard — methods that already checked `hasWriteFailures` now also check `closed`, and methods without existing guards get standalone `if (closed)` checks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
+++ b/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
@@ -20,6 +20,7 @@ import app.cash.zipline.loader.internal.cache.Database
 import app.cash.zipline.loader.internal.cache.FileState
 import app.cash.zipline.loader.internal.cache.Files
 import app.cash.zipline.loader.internal.cache.SqlDriverFactory
+import kotlin.concurrent.Volatile
 import app.cash.zipline.loader.internal.cache.createDatabase
 import app.cash.zipline.loader.internal.fetcher.LoadedManifest
 import okio.ByteString
@@ -51,6 +52,7 @@ class ZiplineCache private constructor(
   private val maxSizeInBytes: Long,
   private val loaderEventListener: LoaderEventListener,
   private var hasWriteFailures: Boolean,
+  @Volatile private var closed: Boolean = false,
 ) : Closeable {
 
   /*
@@ -77,6 +79,7 @@ class ZiplineCache private constructor(
    */
 
   override fun close() {
+    closed = true
     driver.close()
   }
 
@@ -119,7 +122,7 @@ class ZiplineCache private constructor(
     nowEpochMs: Long,
     download: suspend () -> ByteString,
   ): ByteString {
-    if (hasWriteFailures) return download()
+    if (hasWriteFailures || closed) return download()
 
     try {
       val read = read(sha256, nowEpochMs)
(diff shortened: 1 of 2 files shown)


jhollida24:
Here's what a claude session had to say about the root cause of a crash. Because this is a public repo, I can't link to Bugsnag or many more details about Cash's implementation.

```
## Root Cause Analysis

**a. Full flow chain:**
1. **User signs out** → `RealSignOutCleaner.cleanUp()` → `TreehouseAppFactory.stopAllApps()` (cash-ios)
2. `stopAllApps()` closes TreehouseApps (`$0.close()`) then closes `appFactory.close()` (cash-ios)
3. `appFactory.close()` → `ZiplineCache.close()` → `driver.close()` (zipline)
4. Meanwhile, already-dispatched coroutine on `cacheDispatcher` executes `FsCachingFetcher.fetch()` → `cache.getOrPut()` → `database.filesQueries.get().executeAsOneOrNull()` (zipline)
5. SQLite query hits `resetStatement()` on destroyed connection → **EXC_BAD_ACCESS**

**b. Origin layer deep-dive:**
`ZiplineCache.close()` (line 79) simply calls `driver.close()` with zero coordination with in-flight operations. There's no `closed` guard. The `hasWriteFailures` pattern already handles degraded state gracefully — but no equivalent exists for close.

**c. Fix at origin + mitigate downstream:**
The fix is in the **zipline repo**: Add a `@Volatile closed` flag to `ZiplineCache`, checked before any DB access. This follows the existing `hasWriteFailures` pattern. No cash-ios changes needed — the sign-out sequence is correct, the race is purely in ZiplineCache's lack of close-safety.
```

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
```diff
@@ -51,6 +52,7 @@ class ZiplineCache private constructor(
   private val maxSizeInBytes: Long,
   private val loaderEventListener: LoaderEventListener,
   private var hasWriteFailures: Boolean,
+  @Volatile private var closed: Boolean = false,
```
dellisd: This should be moved out of the constructor, as we would never want to construct a cache in an already-closed state.

dellisd (commented):
There are a number of methods that access the database that probably also need guarding.

zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/ZiplineCache.kt
jhollida24: Oh good call. Fixed.

jhollida24:
> There are a number of methods that access the database that probably also need guarding.

Made these updates:
```
  - prune() — returns early when closed                                                                                                                                                                                          
  - countFiles() — returns 0 when closed                                                                                                                                                                                         
  - countPins() — returns 0 when closed     
  ```

dellisd (approved):
I think there are still some methods that access the database which aren't guarded (like `setReady` and `openForWrite`), but once those are updated this looks good to merge

PR #1836, opened 2026-08-26
Upgrade AGP to 9.3.1

dellisd:
- Upgrade Gradle to 9.5.1
- Migrate modules to the Android KMP library plugin
- Split out Android NDK build into a new `zipline-android-ndk` module as required by the Android KMP plugin


eric-labelle (approved):

zipline-loader/build.gradle.kts
```diff
@@ -15,8 +16,17 @@ plugins {
...
 kotlin {
   jvm()
-  androidTarget {
-    publishLibraryVariants("release")
+  android {
+    namespace = "app.cash.zipline.loader"
+    compileSdk = libs.versions.compileSdk.get().toInt()
+    minSdk = libs.versions.minSdk.get().toInt()
+
+    withDeviceTest {}
```
theisenp, line 24: Just double checking, does this do anything if we aren't specifying an `instrumentationRunner` like we are below?

zipline/build.gradle.kts
```diff
@@ -38,9 +51,37 @@ dependencies {
...
+
+    packaging {
+      // We get multiple copies of some license files via JNA, which is a transitive dependency of
+      // kotlinx-coroutines-test. Don't fail the build on these duplicates.
+      resources {
+        excludes += listOf("META-INF/AL2.0", "META-INF/LGPL2.1")
+      }
+
+      dex {
+      }
```
theisenp: Similar question, does this do anything if it's empty?

theisenp (approved):

zipline-loader/build.gradle.kts
dellisd: I don't think we were actually running any of the device tests (I'm not sure why) but this is needed to have the source set created

